PR #166, opened 2017-02-26
Remove tool type from switchBlock -> by default it will saved from previous

khaydarov:
regex match must not be similar to pasted string. Ex:
```https://www.instagram.com/p/BQ3wxmlA5GN/?taken-by=thenotoriousmma``
and match is ```https://www.instagram.com/p/BQ3wxmlA5GN/```.

```SwitchBlock``` have 3 parameter that is not important

### CODE REVIEW

File: modules/paste.js [hunk 1]
```diff
@@ -72,7 +73,7 @@ module.exports = function (paste) {
             var execArray = pattern.regex.exec(string),
                 match     = execArray && execArray[0];
 
-            if ( match && match === string.trim()) {
+            if ( match ) {
```
gohabereg: А так не будет опять багов со ссылкой в тексте?

modules/content.js
```diff
@@ -347,7 +347,8 @@ module.exports = (function (content) {
      */
     content.switchBlock = function (blockToReplace, newBlock, tool) {
 
-        var newBlockComposed = editor.content.composeNewBlock(newBlock, tool);
+        var tool = tool || editor.content.currentNode.dataset.tool,
```
neSpecc: tool уже объявлена в параметрах функции, ее не нужно в var

PR #263, opened 2018-06-24
Events triggering

khaydarov:


docs/events.md
```diff
@@ -0,0 +1,50 @@
+# CodeX Editor Events Module
+
+Module allows Developers subscribe on events or trigger callbacks
```
neSpecc: to subscribe

neSpecc: trigger own events

src/components/block-tunes/block-tune-delete.ts
```diff
@@ -0,0 +1,98 @@
+/**
+ * @class MoveUpTune
```
neSpecc: Delete Tune

src/components/modules/api-events.ts
```diff
@@ -0,0 +1,60 @@
+declare var Module: any;
+
+import { IEventsAPI } from '../interfaces/api';
+
+/**
+ * @class ToolbarsApi
```
neSpecc: EventsAPI

PR #547, opened 2018-12-05
Optimize and improve bundling

gohabereg:
- move bundle to dist folder
- set up builds for prod and dev envs
- minimize and optimize bundle to reduce size
- add banner at the top of the bundle
- add licenses file


gohabereg:
close #422

PR #661, opened 2019-04-01
Rename internal tools

### LINKED ISSUE #637: Cannot add Link Tool with toolname "link"
Why I should avoid using "link" name? "linkTool" name works correctly. 

image

gohabereg:
close #637


gohabereg:
Freezed until next major release

PR #673, opened 2019-04-03
fixed HTMLJanitor

0xflotus:


neSpecc:
Thanks for fixing this.

Issue #696, opened 2019-04-07
Allow holderId work with ref on dom element.


dimensi:
For frameworks like vue or react, very easy give reference on dom element than give id(string) on dom element.

neSpecc:
Good idea, thanks.

dimensi:
@neSpecc i want pick this feature on development.

PR #710, opened 2019-04-09
Allow holderId work with ref on dom element

dimensi:
Refer to #696


src/components/modules/blockManager.ts
```diff
@@ -455,7 +455,7 @@ export default class BlockManager extends Module {
    * @returns {Block}
    */
   public getBlock(element: HTMLElement): Block {
-    if (!$.isElement(element)) {
```
dimensi: Not understand why you do this. Then i change return type isElement method, i start get errors
`[tsl] ERROR in /home/dimensi/projects/@editorjs/editor.js/src/components/modules/blockManager.ts(459,25)
      TS2339: Property 'parentNode' does not exist on type 'never'.`

src/components/modules/ui.ts
```diff
@@ -134,11 +134,7 @@ export default class UI extends Module {
      * Element where we need to append Editor.js
      * @type {Element}
      */
-    this.nodes.holder = document.getElementById(this.config.holderId);
-
-    if (!this.nodes.holder) {
```
dimensi: I remove it, because we check on exists in `validation` method

src/components/modules/ui.ts
```diff
@@ -329,7 +325,8 @@ export default class UI extends Module {
      */
     const target = event.target as HTMLElement;
     const clickedOnInlineToolbarButton = target.closest(`.${this.Editor.InlineToolbar.CSS.inlineToolbar}`);
-    const clickedInsideOfEditor = !!target.closest(`#${this.config.holderId}`) || Selection.isAtEditor;
```
dimensi: If we have reference on holder element,  i think not need use `closest` method.

src/components/dom.ts
```diff
@@ -518,4 +518,27 @@ export default class Dom {
...
+    if (typeof element === 'string') { return document.getElementById(element); }
+    return element;
+  }
+
+  /**
+   * Check exists target in holder
+   * @param element - target
+   * @param holder - holder
+   */
+  public static elementInHolder(element: HTMLElement, holder: HTMLElement): Element {
```
gohabereg: There are `Node.contains` method for the same purpose

https://developer.mozilla.org/en-US/docs/Web/API/Node/contains

dimensi: Thx, I totally forgot about this method.

src/components/modules/ui.ts
dimensi: fixed

src/components/dom.ts
dimensi: fixed

types/configs/editor-config.d.ts
```diff
@@ -5,8 +5,9 @@ import {SanitizerConfig} from './sanitizer-config';
 export interface EditorConfig {
   /**
    * Element where Editor will be append
+   * TODO: rename holderId on holder, element or el in next major for better consistent
```
neSpecc: It's better to add a `holder` config property right now and leave `holderId` for the backward compatibility. We will update docs and remove holderId from there. In next releases we will remove `holderId`

dimensi: Fix it

src/components/core.ts
```diff
@@ -190,16 +190,23 @@ export default class Core {
     /**
      * Check if holderId is not empty
      */
-    if (!this.config.holderId) {
-      throw Error('«holderId» param must being not empty');
+    if (!this.config.holderId && this.config.holder) {
```
gohabereg: I guess here is a typo

```suggestion
    if (!this.config.holderId && !this.config.holder) {
```

src/components/core.ts
```diff
@@ -190,16 +190,23 @@ export default class Core {
...
      */
-    if (!$.get(this.config.holderId)) {
+    if (!this.config.holder &&
+        typeof this.config.holderId === 'string' &&
+        !$.get(this.config.holderId)
+      ) {
       throw Error(`element with ID «${this.config.holderId}» is missing. Pass correct holder's ID.`);
     }
+
+    if (this.config.holder && !$.isElement(this.config.holder)) {
```
gohabereg: I think `holder` property should also accept string because we will remove `holderId` on major release

src/components/core.ts
dimensi: thx

src/components/core.ts
gohabereg: Reference is not always the most convenient way I guess

dimensi: One thing we do it's call getElementById().
```js
new EditorJS({
 holder: document.getElementById('id'),
})

new EditorJS(document.getElementById('id'))
```
Or user write on 23 symbols more or we add 2 if's

gohabereg: Better if it is easier for user

dimensi: update it

PR #711, opened 2019-04-09
#665 API to open and close inline-toolbar

### LINKED ISSUE #665: [Core API]: Provide methods that opens/closes inline toolbar

tanmayv:
This is regarding the issue #665. I tested this on `inline-tool-link` and `inlineToolbar` is available to close it.

Resolves #665


neSpecc:
Resolves #665

types/api/index.d.ts
```diff
@@ -8,3 +8,4 @@ export * from './styles';
 export * from './caret';
 export * from './toolbar';
 export * from './notifier';
+export * from './inline'
```
neSpecc: It would better to name it `inlineToolbar`

src/components/modules/toolbar/inline.ts
```diff
@@ -170,7 +170,7 @@ export default class InlineToolbar extends Module {
   /**
    * Shows Inline Toolbar
    */
-  private open(): void {
```
tanmayv: @neSpecc This file is also named `inline.ts`. Should we rename this as well?

neSpecc: no, it's included at the `/toolbar/` directory, so this is a redundant addition

tanmayv:
Do i need to `dist/editor.js` as well? It was in `dist/*` so I skipped it.

neSpecc:
> Do i need to `dist/editor.js` as well? It was in `dist/*` so I skipped it.

yes, currently we have a bundle in a repo. Maybe removed later.

tanmayv:
Okay I'll push `dist/editor.js` as well.

tanmayv:
@neSpecc Should I build bundle in `prod`, or `dev` is okay?

gohabereg:
@tanmayv 
The example-dev.html page doesn't work in your branch

tanmayv:
@gohabereg Fixed the issue. `src/components/modules/api/inline-toolbar.ts` can't have hyphen in it so change it to `src/components/modules/api/inlineToolbar.ts`

neSpecc:
Also error throws event when one block is selected, but without caret in there. Click at the Block's right zone, near three-dots, then call  `open`

tanmayv:
>     * open editor
> 
>     * do not select any block
> 
>     * call `api.inlineToolbar.open()`
> 
> 
> Error will be thrown
> 
> ```
> TypeError: null is not an object (evaluating 't.parentNode')
> value — editor.js:4566
> value — editor.js:6737
> value — editor.js:6712
> value — editor.js:3925
> Global Code
> evaluateWithScopeExtension
> _evaluateOn
> _evaluateAndWrap
> ```
> 
> You should add a checking for a `BlockManager.currentBlock` existence at the `toolbar/inline.js@open`

This should be fixed now.

tanmayv:
> Also error are throwed even when one block is selected, but without caret in there. Click at the Block's right zone, near three-dots, then call `open`

I wasn't able to recreate this. Is this still happening?

tanmayv:
Updated the `changelog.md` and `api.md` as well to document new api methods.

PR #722, opened 2019-04-11
Close inline toolbar after creating new link by pressing ENTER

### LINKED ISSUE #600: Close Toolbar after Link insertion
1. Select fragment of text
2. Click on Link Tool at the Inline Toolbar (or press CMD+K)
3. Enter or paste a link and press Enter
4. Link will be inserted, but Inline Toolbar still opened

We need to close the Inline Toolbar after link insertion

tanmayv:
I have kept old toolbar as well. But I think have to remove that. And use the inlineToolbar open/close method. Give me a confirmation and I'll make the change.

Reason for adding clear selection method. If selection was present, inline toolbar will reopen again after closing for that selection.
resolved #600


src/components/selection.ts
```diff
@@ -267,6 +267,28 @@ export default class SelectionUtils {
...
+  /**
+   * Clear current window selection
+   */
+  public clearWindowSelection(): void {
+    const sel = window.getSelection();
+    const range = document.createRange();
+    /**
+     * Set node content to node where selection ends
+     */
+    range.selectNodeContents(sel.focusNode);
```
khaydarov: I think `clearWindowSelection` should not contain Range restoration functionality, because It might be used separately. Probably, it is better to move this logic to another method and add to this function optional argument `restoreRange` which is `false` by default? And then according to the value of this flag call additional function

tanmayv: Selection has API for collapse to END, i'm trying that. I might not need clear window selection method.

khaydarov: Are you sure about browsers compatibility? https://developer.mozilla.org/en-US/docs/Web/API/Selection/collapseToEnd

tanmayv: Yeah, `collapseToEnd` is marked as experimental. So I replaced it.

khaydarov: Good

src/components/selection.ts
```diff
@@ -267,6 +267,18 @@ export default class SelectionUtils {
     this.savedSelectionRange = null;
   }
 
+  /**
+   * Collapse current selection
+   */
+  public collapseToEnd(): void {
+    const sel = window.getSelection();
+    const range = document.createRange();
```
gohabereg:
```suggestion
    const range = document.createRange();
    
```

Issue #715, opened 2019-04-10
Add `blocks.append(type, data)` API method


gohabereg:
This method should allow to append (or insert) new Blocks programmatically

Example:

```
editor.blocks.append('paragraph', {text: 'Hello world'});

editor.blocks.insert(4, 'header', {text: 'Hello world'});
```

neSpecc:
I thought, the position can be an optional argument of the `append`

gohabereg:
Better to name it `insert` then

khaydarov:
I tend to think that it is better to use output format as a argument of append function, like 
```js
{
  type: 'header',
  data: { 
    ....
  } 
}
```

and second argument will be `position`

```
const newBlock = {
  type: 'paragraph', 
  data: {
     text: 'Hello world'
  }
};
editor.blocks.append(newBlock, 4);
```

neSpecc:
@cjagannath please, describe the use case of that feature. How you would generate a block's data structure? It is an internal logic of a plugin. And `type` is the variable that defined by plugins users.

I can't see a good implementation of it, except hardcoding.

dimensi:
@neSpecc for inline tool for example. We can use this api for extract text from selection and create new Header or Paragraph element.

neSpecc:
@dimensi to create a Header, you should pass for example

```
{
   text: 'Extracted fragment',
   level: 3
}
```

How you will create this structure? Your Inline Tool doesn't know about properties that used by Header data (I mean `text` and `label`). So you are only able to hard code them.

neSpecc:
For blocks convertation we already develop an ability (#704). Any other use cases?

dimensi:
> @dimensi to create a Header, you should pass for example
> 
> ```
> {
>    text: 'Extracted fragment',
>    level: 3
> }
> ```
> How you will create this structure? Your Inline Tool doesn't know about properties that used by Header data (I mean `text` and `label`). So you are only able to hard code them.

Yes hardcode, for you use inline tools like this, inline tool must check through validation whether or not there's a Header class. Something like that
```js
{
  render() {
   const blocks = this.api.getBlocks()  // this api not implement
   const hasHeader = blocks.findIndex(el => el.class instanceof Header)
   if (hasHeader) return button;
   return;
  }
}
```

cjagannath:
> @cjagannath please, describe the use case of that feature. How you would generate a block's data structure? It is an internal logic of a plugin. And `type` is the variable that defined by plugins users.
[...]

PR #753, opened 2019-04-30
issue-715: insert method in blocks API

kleoxhindoli:
Does the following:
1. Adds a `insertAt` method in the `blockManager`. Same as `insert` but accepts an index param as well
1. Refactored `insert` to use `insertAt` to avoid code duplication
1. Exposes the `insertAt` method to the public API, via the `editor.blocks.insert` method, to allow insertion of *any* blocks programmatically at any index

Solves https://github.com/codex-team/editor.js/issues/715

*First PR, feedback is welcomed*

### DIFF
--- a/src/components/modules/api/blocks.ts
+++ b/src/components/modules/api/blocks.ts
@@ -1,9 +1,9 @@
 import Module from '../../__module';
 
-import {Blocks} from '../../../../types/api';
-import {OutputData} from '../../../../types';
+import { Blocks } from '../../../../types/api';
+import { BlockToolData, OutputData, ToolConfig } from '../../../../types';
 import Block from '../../block';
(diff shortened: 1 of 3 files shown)


src/components/modules/api/blocks.ts
```diff
@@ -26,6 +26,12 @@ export default class BlocksAPI extends Module {
       getBlocksCount: () => this.getBlocksCount(),
       stretchBlock: (index: number, status: boolean = true) => this.stretchBlock(index, status),
       insertNewBlock: () => this.insertNewBlock(),
+      insert: (
+        index: number,
```
gohabereg: Maybe make `index` optional and insert new block at the current position?

gohabereg:
Hi @kleoxhindoli 
Thank you for the contribution, the issue resolved by #814

PR #786, opened 2019-05-30
Reference enableLineBreaks setting correctly

bkroeker:
The enableLineBreaks setting is currently broken.  These changes ensure that it's referenced correctly.

### DIFF
--- a/src/components/modules/blockEvents.ts
+++ b/src/components/modules/blockEvents.ts
@@ -225,13 +225,13 @@ export default class BlockEvents extends Module {
   private enter(event: KeyboardEvent): void {
     const {BlockManager, Tools} = this.Editor;
     const currentBlock = BlockManager.currentBlock;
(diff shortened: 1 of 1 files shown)


neSpecc:
Thanks. We need to some more research and work here, the problem is little deeper. Will be included in next releases.

gohabereg:
Hi @bkroeker 
The code is actually correct - enableLineBreak option should be provided by Tool developer as Tool`s class property, so you are not able to define it in Editor`s config. 
If you saw the incorrect description somewhere in the docs, please let us know.

bkroeker:
Ah, I see now.  I was trying to find a way to disable line breaks for the tools I was adding, and I didn't see anything in the docs, so I was scrounging through the code.  That setting showed up in the same place as other settings you set in the config when initially registering tools, so I assumed it should behave the same way.  I guess if I want to disable line breaks on default tools, I need to clone them each and set that static property on each.

Issue #788, opened 2019-06-02
Types are missing shorthands


jansauer:
The editor exports a hand full of [shorthands](https://github.com/codex-team/editor.js/blob/9e61e3ddea47d426cb7c45994697167ca8f96b25/src/codex.ts#L106) as part of it's external API methods. These shorthands are missing in the type definitions. This leads to an unnecessarily higher barrier of entry.

For example, in the documentation, saving data is multiple times describe with `editor.save()`. This leads to a type error in typescript since without the shorthand it would need to be `editor.saver.save()`.

PR #789, opened 2019-06-02
Add the save api shorthand to the types

jansauer:
This mr intentionally does not add all shorthands as long it is not
clear if this is the intended goal.

See #788


gohabereg:
Thanks @jansauer !

Issue is resolved by #800

PR #862, opened 2019-08-07
Enable user order based on specifications

despeauxz:


src/components/modules/toolbar/inline.ts
```diff
@@ -96,12 +96,17 @@ export default class InlineToolbar extends Module {
...
-      this.toolsInstances = new Map();
-      for (const tool in allTools) {
-        if (allTools.hasOwnProperty(tool)) {
-          this.toolsInstances.set(tool, allTools[tool]);
-        }
-      }
+      const orderedTools = new Map();
+
+      // get the user order specifications
+      const order = this.config.tools.header.inlineToolbar;
```
neSpecc: you should not hardcode Header tool. Instead of this, get current Block's tool. 

User can specify different `inlineToolbar` options for all tools, or not to specify it at all. So, it seems like we need to reorder the Inline Toolbar on every opening.

To optimize calculations, you can cache toolbar order for each tool.

neSpecc (changes requested):

src/components/modules/toolbar/inline.ts
despeauxz: Okay

src/components/modules/toolbar/inline.ts
```diff
@@ -96,12 +96,17 @@ export default class InlineToolbar extends Module {
...
+
+      // iterate through the list of user specified tools
+      order.map((tool: string) => {
```
neSpecc: order (`inlineToolbar config option`) can be unspecified or specified as `Boolean`. You need to consider this.

PR #866, opened 2019-08-15
Fix case when property 'observer' in modificationObserver is not defi…

VLZH:
Property 'observer' in modificationObserver will define in timeout(1s). There is situation when we will destroy editor before this timeout.


VLZH:
Maybe you need to set on strict mode in tsconfig.json?

VLZH:
@neSpecc I added changes.

PR #902, opened 2019-09-27
Process patterns on each line on paste & pattern_processing_max_length moved to config

step4:
Before when you paste something which can be split by a newline each line will be seen as a 'tag'-event
Now each line will also be processed by `processPatterns` so if a pattern is found the line will be seen as a 'pattern'-event.

Also you can now add a `pattern_processing_max_length` prop in the config


neSpecc:
Seems outdated. If it still actual, please describe the usecase via the Issue

PR #904, opened 2019-09-30
pass options on removeEventListener

MarinaZadoyanchuk:
According to https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener, when we remove listener, we should pass `options` which was used for the listener on `addEventListener`.

In `removeAll` method `options` are missed and after calling `editor.destroy()` some listeners stay on


khaydarov:
@MarinaZadoyanchuk could you update the bundle, please? But before you need to merge your branch with master

PR #905, opened 2019-10-01
adds delete-block event

alexandrmih89:
Adds a new event when a block is deleted.


src/components/block.ts
```diff
@@ -446,7 +446,7 @@ export default class Block {
    * @return {BlockTune[]}
    */
   public makeTunes(): BlockTune[] {
-    const tunesList = [MoveUpTune, DeleteTune, MoveDownTune];
+    const tunesList = [];
```
khaydarov: These three tunes are default Block settings. Your line is removing default behavior

khaydarov:
And remove package-lock.json please

gohabereg:
There is [Block lifecycle hooks](https://github.com/codex-team/editor.js/pull/968) introduced in 2.16

PR #906, opened 2019-10-03
Block lifecycle hooks

gohabereg:


src/components/block.ts
```diff
@@ -28,6 +28,18 @@ import DeleteTune from './block-tunes/block-tune-delete';
 import MoveDownTune from './block-tunes/block-tune-move-down';
 import SelectionUtils from './selection';
 
+/**
+ * Available Block Tool API methods
+ */
+export enum BlockToolAPI {
```
neSpecc: maybe it would better to place in `tools.ts` near `INTERNAL_SETTINGS` and `USER_SETTINGS` ? and use same structure

src/components/block.ts
```diff
@@ -28,6 +28,18 @@ import DeleteTune from './block-tunes/block-tune-delete';
...
+  /** TODO: deprecate method */
```
neSpecc:
```suggestion
  /** 
  * @todo remove method in 3.0.0 
  * @deprecated — use 'rendered' hook instead
  */
```

gohabereg:
> * Errors thrown inside block hooks does no logged and handled anywhere

Errors are logged by Editor`s internal log method

gohabereg:
> * `updated` does not called when block moving/swap

Don't think it should. Block moving doesn't affect the block itself.

src/components/block.ts
gohabereg: I think it is better to left it here because in this case `Block` class be encapsulated as it is now

src/components/blocks.ts
```diff
@@ -220,4 +231,36 @@ export default class Blocks {
...
+  /**
+   * Fires when Editor is loaded.
+   *
+   * All pending tasks should be processed here.
+   */
+  public onEditorReady(): void {
+    this.editorIsReady = true;
+
+    this.pendingBlocks.forEach((block) => block.call(BlockToolAPI.RENDERED));
+  }
```
neSpecc: let's document `@this` context type

src/components/blocks.ts
```diff
@@ -220,4 +231,36 @@ export default class Blocks {
...
     return this.blocks.indexOf(block);
   }
+
+  /**
+   * Fires when Editor is loaded.
+   *
+   * All pending tasks should be processed here.
+   */
+  public onEditorReady(): void {
+    this.editorIsReady = true;
```
neSpecc: would be nice to have a comment describing this set;

src/components/blocks.ts
```diff
@@ -47,16 +46,17 @@ export default class Blocks {
    * blocks[0] = new Block(...)
    *
    * @param {Blocks} instance — Blocks instance
-   * @param {Number|String} index — block index
-   * @param {Block} block — Block to set
+   * @param {Number|String} key — block index or any Blcoks class property
```
neSpecc:
```suggestion
   * @param {Number|String} key — block index or any Blocks class property to set
```

src/components/blocks.ts
```diff
@@ -47,16 +46,17 @@ export default class Blocks {
...
+   * @param {Block} value — value to set
    * @returns {Boolean}
    */
-  public static set(instance: Blocks, index: number, block: Block) {
-    if (isNaN(Number(index))) {
-      return false;
+  public static set(instance: Blocks, key: number | string, value: Block | any) {
+    if (isNaN(Number(key))) {
```
neSpecc: a comment describing this case would be helpful

src/components/blocks.ts
```diff
@@ -68,9 +68,9 @@ export default class Blocks {
    * @param {Number|String} index — Block index
    * @returns {Block|*}
    */
-  public static get(instance: Blocks, index: number) {
+  public static get(instance: Blocks, index: any | number) {
     if (isNaN(Number(index))) {
```
neSpecc: a comment describing this case would be helpful

src/components/blocks.ts
```diff
@@ -86,6 +86,10 @@ export default class Blocks {
    */
   public workingArea: HTMLElement;
 
+  private editorIsReady: boolean = false;
```
neSpecc: docs are missed

src/components/core.ts
```diff
@@ -93,6 +93,8 @@ export default class Core {
            * Resolve this.isReady promise
            */
           onReady();
+
+          this.moduleInstances.Events.emit('editor-ready');
```
neSpecc: events should be places in some constant

src/components/blocks.ts
mmiszy: You could also use TypeScript to add type to `this`:
`public onEditorReady(this: Proxy<Blocks>): void {`

src/components/core.ts
```diff
@@ -268,16 +273,24 @@ export default class Core {
...
     return this.moduleInstances.Renderer.render(this.config.data.blocks);
   }
 
   /**
    * Make modules instances and save it to the @property this.moduleInstances
    */
   private constructModules(): void {
-    modules.forEach( (Module) => {
+    modules.forEach( (module) => {
+      const Module = typeof module === 'function' ? module : module.default;
```
neSpecc: need a comment describing this case

src/components/core.ts
```diff
@@ -93,6 +96,8 @@ export default class Core {
            * Resolve this.isReady promise
            */
           onReady();
+
+          this.moduleInstances.Events.emit(EDITOR_EVENTS.READY);
```
neSpecc: where this event is using at this PR?

Issue #670, opened 2019-04-02
Add RTL support


eyy:
https://en.wikipedia.org/wiki/Bi-directional_text

Thanks.

amitassaraf:
+1

abublihi:
Hi, I have examine Editor.js and it is mostly support RTL,

you can find the what i have done on this codepin [https://codepen.io/anon/pen/zQqeeM](https://codepen.io/anon/pen/zQqeeM)

IssaMokdad:
The rtl feature is broken, I am setting the direction to rtl in i18n config, but it doesn't work. any idea regarding the issue?

ShahriarKh:
Is this implemented? If so, can somebody please explain how to use it?

RTL/LTR buttons in Word:
![image](https://user-images.githubusercontent.com/31452340/136075128-93a251d9-ed4d-4dc3-9cf4-202069c79588.png)

PR #909, opened 2019-10-04
Add RTL support

GINCHER:
Issue: https://github.com/codex-team/editor.js/issues/670

I have created `example-dev-rtl.html` with Hebrew in the editor to be able to test the RTL. To make the editor RTL, the user should add to the config `{direction: 'rtl'}`.

**Tasks done:**
- Modify CSS to support RTL
- Calculate distance from the right for the tooltip
- Right and left keys will move the user between blocks correctly
- Inline URL input field will remain LTR


hoseinz3xs:
@neSpecc  @khaydarov @gohabereg Could you please review this feature? I definitely wanna this one

neSpecc:
@hoseinz3xs sure, we will review it soon.

GINCHER:
- Flipper now works correctly with the keyboard arrows when RTL
- Conversion window now appears on the right side

rezam90:
@neSpecc any updates on this feature?

neSpecc:
@rezam90 we'll try to include it to the next release.

kfirba:
@neSpecc Is there any ETA as of when will this be merged? This seems like the only missing piece so we can use the editor in production :)

neSpecc:
@kfirba May — June

kfirba:
@neSpecc oh wow, that was a snappy response. Thanks!

Is it guaranteed that RTL support is coming in the next version?

khaydarov:
@GINCHER could you re-open your PR, please? It seems to be unavailable

arfrix:
I'll need that too

PR #911, opened 2019-10-05
Fix tooltip plus icon hover

torelafes:
When EditorJS gets initialized, hovering over plus icon doesn't update `transform` property unless toolbars are open and hovered once.

image

This PR sets required `transform` property for tooltip on initialization.


neSpecc:
Resolved in 2.16 #966

PR #913, opened 2019-10-08
Fix inline toolbar closing on escape key press

### LINKED ISSUE #692: Close inlineToolbar by escape

shannonrothe:
When the inline toolbar _or_ conversion toolbar are open, the escape key should close either of them. Currently both the toolbars close momentarily and then re-open when escape is pressed.

This PR removes code that appears to have been duplicated from the `mouseUp` event.

Resolves #692

### DIFF
--- a/src/components/modules/blockEvents.ts
+++ b/src/components/modules/blockEvents.ts
@@ -4,7 +4,7 @@
 import Module from '../__module';
 import _ from '../utils';
 import SelectionUtils from '../selection';
-import Flipper from "../flipper";
+import Flipper from '../flipper';
 
 export default class BlockEvents extends Module {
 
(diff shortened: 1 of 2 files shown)


Shazer2:
> This code was used to open Inline Toolbar or Conversion Toolbar by keyboard-selection. Try to use `SHIFT+ARROWS` to select some fragment — in your branch Toolbars won't be shown.
> 
> You can try to determine when keyup fired with ESCAPE key or try to use `selectionchange` event (you'll find it at the UI module) to show Toolbar by `SHIFT+ARROWS` instead of keyup. The second way looks more desirable.

Is there a reason the selectionchange event is only listened for on mobile devices currently?

src/components/modules/blockEvents.ts
```diff
@@ -112,10 +112,8 @@ export default class BlockEvents extends Module {
     if (SelectionUtils.almostAllSelected(block.pluginsContent.textContent)) {
```
neSpecc: does this code still required here?

Shazer2: I don't think so - do we still need to handle `BlockSettings` being closed if `almostAllSelected` evaluates to true on a `keyup`?

Shazer2:
@neSpecc I've removed the redundant code from `blockEvents.ts`, please let me know if any further changes are required :)

neSpecc:
@shannonrothe thanks for the contribution. This code was moved to #932 along with other updates of Inline Toolbar

PR #924, opened 2019-10-20
Add underline inline tool

GINCHER:
Added underline inline tool. 

The SVG don't look as good as the other ones, sorry.


neSpecc:
It is better to move this code to the separate Tool's repo: https://github.com/editor-js/underline

natterstefan:
Hello, I've created a generic inline-tool. It comes with underline support. You can find it here: https://www.npmjs.com/package/editorjs-inline-tool

Issue #836, opened 2019-07-02
Layout scheme


neSpecc:
- [ ] Create rows on render, by default each row haw 1 column
- [ ] Add ability to drag-n-drop block to the left or to the right of other block, creating several columns layout
- [ ] Add `id` to each block to saved data `blocks` property 
- [ ] Add `layout` property to saved data object.

The `layout` should describe layout scheme of Blocks by their ids. For example

```
[---1---]
[---2---]
[ 3 ] [ 4 ]
[---5---]
```

should be described as 

```
[1, 2, [3, 4], 5]
```
```

**Comment by Mehdi-Hp** [comment]
Hey. Smart feature. Nice 👏
We really need this in our project. Do you think there is any simpler and temporary way I can implement this with current API?

**Comment by calumk** [comment]
+1 for this feature

**Comment by hata6502** [comment]
This is very interested feature! 😃
But Editor.js may become difficult to maintain.

**Comment by christianostrem** [comment]
This is the only feature I need to start using Editor.js in my project.
I think that this feature should be thought of I a more generic mater.
Though it's great in its use for layout. It could do more than that. The setup inside Editor.js would be the same but the output could also be:

Tabs (vertical/horizontal)
Slides
Ul => li
Bootstrap => Rows and columns
And probably more...

Since editor.js allows us to set our own options. You can add options for:
Tabs and bootstrap like => Type: Tab / Boostrap or row size, it's basically up to you.

I have also explained some of it on my post #1199

**Comment by tvvignesh** [comment]
+1 from my end too please. This would be really important to have. 

@neSpecc Also, is it limited to 2 columns or can it be more?

For eg. if I need a row with 4 columns, while the data model should remain the same making it `[1, 2, [3, 4,5,6], 7]` but I hope editor.js would be supporting that.

Thanks.

**Comment by smeijer** [comment]
Is the idea to limit it to a single level? Or will it be recursive nesting?

In other words, will this be possible?

```
[         1         ]
[         2         ]
[   3   ]   [   4   ]
[ 5 [ 6 [ 7 ] 8 ] 9 ]
```

oodeveloper:
Hi, I think this feature will make editorjs more awesome.
A Simple Idea:
Brief: By using CSS flex and bootstrap like column layout system there is no need to have row blocks in order to achieve grid layouting or even nested elements ( not an ultimate solution but a minimal editorjs style :| )
You can 
[...]

PR #925, opened 2019-10-21
added unique id to blocks

nmannitz:


neSpecc:
Nice work, it is useful feature. Can you please add an issue describing problem that resolves by this PR?

nmannitz:
Of cause. I've different use cases where i need a kind of id on any block. Currently i would like to build a custom order on top of editorjs to manipulate my DOM with Drag&Drop and a column grid. Now i can build my additional order for my DOM without core changes in editorjs. Thats my idea.

sumitvekariya:
@nmannitz Even I support to have a unique ID on each blocks, because I am implementing the micro comment feature, where users can "Comment" on the specific block. And to refer to that particular block, a unique id is required.

src/mixin.ts
```diff
@@ -0,0 +1,21 @@
+import uuidv4 from 'uuid/v4';
+
+export default {
+  createId() {
+    let id = uuidv4();
```
mupkoo: I wonder if a simple, dependency free solution can work here. Something like

```javascript
let counter = 0;

export default {
  createId() {
    let id = `block-${counter++}`;
    ...
```

nmannitz: Hi, i'll use unique id's because if you have more than one instance of editor on one page, you can't find a specific block without it on DOM.

src/mixin.ts
```diff
@@ -0,0 +1,21 @@
...
+      if (document.getElementById(id)) {
+        return true;
+      } else {
+        return false;
+      }
+    };
+
+    do {
+      id = uuidv4();
+    } while (!existId);
```
kwangure: This should be`while (!existId())` since it's a function.

arpadgabor:
Any update on this?

neSpecc:
Still no issue attached

neSpecc:
Blocks' ids are a useful feature, but the current version does not require it. We will add it along with the features that will use it. Maybe with #836 

If someone will have a use case of this feature, please give us a detailed description via an issue.

vincentdesmares:
@neSpecc I have two more use cases for this feature:

- Be able to diff two editors content
- Be able to apply modifications to another editor content (aka multi-live-editors)

alibail:
@neSpecc additional to the above use cases;

- The need to persist additional data per block that is used by a backend or additional system - the blocks themselves don't need to know about it, but for consistency, there needs to be a consistent way to either carry the assigned data or a reference to it through the editor and back out again through the save feature.

I already implemented unique IDs and metadata in a fork, that meets my needs after this was closed - so have solved the issue for myself - but there are many applications for being able to pass hidden data through the editor.

vincentdesmares:
@neSpecc Can you look at our last use cases? I really want to contribute to the project but the source team must be aligned on the goals.

qwerin:
next implement uniqueID on my fork thx @alibail for idea :-) 
https://github.com/qwerin/editor.js
PS. my first code in TypeScript

PR #928, opened 2019-10-24
Fix image example

floatas:
Looks like old tool is used in example. Removed git submodule and updated example.


neSpecc:
In example we use Simple Image tool instead of Image because the second one is requires server-side implementation.

PR #934, opened 2019-10-31
Add addCaption and tabCaption settings into editor-config.d.ts,

m-kur:
..., modify toolbar/index.ts.

In the EditorJS user interface, only the plus button's tooltip caption cannot be changed in the user settings. The static values ​​"Add" and "Tab" can be replaced with arbitrary words for the purpose of i18n. This feature allows me to use Japanese, "追加" and "タブ".

### DIFF
--- a/src/components/modules/toolbar/index.ts
+++ b/src/components/modules/toolbar/index.ts
@@ -121,9 +121,11 @@ export default class Toolbar extends Module {
       const tooltip = this.Editor.Toolbox.nodes.tooltip;
       const fragment = document.createDocumentFragment();
 
-      fragment.appendChild(document.createTextNode('Add'));
+      const addCaption = this.config.addCaption || 'Add';
+      const tabCaption = this.config.tabCaption || 'Tab';
+      fragment.appendChild(document.createTextNode(addCaption));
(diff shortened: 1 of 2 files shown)


gohabereg:
Hi @m-kur , thanks for contribution!
Using root properties of config object seems not really convenient. I think we should implement some more general i18n tools. Let me know if you want to help with that please!

m-kur:
For natural language appearing in EditorJS's UI, it is expected that language resources can be changed depending on both the application settings and the browser locale of the user.
EditorJS can be internationalized in Config (except for the additional buttons discussed this time), so the additional expectation is an implementation that changes language resources in the browser locale.
There are many/major libraries for internationalization that every application implementers prefer, so it would be good if you could specify function-typed-properties that takes the browser locale as an argument in addition to string-typed-properties in the EditorJS's display string.
This is necessary for non-native English speakers. 😄 

The caption of the `Add` and `Tab` pointed out this time is not set by a character string, so my requirements cannot be met unless something is prepared.

m-kur:
Can I close this PR?

m-kur:
Close it.

Issue #938, opened 2019-11-06
IndexSizeError: Failed to execute 'getRangeAt' on 'Selection': 0 is not a valid index.


breezes1:
https://github.com/codex-team/editor.js/blob/0600233e63fb92d17f0306c6b6c0beb2071c2880/src/components/selection.ts#L190-L197

PR #939, opened 2019-11-06
fix getRangeAt IndexSizeError

breezes1:
#938

PR #953, opened 2019-11-22
fix getRangeAt IndexSizeError

breezes1:
#938


breezes1:
#939

src/components/selection.ts
```diff
@@ -176,6 +176,8 @@ export default class SelectionUtils {
       return rect;
     }
 
+    if (sel.rangeCount === 0) { return rect; }
```
neSpecc: Following the code style guide: 

```suggestion
    if (sel.rangeCount === 0) { 
      return rect; 
    }
```

gohabereg:
Resolved by #968

Issue #984, opened 2019-12-05
Unable to hide conversion toolbar item via config


mpandzo:
With the introduction of the conversion toolbar action in 2.16.0 no option is provided in config to disable this feature for those of us that do not intend to use it. This is especially important when no addons are used which results in the following display:

![empty](https://user-images.githubusercontent.com/7498710/70246252-ba28bb00-1777-11ea-8b82-9353cc09bd9d.png)

neSpecc:
Resolved in 2.16.1

mikekoro:
> Resolved in 2.16.1

How do you actually disable it? I cant seem to find anything in the docs.

PR #991, opened 2019-12-07
Fix/hide conversion toolbar

khaydarov:


src/components/modules/toolbar/conversion.ts
```diff
@@ -141,6 +141,15 @@ export default class ConversionToolbar extends Module {
     }
   }
 
+  /**
+   * Returns true if it has any more than one conversion tool
```
neSpecc:
```suggestion
   * Returns true if it has more than one tool available for convert in
```

khaydarov:
related #984

src/components/modules/toolbar/conversion.ts
```diff
@@ -141,6 +141,15 @@ export default class ConversionToolbar extends Module {
...
+   */
+  public hasConversionTools(): boolean {
```
neSpecc: better to name `hasTools` because we have no such entity as «conversion tool»

Issue #1017, opened 2020-01-21
Maximum call stack size exceeds by setting big object to tool.config


hata6502:
When set big object to tool's config: 

<img width="276" alt="スクリーンショット 2020-01-21 19 31 44" src="https://user-images.githubusercontent.com/7702653/72797577-b946dc80-3c84-11ea-9de9-483305dbf0bb.png">

EditorJS can't be launched: 

```
Module Tools was skipped because of RangeError: Maximum call stack size exceeded
```

I think that this is resolved by fixing deepMerge of tools!
<img width="578" alt="スクリーンショット 2020-01-21 19 34 52" src="https://user-images.githubusercontent.com/7702653/72797781-20fd2780-3c85-11ea-9ed1-d3548acd12bd.png">
(src/components/modules/tools.ts  224)

I'll send a Pull Request.

hata6502:
Sorry for lack of information.
I write more detail :)

In my case, I'm making my Image block.
Apollo Client is used to get picture url.
I tried to propagate Apollo Client via tools.config.

I think deep merge should be used little, because it is high load.
Especially internalTools's nest levels are fixed, so deep merge is not needed.

hata6502:
I resolved myself!

To propagate big object via config, use callback. 

```
config: {
  bigObject: () => bigObject
}
```

PR #1018, opened 2020-01-21
Fix deep merge of tools

hata6502:
https://github.com/codex-team/editor.js/issues/1017


khaydarov:
@hata6502 hi and thanks! Can you describe the real case of passing a "very big object" to the config? I think it is better to throw an exception

src/components/modules/tools.ts
```diff
@@ -221,7 +221,30 @@ export default class Tools extends Module {
     /**
      * Assign internal tools
      */
-    this.config.tools = _.deepMerge({}, this.internalTools, this.config.tools);
+    this.config.tools = {
+      ...this.internalTools,
+      ...this.config.tools,
+      bold: {
```
neSpecc: I don't think it is good idea to hardcode internal tools here. It would better to store their names in one place

hata6502:
https://github.com/codex-team/editor.js/pull/1018#issuecomment-576671763

In my case, I tried setting ApolloClient to config. 
https://github.com/apollographql/apollo-client

hata6502:
Sorry for lack of information. 
I write more detail :)

In my case, I'm making my Image block. 
Apollo Client is used to get picture url. 
I tried to propagate Apollo Client via tools.config. 

I think deep merge should be used little, because it is high load. 
Especially internalTools's nest levels are fixed, so deep merge is not needed.

hata6502:
This PR causes ToolConfig error when launched multiple EditorJS. 
So I close it.

Issue #946, opened 2019-11-12
React: Clicks on custom plugin with react component are not propagated


natterstefan:
## Issue

When rendering a React component in the plugin's `render` method, clicks on the React element do not trigger it's `onClick`.

## Expected outcome

I can interact with the react component properly (eg. `onClick` is invoked).

## Example

An example codesandbox can be found here: https://codesandbox.io/s/solitary-night-96hzk?fontsize=14. After some more testing it looks like eg. `onChange` on `input` elements is not an issue: https://codesandbox.io/s/sweet-resonance-p7xwu.

### The React component

```jsx
import React, { useEffect, useState } from "react";
import ReactDOM from "react-dom";

export default class Custom {
  render() {
    const Button = () => {
      const [count, setCount] = useState(0);

      useEffect(() => console.log("mounting"), []);

      const onClick = e => {
        console.log("clicked"); // << is not triggered
        setCount(prevCount => prevCount + 1);
      };

      return (
        <button onClick={onClick} type="button" style={{ padding: 10 }}>
          Click me [clicked: {count} times]
        </button>
      );
    };

    const container = document.createElement("div");
    ReactDOM.render(<Button />, container);
    return container;
  }

  save(element) {
    console.log(element);
    return {
      text: element.value
    };
  }
}
```

## Environment

- React: `"react": "16.11.0"`
- Editorjs: `"@editorjs/editorjs": "2.15.1"`

## Related Issues/Topics

- https://github.com/codex-team/editor.js/issues/849#issuecomment-552821805

@gohabereg already [posted a gif illustrating the issue](https://github.com/codex-team/editor.js/issues/849#issuecomment-552821805):

![](https://user-images.githubusercontent.com/23050529/68661580-3dc20400-054c-11ea-9645-1bddbfdd7541.gif)

And I created another one. As soon as you select some text, it works:

![Screen Capture on 2019-11-12 at 11-21-23](https://user-images.githubusercontent.com/1043668/68663282-ac986080-053e-11ea-9b51-9e9755f94c5f.gif)

IzotovDenis:
Try this
```
import React, { useEffect, useState, useRef } from "react";
import ReactDOM from "react-dom";

export default class Custom {
  constructor({ api }) {
    this.api = api;
  }
  render() {
    const Button = ({ api, data }) => {
      const [count, setCount] = useState(0);
      const buttonRef = useRef();
      useEffect(() => console.log("mounting"), []);
      useEffect(
        () =>
[...]

PR #1020, opened 2020-01-24
Make click event of block propagate

hata6502:
https://github.com/codex-team/editor.js/issues/1021


neSpecc:
Thanks for the PR. Please, attach an issue that describes a problem you resolve.

natterstefan:
Is this also related to #946 @hata6502?

hata6502:
Yes, I also use React.

ryandunnewold:
Running into this issue as well. Any updates here?

dyrdzik:
The same problem on our end. We've also discovered that usually (which looks weird) buttons fire events. When having select element it does not go through. Only when you double click in block area as it makes `window.getSelection().isCollapsed` return false.

2277Dhruv:
Running into this issue as well. Are there any updates?!

hata6502:
Very sorry that I didn't update this PR.
I seem this issue doesn't cause with Vanilla JS.
So, I think it is not a bug in Editor JS.

**Demo**
Listen click event correctly. 
https://codesandbox.io/s/editor-js-propagate-click-events-w5371?fontsize=14&hidenavigation=1&theme=dark

allankikkas:
https://github.com/codex-team/editor.js/issues/1051#issuecomment-600098345

PR #1035, opened 2020-02-12
Read only mode

### LINKED ISSUE #837: Read only mode
Add the ability to enable read-only mode. We need to support initialization in read-only mode and toggling read-only mode on the fly.

**For each plugin in [list](http://github.com/editor-js):**

- [ ] Improve Tools design that can be represented as WYSIWYG

- [ ] Accept `readOnly` boolean flag in tools constructor and save it to the class's property.

- [ ] Separate event bindings and `contenteditable` attributes activating to the `toggleReadonly(status)` method. Call this method with `true` status on the render in read-only mode. Also, this method can be called with `true/false` statuses by editorjs core — it should bind/unbind events corresponded by status.

**Working with editor.js core:**

- [ ] Add `readOnly` option to the Editor Config

- [ ] Make `readonly.ts` module that will store `readOnlyEnabled` property and have `toggleReadonly(status)` public method.

- [ ] `toggleReadonly(status)` method of `readonly.ts` module will call `toggleReadonly(status)` methods of each editor module implemented this method.

- [ ] Implement `toggleReadonly(status)` method in modules that requires some readonly toggling logic: `UI`, `Toolbar`, `InlineToolbar`, `Paste`, `Saver`, `Shortcuts` (for disabling `cmd+x`), `BlockSettings`, `BlockEvents`, `BlockManager`, `DragNdrop`, `ModificationObserver`, `API`. This method will do two things: bind/unbind events and make/remove dom elements.

- [ ] If some module makes own ui and bind events in `prepare` method (for example, `UI`, `Toolbar`): move events binding and ui making in `toggleReadonly()` method of this module. Then call `toggleReadonly()` in prepare.

- [ ] Pass `readOnly` option to Tools constructor on render

- [ ] If some of Tools does not provide read-only functionality (can be detected by `toggleReadonly () implemented or not`), throw an error on trying to enable read-only mode. 

- [ ] add `toggleReadonly()` method to the public API. This method will call `toggleReadonly(status)` method of `readonly.ts` module internally.

hboylan:
Fixes #837

[Design](https://github.com/codex-team/editor.js/issues/837#issue-463226157):
**For each plugin in list:**
- [x]  Improve Tools design that can be represented as WYSIWYG

- [x]  Accept readOnly boolean flag in tools constructor and save it to the class's property.

- [x]  Separate event bindings and contenteditable attributes activating to the toggleReadonly(status) method. Call this method with true status on the render in read-only mode. Also, this method can be called with true/false statuses by editorjs core — it should bind/unbind events corresponded by status.

**Working with editor.js core:**
- [x]  Add readOnly option to the Editor Config

- [x]  Make readonly.ts module that will store readOnlyEnabled property and have toggleReadonly(status) public method.

- [x]  toggleReadonly(status) method of readonly.ts module will call toggleReadonly(status) methods of each editor module implemented this method.

- [x]  Implement toggleReadonly(status) method in modules that requires some readonly toggling logic. This method will do two things: bind/unbind events and make/remove dom elements.
  - [x] UI
  - [x] Toolbar
  - [x] InlineToolbar
  - [x] Paste
  - [x] Saver
  - [x] Shortcuts (for disabling cmd+x)
  - [x] BlockSettings
  - [x] BlockEvents
  - [x] BlockManager
  - [x] DragNDrop
  - [x] ModificationObserver
  - [x] API

- [x]  If some module makes own ui and bind events in prepare method, move events binding and ui making in toggleReadonly() method of this module. Then call toggleReadonly() in prepare.
  - [x] Tools
  - [x] UI
  - [x] BlockManager
  - [x] Paste
  - [x] DragNDrop
  - [x] ModificationsObserver
  - [x] BlockSelection
  - [x] RectangleSelection

- [x]  Pass readOnly option to Tools constructor on render

- [x]  If some of Tools does not provide read-only functionality (can be detected by toggleReadonly () implemented or not), throw an error on trying to enable read-only mode.

- [x]  add toggleReadonly() method to the public API. This method will call toggleReadonly(status) method of readonly.ts module internally.


hboylan:
Just getting the ball rolling. Will continue adding the toggle method to modules that need it and updating their prepare methods.

neSpecc:
@hboylan great

src/components/modules/ui.ts
```diff
@@ -132,45 +132,58 @@ export default class UI extends Module {
...
-     * Bind events for the UI elements
-     */
-    await this.bindEvents();
+  /**
+   * Set or toggle read-only state
+   */
+  public toggleReadOnly(readOnlyEnabled: boolean) {
+    if (readOnlyEnabled) {
+      // Do nothing
+    } else {
```
khaydarov: not all the function calls inside the UI prepare method must be in toggleReadOnly, but those modules that won't be used

types/api/readonly.d.ts
```diff
@@ -0,0 +1,12 @@
...
+ * ReadOnly API
+ */
+export interface ReadOnly {
+  /**
+   * Set or toggle read-only state
+   *
+   * @param {Boolean|undefined} state - set or toggle state
+   * @returns {Boolean} current value
+   */
+  toggleReadOnly: (state?: boolean) => boolean;
```
khaydarov: it is better to set by default false rather than undefined

khaydarov:
@hboylan Hi and welcome! :)

types/api/readonly.d.ts
hboylan: My goal is to toggle the current value when state is undefined, then return the new value. Does this make sense?
https://github.com/codex-team/editor.js/pull/1035/files#diff-f1ccc28f8e2fb341fa79ad3ac36a0a16R36

src/components/modules/ui.ts
hboylan: After some experimenting, it looks like `bindEvents` is the only one that must be toggle on/off. We'll need to add an `unbindEvents` method.

types/api/readonly.d.ts
khaydarov: Oh, I got it. Yes, it won't make sense, when clicking twice with the false value :)

src/components/modules/rectangleSelection.ts
```diff
@@ -96,33 +96,47 @@ export default class RectangleSelection extends Module {
...
-
-    Listeners.on(window, 'scroll', (event) => {
-      this.changingRectangle(event);
-    }, false);
-
-    Listeners.on(document.body, 'mouseup', () => {
-      this.endSelection();
-    }, false);
+    if (readOnlyEnabled) {
+      Listeners.removeAll();
```
neSpecc: it will remove all listeners, not only from this module

src/components/modules/readonly.ts
```diff
@@ -0,0 +1,51 @@
...
+  public async prepare(): Promise<void> {
+    this.readOnlyEnabled = this.config.readOnly;
+  }
+
+  /**
+   * Set read-only mode or toggle current state
+   *
+   * @param {Boolean} state - (optional) read-only state or toggle
+   */
+  public toggleReadOnly(state: boolean = !this.readOnlyEnabled) {
```
neSpecc: better change this name, but leave `toggleReadOnly` in modules instead of `setReadOnly`.

neSpecc: I would suggest to name this method just `toggle` because it is already in ReadOnly class.

src/components/modules/shortcuts.ts
```diff
@@ -62,4 +62,11 @@ export default class Shortcuts extends Module {
     this.registeredShortcuts[index].remove();
     this.registeredShortcuts.splice(index, 1);
   }
+
+  /**
+   * Remove all shortcuts
+   */
+  public removeAll() {
+    this.registeredShortcuts = [];
```
neSpecc: to remove shortcut, you can call `shortcut.remove()`

see https://github.com/codex-team/codex.shortcuts

src/components/modules/ui.ts
```diff
@@ -150,27 +152,50 @@ export default class UI extends Module {
...
      * Append SVG sprite
      */
-    await this.appendSVGSprite();
+    this.appendSVGSprite();
 
     /**
      * Make toolbar
      */
-    await this.Editor.Toolbar.make();
+    this.Editor.Toolbar.make();
```
neSpecc: making a toolbar also can be placed at toggleReadOnly — we don't need it in read-only mode.

src/components/modules/ui.ts
```diff
@@ -150,27 +152,50 @@ export default class UI extends Module {
...
 
     /**
      * Make the Inline toolbar
      */
-    await this.Editor.InlineToolbar.make();
+    this.Editor.InlineToolbar.make();
```
neSpecc: and this.

src/components/modules/rectangleSelection.ts
hboylan: I thought it was simpler to call `removeAll` considering each module's listener events aren't being added in a way that makes it easy to remove them. Should any listeners remain active in read-only mode?

https://github.com/hboylan/editor.js/blob/7b1cffe69516f33ec3313536fa8619fb55d26f4e/src/components/modules/dragNDrop.ts#L41-L53

src/components/modules/shortcuts.ts
hboylan: Same line of reasoning as Listeners. Should any shortcuts remain active in read-only mode?

src/components/modules/ui.ts
hboylan: I thought the same, but I'm not sure how to make and un-make the toolbars considering read-only mode can be toggled. I figured it was best to just make them, and call `closeAllToolbars` when `readOnlyEnabled === true`

src/components/modules/shortcuts.ts
neSpecc: yes, for example cmd+c, cmd+a

calumk:
Very much looking forward to this

types/tools/block-tool.d.ts
```diff
@@ -16,6 +16,11 @@ export interface BlockTool extends BaseTool {
    */
   sanitize?: SanitizerConfig;
 
+  /**
+   * @param {boolean} readOnly - render HTML on readonly mode
+   */
+  render(readOnly?: boolean): HTMLElement;
```
neSpecc, line 22: Maybe it would better to pass this flag through the constructor?

types/index.d.ts
```diff
@@ -17,7 +17,7 @@ import {
   Events,
   InlineToolbar,
   Listeners,
-  Notifier,
+  Notifier, ReadOnly,
```
neSpecc, line 20: + add export for this type

src/components/utils.ts
```diff
@@ -556,3 +556,13 @@ export function getValidUrl(url: string): string {
...
 }
+
+/**
+ * Returns random generated identifier
+ *
+ * @returns {string}
+ */
+export function generateId(): string {
+  // tslint:disable-next-line:no-bitwise
+  return `f${(~~(Math.random() * 1e8)).toString(16)}`;
```
neSpecc, line 567: if `f` is a prefix, it should be added outside this method

src/components/modules/shortcuts.ts
```diff
@@ -64,6 +64,10 @@ export default class Shortcuts extends Module {
   public remove(shortcut: string): void {
     const index = this.registeredShortcuts.findIndex((shc) => shc.name === shortcut);
 
+    if (!this.registeredShortcuts[index]) {
```
neSpecc, line 67:
```suggestion
    if (index === -1 || !this.registeredShortcuts[index]) {
```

src/components/modules/renderer.ts
```diff
@@ -44,11 +42,12 @@ export default class Renderer extends Module {
    * Make plugin blocks from array of plugin`s data
    *
    * @param {BlockToolData[]} blocks - blocks to render
+   * @param {boolean} readOnly - readOnly flag value
    */
-  public async render(blocks: BlockToolData[]): Promise<void> {
-    const chainData = blocks.map((block) => ({ function: (): Promise<void> => this.insertBlock(block) }));
+  public async render(blocks: BlockToolData[], readOnly: boolean): Promise<void> {
```
neSpecc, line 47: why not to get the `readOnly` state from the public property of ReadOnly module? Then you won't need to pass `readOnly` in every method.

src/components/modules/readonly.ts
```diff
@@ -0,0 +1,65 @@
...
+ * @typedef {ReadOnly} ReadOnly
+ * @property {boolean} readOnlyEnabled - read-only state
+ */
+export default class ReadOnly extends Module {
+  /**
+   * Value to track read-only state
+   *
+   * @type {boolean}
+   */
+  private readOnlyEnabled = false;
```
neSpecc, line 20: we can make it public to allow access it from the different methods

src/components/modules/modificationsObserver.ts
```diff
@@ -64,12 +65,26 @@ export default class ModificationsObserver extends Module {
...
+  }
+
+  /**
+   * Set read-only state
+   *
+   * @param {boolean} readOnlyEnabled
+   */
+  public toggleReadOnly(readOnlyEnabled: boolean) {
+    if (readOnlyEnabled) {
+      this.disable();
```
neSpecc, line 78: also, there should be a negative version of `setObserver`?

src/components/modules/listeners.ts
```diff
@@ -204,4 +231,19 @@ export default class Listeners extends Module {
...
+
+  /**
+   * Returns listener data found by id
+   *
+   * @param {string} id - listener identifier
+   *
+   * @returns {ListenerData}
+   */
+  private findById(id: string): ListenerData {
+    for (const listener of this.allListeners) {
```
neSpecc, line 243: why not to use `Array.find()`?

src/components/modules/dragNDrop.ts
```diff
@@ -13,39 +13,80 @@ export default class DragNDrop extends Module {
...
+    this.listenerIds = [];
   }
 
   /**
    * Handle drop event
    *
    * @param {DragEvent} dropEvent - drop event
    */
-  private processDrop = async (dropEvent: DragEvent): Promise<void> => {
+  private async processDrop(dropEvent: DragEvent): Promise<void> {
```
neSpecc, line 89: need to test this

src/components/modules/blockSelection.ts
```diff
@@ -130,30 +119,55 @@ export default class BlockSelection extends Module {
...
+   *
+   * @param {boolean} readOnlyEnabled - "read only" state
+   */
+  public toggleReadOnly(readOnlyEnabled: boolean): void {
+    const { Shortcuts } = this.Editor;
 
-    this.selection = new SelectionUtils();
+    if (readOnlyEnabled) {
+      Shortcuts.remove('CMD+A');
+      this.selection = null;
```
neSpecc, line 143: also, need to unselect all blocks

src/components/modules/blockManager.ts
```diff
@@ -157,6 +157,16 @@ export default class BlockManager extends Module {
    */
   private _blocks: Blocks = null;
 
+  /**
+   * Read only flag
+   */
+  private readOnlyEnabled = false;
```
neSpecc, line 163: we can get if from the single place - public property of the ReadOnly method

src/components/modules/blockManager.ts
```diff
@@ -207,15 +228,23 @@ export default class BlockManager extends Module {
...
+    settings: ToolConfig = {},
+    readOnly = false,
+  ): Block {
     const toolInstance = this.Editor.Tools.construct(toolName, data) as BlockTool;
     const toolClass = this.Editor.Tools.available[toolName] as BlockToolConstructable;
-    const block = new Block(toolName, toolInstance, toolClass, settings, this.Editor.API);
+    const block = new Block(toolName, toolInstance, toolClass, settings, this.Editor.API, readOnly);
 
-    this.bindEvents(block);
+    if (!this.readOnlyEnabled) {
```
neSpecc, line 245: also, can be accessed from the readonly module

src/components/modules/blockManager.ts
```diff
@@ -649,11 +680,107 @@ export default class BlockManager extends Module {
   private bindEvents(block: Block): void {
```
neSpecc, line 680: better to change name for `bindEventToBlock` because there is a `enableModuleEvents ` with the same meaning

src/components/modules/blockManager.ts
```diff
@@ -649,11 +680,107 @@ export default class BlockManager extends Module {
...
+   *  - Enable shortcuts again
+   *  - Restore `copy` and `cut` bindings
+   *  - Bind all events handlers for all Blocks
+   */
+  private enableModuleEvents(): void {
+    const { Listeners, Shortcuts, BlockEvents } = this.Editor;
+
+    /** Copy shortcut */
+    Shortcuts.add({
+      name: 'CMD+C',
```
neSpecc, line 749: need to check is this shourtcuts (cmd+c, cmd+x) is steel used.

src/components/modules/ui.ts
```diff
@@ -155,27 +160,54 @@ export default class UI extends Module {
...
+     * Prepare with read-only state from config
+     */
+    this.toggleReadOnly(this.config.readOnly);
+  }
 
+  /**
+   * Toggle read-only state
+   *
+   * If readOnly is true:
+   *  - bb
```
neSpecc, line 190: missed comments

src/components/modules/blockManager.ts
```diff
@@ -649,11 +680,107 @@ export default class BlockManager extends Module {
...
+  private disableModuleEvents(): void {
+    const { Listeners, Shortcuts } = this.Editor;
+
+    for (const id of this.listenerIds) {
+      Listeners.offById(id);
+    }
+
+    this.listenerIds = [];
+
+    Shortcuts.remove('CMD+C');
```
neSpecc, line 733: cmd+c should work in readonly-mode

src/components/modules/blockManager.ts
```diff
@@ -649,11 +680,107 @@ export default class BlockManager extends Module {
...
+      handler: (event) => {
+        BlockEvents.handleCommandX(event);
+      },
+    });
+
+    /** Copy event */
+    this.listenerIds.push(
+      Listeners.on(
+        document,
+        'copy',
```
neSpecc, line 767: should work in readonly

types/tools/block-tool.d.ts
khaydarov: yeah, definitely

src/components/modules/readonly.ts
khaydarov: I implemented `isEnabled()` method that returns its value

src/components/modules/toolbar/index.ts
```diff
@@ -320,13 +320,13 @@ export default class Toolbar extends Module {
...
      */
-    this.Editor.Listeners.on(this.nodes.settingsToggler, 'click', () => this.settingsTogglerClicked());
+    this.Editor.Listeners.on(this.nodes.settingsToggler, 'click', this.settingsTogglerClicked);
   }
 
   /**
    * Clicks on the Block Settings toggler
    */
-  private settingsTogglerClicked(): void {
+  private settingsTogglerClicked = (): void => {
```
khaydarov, line 329: restore, use lambda and delete via id

src/components/modules/toolbar/index.ts
```diff
@@ -308,7 +308,7 @@ export default class Toolbar extends Module {
   /**
    * Handler for Plus Button
    */
-  private plusButtonClicked(): void {
+  private plusButtonClicked = (): void => {
```
khaydarov, line 311: restore, use lambda and delete via id

docs/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 - `Fix` - Fixed wrong toolbar icon centering in Firefox [#1120](https://github.com/codex-team/editor.js/pull/1120)
 - `Improvements` - Improve performance of DOM traversing at the `isEmpty()` method [#1095](https://github.com/codex-team/editor.js/issues/1095)
 - `Fix` - Toolbox: Tool's order in Toolbox now saved in accordance with `tools` object keys order [#1073](https://github.com/codex-team/editor.js/issues/1073)
+- `New` - Read only mode [#837](https://github.com/codex-team/editor.js/issues/837)
```
neSpecc, line 14:
```suggestion
- `New` - Read-only mode 🥳 [#837](https://github.com/codex-team/editor.js/issues/837)
```

src/components/block.ts
```diff
@@ -154,20 +154,20 @@ export default class Block {
    * @param {object} toolInstance — passed Tool`s instance that rendered the Block
    * @param {object} toolClass — Tool's class
    * @param {object} settings - default settings
-   * @param {ApiModule} apiModule - Editor API module for pass it to the Block Tunes
+   * @param {ApiModule} apiMethods - Editor API module for pass it to the Block Tunes
```
neSpecc, line 157: this is an API module, not methods

src/components/modules/readonly.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+   * Returns state of read only mode
+   */
+  public isEnabled(): boolean {
```
neSpecc, line 25: it can be a getter

src/components/modules/blockSelection.ts
```diff
@@ -402,4 +386,49 @@ export default class BlockSelection extends Module {
...
+    });
+  }
+
+  /**
+   * Unbinds module events
+   */
+  private disableModuleBindings(): void {
+    const { Shortcuts } = this.Editor;
+
+    Shortcuts.remove('CMD+A');
```
neSpecc, line 427: cmd+a should work in read-only

src/components/modules/blockManager.ts
```diff
@@ -647,18 +681,65 @@ export default class BlockManager extends Module {
...
+   * Enables all module handlers and bindings
+   * The sequence is following:
+   *  - Enable shortcuts again
+   *  - Restore `copy` and `cut` bindings
+   *  - Bind all events handlers for all Blocks
+   */
+  private enableModuleBindings(): void {
+    this.blocks.forEach((block: Block) => {
+      this.bindBlockEvents(block);
+    });
```
gohabereg, line 742: You can make `bindBlockEvents` arrow function and write here like this:

```suggestion
    this.blocks.forEach(this.bindBlockEvents);
```

src/components/modules/blockManager.ts
```diff
@@ -647,18 +681,65 @@ export default class BlockManager extends Module {
...
+   * The sequence is following:
+   *  - Removes all listeners by id
+   *  - Removes all shortcuts
+   */
+  private disableModuleBindings(): void {
+    const { Listeners } = this.Editor;
+
+    for (const id of this.listenerIds) {
+      Listeners.offById(id);
+    }
```
gohabereg, line 727: this.listener
```suggestion
    this.listenersIds.forEach(Listeners.offById);
```

src/components/modules/crossBlockSelection.ts
```diff
@@ -44,6 +57,22 @@ export default class CrossBlockSelection extends Module {
       !!this.lastSelectedBlock;
   }
 
+  /**
+   * Enables Cross Block Selection
+   *
+   * @param {MouseEvent} event - mouse down event
+   */
+  private enableCrossBlockSelection(event: MouseEvent): void {
```
gohabereg, line 65: I don't think `enableCrossBlockSelection` is a good name here but I can't come up with a better one at the moment

src/components/modules/dragNDrop.ts
```diff
@@ -13,39 +13,82 @@ export default class DragNDrop extends Module {
    */
   private isStartedAtEditor = false;
 
+  /**
+   * Listener identifiers
+   */
+  private listenerIds: string[] = [];
```
gohabereg, line 19: It seems we can move this to base `Module` class and add there `listeners` property as a wrapper for `Listeners` module methods as it doesn't use anything from `this.Editor`

```
class Module {
  private listenersIds: string[] = [];

  protected listeners = {
    on(...) {
      this.listenersId.push(Listeners.on(...))
    },
    off() {},
    clearAll() {},
    ...
  }
}
```

src/components/modules/dragNDrop.ts
```diff
@@ -13,39 +13,82 @@ export default class DragNDrop extends Module {
...
+        this.processDragStart();
+      })
+    );
 
-      this.Editor.InlineToolbar.close();
-    });
+    /**
+     * Prevent default browser behavior to allow drop on non-contenteditable elements
+     */
+    this.listenerIds.push(
```
gohabereg, line 68: Again why not 

```
this.listenerIds.push(
  Listeners.on(...),
  Listeners.on(...),
)
```

src/components/modules/dragNDrop.ts
```diff
@@ -13,39 +13,82 @@ export default class DragNDrop extends Module {
...
+      this.Editor.Listeners.on(this.Editor.UI.nodes.holder, 'dragover', (dragEvent: DragEvent) => {
```
gohabereg, line 69: Better to rewrite handlers to arrow functions. Then arrow function is not needed here

src/components/modules/paste.ts
```diff
@@ -145,8 +145,27 @@ export default class Paste extends Module {
...
+      this.processTools();
+      this.setCallback();
+    } else {
+      this.unsetCallback();
+
+      this.toolsTags = {};
+      this.tagsByTool = {};
+      this.toolsPatterns = [];
+      this.toolsFiles = {};
+      this.exceptionList = [];
```
gohabereg, line 167: Do we really need to empty all of these properties?

Maybe it's better to add cache check in the `processTool` method

src/components/modules/rectangleSelection.ts
```diff
@@ -196,6 +175,99 @@ export default class RectangleSelection extends Module {
...
+        this.processScroll(mouseEvent);
+      }, false)
+    );
+
+    this.listenerIds.push(
+      Listeners.on(document.body, 'mouseup', () => {
+        this.processMouseUp();
+      }, false)
+    );
+  }
```
gohabereg, line 214: And here you can use push method only once

gohabereg: And rewrite handlers to arrow functions

src/components/modules/ui.ts
```diff
@@ -307,39 +340,64 @@ export default class UI extends Module {
...
+        this.selectionChanged(event);
+      }, true)
+    );
 
-    this.Editor.Listeners.on(window, 'resize', () => {
-      this.resizeDebouncer();
-    }, {
-      passive: true,
-    });
+    this.listenerIds.push(
```
gohabereg, line 383: The same comments as above

src/components/modules/ui.ts
```diff
@@ -307,39 +340,64 @@ export default class UI extends Module {
...
+    );
+  }
+
+  /**
+   * Unbind events on the Editor.js interface
+   */
+  private disableModuleBindings(): void {
+    for (const id of this.listenerIds) {
+      this.Editor.Listeners.offById(id);
+    }
```
gohabereg, line 398:
```suggestion
    this.listenerIds.forEach(this.Editor.Listeners.offById);
```

src/components/modules/rectangleSelection.ts
```diff
@@ -196,6 +175,99 @@ export default class RectangleSelection extends Module {
...
+
+  /**
+   * Removes Modules bindings
+   */
+  private disableModuleBindings(): void {
+    for (const id of this.listenerIds) {
+      this.Editor.Listeners.offById(id);
+    }
```
gohabereg, line 222:
```suggestion
    this.listenerIds.forEach(this.Editor.Listeners.offById);
```

src/components/modules/blockManager.ts
khaydarov: what is the difference?

src/components/modules/blockManager.ts
```diff
@@ -649,11 +680,107 @@ export default class BlockManager extends Module {
...
+      }),
+    );
+
+    this.listenerIds.push(
+      Listeners.on(block.holder, 'dragover', (event: DragEvent) => {
+        BlockEvents.dragOver(event);
+      }),
+    );
+
+    this.listenerIds.push(
```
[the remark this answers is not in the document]
khaydarov: it will look better maybe, I'll try it on

src/components/modules/blockManager.ts
khaydarov: It is not good to make some methods as arrows function and some not don't you think so? we need to have an agreement that if it is a class method so they won't be arrow function, but for lambdas or nested functions it is very useful to use them instead of `.bind()`

src/components/modules/blockManager.ts
khaydarov: I think it is more readable with
```js
for (const id of this.listenerIds) {
      Listeners.offById(id);
}

this.listenerIds = [];
```

than

```js
this.listenersIds.forEach(Listeners.offById);
this.listenersIds = [];
```

src/components/modules/dragNDrop.ts
khaydarov: we can discuss it

src/components/modules/dragNDrop.ts
khaydarov: it is a good idea. Maybe

src/components/modules/paste.ts
khaydarov: fixed

src/components/modules/blockManager.ts
gohabereg: Writing class methods that are event handlers or used in map array methods as arrow functions is a common practice.
The advantages are important: 
1. 
```
Listeners.on(element, this.bindEvent) //more readable than:
Listeners.on(element, (e) => {
  this.bindEvent(e)
});
```

2. If you pass an arrow function as event handler in-place, you can't remove this handler as function is anonymous.
Moreover, if for some reason, `addEventListener` is called twice (or more), there will be 2 same handlers on one event. Whereas, if you pass named arrow function (or arrow class method), user agent won't add two listeners and you would be able to remove handler by passing the same function object to `removeEventListener`.,

src/components/modules/blockManager.ts
```diff
@@ -647,18 +677,52 @@ export default class BlockManager extends Module {
...
-  private bindEvents(block: Block): void {
+  private bindBlockEvents(block: Block): void {
     const { BlockEvents, Listeners } = this.Editor;
 
-    Listeners.on(block.holder, 'keydown', (event) => BlockEvents.keydown(event as KeyboardEvent), true);
-    Listeners.on(block.holder, 'mousedown', (event: MouseEvent) => BlockEvents.mouseDown(event));
-    Listeners.on(block.holder, 'keyup', (event) => BlockEvents.keyup(event));
-    Listeners.on(block.holder, 'dragover', (event) => BlockEvents.dragOver(event as DragEvent));
-    Listeners.on(block.holder, 'dragleave', (event) => BlockEvents.dragLeave(event as DragEvent));
+    this.listenerIds.push(
```
neSpecc, line 687: if you remove `mousedown` listener, you should remove `mouseDown` callback from BlockEvents

src/components/modules/crossBlockSelection.ts
```diff
@@ -17,6 +17,19 @@ export default class CrossBlockSelection extends Module {
...
 
+  /**
+   * Module preparation
+   *
+   * @returns {Promise}
+   */
+  public async prepare(): Promise<void> {
+    const { Listeners } = this.Editor;
+
+    Listeners.on(document, 'mousedown', (event: MouseEvent) => {
```
neSpecc, line 28: are you sure that this event should belong to the document, instead of editor wrapper?

src/components/modules/tools.ts
```diff
@@ -364,7 +365,12 @@ export default class Tools extends Module {
    *
    * @returns {InlineTool} — instance
    */
-  public constructInline(tool: InlineToolConstructable, name: string, toolSettings: ToolSettings = {} as ToolSettings): InlineTool {
+  // tslint:disable-next-line:max-line-length
```
neSpecc, line 368: we do not use tslint

src/components/modules/ui.ts
```diff
@@ -155,27 +161,57 @@ export default class UI extends Module {
...
      * Append SVG sprite
      */
-    await this.appendSVGSprite();
+    this.appendSVGSprite();
 
     /**
      * Make toolbar
      */
-    await this.Editor.Toolbar.make();
+    this.Editor.Toolbar.make();
```
neSpecc, line 169: still actual

src/components/modules/crossBlockSelection.ts
khaydarov: yes, I've tested and it didn't work on editor's wrapper. Maybe @gohabereg could help

src/components/modules/readonly.ts
```diff
@@ -0,0 +1,69 @@
...
+      /**
+       * set or toggle read-only state
+       */
+      this.Editor[name].toggleReadOnly(state);
+    }
+
+    /**
+     * Save current Editor Blocks and render again
+     */
+    const savedBlocks = await this.Editor.Saver.save();
```
neSpecc, line 62: it would better to put the re-rendering part to the separated method

src/components/__module.ts
```diff
@@ -12,6 +12,11 @@ import { ModuleConfig } from '../types-internal/module-config';
  * @property {EditorModules} Editor - List of Editor modules
  */
 export default class Module {
+  /**
+   * Module memorizes HTMLElements
+   */
+  public nodes: any;
```
neSpecc, line 18: then, you should remove `public nodes(){}` from the modules? But it this case, it would not be clear of which nodes the module has 🤔

For example, for now, BlockSettings specify own nodes like this:

image
— it is clear structure.


We need to think about it.

src/components/modules/toolbar/blockSettings.ts
```diff
@@ -103,6 +103,15 @@ export default class BlockSettings extends Module {
     this.enableFlipper();
   }
 
+  /**
+   * Destroys module
+   */
+  public destroy(): void {
+    this.flipper.deactivate();
+    this.flipper = null;
+    this.removeAllNodes();
```
neSpecc, line 112: this module still has its own `nodes` property.

src/components/modules/toolbar/conversion.ts
```diff
@@ -88,6 +88,15 @@ export default class ConversionToolbar extends Module {
     return this.nodes.wrapper;
   }
 
+  public destroy(): void {
```
neSpecc, line 91: docs are missed

src/components/modules/toolbar/conversion.ts
```diff
@@ -88,6 +88,15 @@ export default class ConversionToolbar extends Module {
...
+    this.flipper.deactivate();
+    this.flipper = null;
+
+    for (const key in this.nodes) {
+      const node = this.nodes[key];
```
neSpecc, line 96: what is it?

src/components/modules/toolbar/index.ts
```diff
@@ -57,22 +57,6 @@ import { I18nInternalNS } from '../../i18n/namespace-internal';
  * @property {Element} nodes.defaultSettings   - Default Settings section of Settings Panel
  */
 export default class Toolbar extends Module {
-  /**
-   * HTML Elements used for Toolbar UI
-   */
-  public nodes: {[key: string]: HTMLElement} = {
```
neSpecc, line 63: maybe we can create an interface that will represent the nodes' structure of the module?

src/components/modules/toolbar/index.ts
```diff
@@ -99,89 +83,92 @@ export default class Toolbar extends Module {
   }
 
   /**
-   * Makes toolbar
+   * returns toolbar opened state
```
neSpecc, line 86:
```suggestion
   * Returns the Toolbar opening state
```

src/components/__module.ts
neSpecc: I see, that you have already remove the `nodes` from the modules, except those which are included at other modules. I will think about it.

types/api/blocks.d.ts
```diff
@@ -13,10 +13,13 @@ export interface Blocks {
 
   /**
    * Render passed data
-   * @param {OutputData} data
-   * @return {Promise<void>}
+   *
+   * @param {OutputData} data - saved Block data
+   * @param {boolean} readOnly - read only flag
```
neSpecc, line 18:
```suggestion
   * @param {boolean} readOnly - the flag that should be used to render a block in the read-only mode
```

src/components/modules/blockEvents.ts
```diff
@@ -107,21 +107,6 @@ export default class BlockEvents extends Module {
     this.Editor.UI.checkEmptiness();
   }
 
-  /**
-   * Set up mouse selection handlers
-   *
-   * @param {MouseEvent} event - mouse down event
-   */
-  public mouseDown(event: MouseEvent): void {
```
neSpecc, line 115: where is this handler?

khaydarov: it is removed in 2.18

src/components/modules/toolbar/index.ts
khaydarov: @gohabereg any suggestion?

src/components/modules/toolbar/conversion.ts
khaydarov: yeah, I removed this block of code and used removeAllNodes()

ageddesi:
How close is this feature to being done, I am waiting on this to launch a piece of software. I want to use your version instead of hacking in my own. :)

neSpecc:
> How close is this feature to being done, I am waiting on this to launch a piece of software. I want to use your version instead of hacking in my own. :)

It will be shipped with the upcoming 2.19 release.

anuragk15:
Is the feature live? I tired passing readOnly but nothing happened.

src/components/utils.ts
```diff
@@ -558,6 +558,18 @@ export function openTab(url: string): void {
...
+/**
+ * Returns random generated identifier
+ *
+ * @param {string} prefix - identifier prefix
+ *
+ * @returns {string}
+ */
+export function generateId(prefix = ''): string {
+  // tslint:disable-next-line:no-bitwise
+  return `${prefix}${(~~(Math.random() * 1e8)).toString(16)}`;
```
gohabereg, line 570: Better to use `Math.floor`

gohabereg (approved):

neSpecc (changes requested):
I can modify editor content (add or remove blocks, and so on) in readony mode through the BlocksAPI

try `editor.blocks.clear()`

We need to disable such methods in the read-only mode. Add warnings to them.

src/components/modules/crossBlockSelection.ts
```diff
@@ -34,6 +47,7 @@ export default class CrossBlockSelection extends Module {
 
     Listeners.on(document, 'mouseover', this.onMouseOver);
     Listeners.on(document, 'mouseup', this.onMouseUp);
+    Listeners.on(document, 'selectionchange', this.onSelectionChange);
```
neSpecc, line 50: we already have such a handler in the UI module, you can use it.

src/components/modules/crossBlockSelection.ts
```diff
@@ -173,6 +204,17 @@ export default class CrossBlockSelection extends Module {
     this.lastSelectedBlock = targetBlock;
   }
 
+  /**
+   * Removes all ranges when any Block is selected
+   */
+  private onSelectionChange = (): void => {
+    const { BlockSelection } = this.Editor;
+
+    if (BlockSelection.anyBlockSelected) {
```
neSpecc, line 213: it will iterate through all the blocks on almost every mousemove firing. Maybe we can optimize it?

src/components/modules/readonly.ts
```diff
@@ -0,0 +1,104 @@
...
+    const toolsDontSupportReadOnly: string[] = [];
+
+    Object.entries(blockTools).forEach(([name, tool]) => {
+      if (!Tools.isReadOnlySupported(tool)) {
+        toolsDontSupportReadOnly.push(name);
+      }
+    });
+
+    if (toolsDontSupportReadOnly.length > 0) {
+      _.logLabeled(
```
neSpecc, line 94: for now, this log is displaying event if I do only initializing editor in a regular mode, on the first load.

It should be displayed only when initializing in read-only mode or when using 'toggle' method.

src/components/modules/readonly.ts
```diff
@@ -0,0 +1,104 @@
...
+  public get isEnabled(): boolean {
+    return this.readOnlyEnabled;
+  }
+
+  /**
+   * Set initial state
+   */
+  public async prepare(): Promise<void> {
+    if (!this.isToolsReadOnlyCompatible()) {
+      return;
```
neSpecc, line 35: also, the editor should not be initialized if readOnly enabled, but some tool does not support it. I mean it should throw an error and do not render anything.

src/components/modules/toolbar/blockSettings.ts
```diff
@@ -60,7 +60,7 @@ export default class BlockSettings extends Module {
   /**
    * Block settings UI HTML elements
    */
-  public nodes: {[key: string]: HTMLElement} = {
+  public nodes = {
```
neSpecc, line 63: why not use the new `ModuleNodes` type in generic instead of declaring nodes here?

src/components/modules/toolbar/conversion.ts
```diff
@@ -32,7 +32,7 @@ export default class ConversionToolbar extends Module {
   /**
    * HTML Elements used for UI
    */
-  public nodes: { [key: string]: HTMLElement } = {
+  public nodes = {
```
neSpecc, line 35: same here

src/components/modules/toolbar/index.ts
neSpecc: so it should be moved to the ModuleNodes

src/components/modules/toolbar/inline.ts
```diff
@@ -45,13 +45,7 @@ export default class InlineToolbar extends Module {
...
    * Inline Toolbar elements
    */
-  private nodes: {
-    wrapper: HTMLElement;
-    buttons: HTMLElement;
-    conversionToggler: HTMLElement;
-    conversionTogglerContent: HTMLElement;
-    actions: HTMLElement;
-  } = {
+  public nodes = {
```
neSpecc, line 48: same here

src/components/modules/toolbar/toolbox.ts
```diff
@@ -55,10 +55,7 @@ export default class Toolbox extends Module {
   /**
    * HTMLElements used for Toolbox UI
    */
-  public nodes: {
-    toolbox: HTMLElement;
-    buttons: HTMLElement[];
-  } = {
+  public nodes = {
```
neSpecc, line 58: same here

src/components/modules/ui.ts
```diff
@@ -93,10 +93,11 @@ export default class UI extends Module {
   /**
    * HTML Elements used for UI
    */
-  public nodes: { [key: string]: HTMLElement } = {
+  public nodes = {
```
neSpecc, line 96: same here

types/index.d.ts
```diff
@@ -88,6 +91,7 @@ export interface API {
   inlineToolbar: InlineToolbar;
   tooltip: Tooltip;
   i18n: I18n;
+  readonly: ReadOnly;
```
neSpecc, line 94: readOnly

types/index.d.ts
```diff
@@ -108,6 +112,7 @@ declare class EditorJS {
   public styles: Styles;
   public toolbar: Toolbar;
   public inlineToolbar: InlineToolbar;
+  public readonly: ReadOnly;
```
neSpecc, line 115: readOnly

neSpecc (changes requested):

src/components/__module.ts
```diff
@@ -37,6 +38,11 @@ export default class Module<T extends ModuleNodes = {}> {
    */
   protected config: EditorConfig;
 
+  /**
+   * Method names that must be decorated
+   */
+  protected decorateList: string[] = [];
```
neSpecc, line 44:
```suggestion
  protected methodsToDisableInReadonly: string[] = [];
```

neSpecc: I think we should make another abstract module called `ApiModule extends Module`. All the API modules will extend the `ApiModule`. And the `methodsToDisableInReadonly` should be at `ApiModule` because it is not related to other editor modules.

src/components/modules/api/blocks.ts
```diff
@@ -10,13 +10,31 @@ import BlockAPI from '../../block/api';
  * provides with methods working with Block
  */
 export default class BlocksAPI extends Module {
+  /**
+   * Method names that must be decorated
```
neSpecc, line 14:
```suggestion
   * Method names that should be disabled in the Read-Only mode
```

src/components/modules/api/blocks.ts
```diff
@@ -30,6 +48,14 @@ export default class BlocksAPI extends Module {
       insertNewBlock: (): void => this.insertNewBlock(),
       insert: this.insert,
     };
+
+    for (const method in methods) {
```
neSpecc, line 52: leave a comment about this feature

src/components/modules/api/blocks.ts
```diff
@@ -10,13 +10,31 @@ import BlockAPI from '../../block/api';
...
+   * Method names that must be decorated
+   */
+  protected decorateList: string[] = [
+    'clear',
+    'render',
+    'renderFromHTML',
+    'delete',
+    'swap',
+    'move',
+    'getBlockByIndex',
```
neSpecc, line 23: this method can work in read-only

src/components/modules/api/blocks.ts
```diff
@@ -10,13 +10,31 @@ import BlockAPI from '../../block/api';
...
+    'getCurrentBlockIndex',
+    'getBlocksCount',
```
neSpecc, line 25: this method can work in read-only

src/components/modules/api/listeners.ts
```diff
@@ -6,16 +6,32 @@ import { Listeners } from '../../../../types/api';
  * Provides with methods working with DOM Listener
  */
 export default class ListenersAPI extends Module {
+  /**
+   * Method names that must be decorated
+   */
+  protected decorateList: string[] = [
+    'on',
```
neSpecc, line 13: all these methods can work in read-only

src/components/modules/api/tooltip.ts
```diff
@@ -7,11 +7,20 @@ import { TooltipContent, TooltipOptions } from 'codex-tooltip';
  * @classdesc Tooltip API
  */
 export default class TooltipAPI extends Module {
+  /**
+   * Method names that must be decorated
+   */
+  protected decorateList: string[] = [
+    'show',
```
neSpecc, line 14: all these methods can work in read-only

src/components/modules/readonly.ts
```diff
@@ -83,4 +84,19 @@ export default class ReadOnly extends Module {
 
     return this.readOnlyEnabled;
   }
+
+  /**
+   * @param {Function} method - any method
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  public decorator(method: Function): any {
```
neSpecc, line 92:
```suggestion
  public offDecorator(method: Function): unknown {
```

src/components/modules/readonly.ts
```diff
@@ -83,4 +84,19 @@ export default class ReadOnly extends Module {
 
     return this.readOnlyEnabled;
   }
+
+  /**
+   * @param {Function} method - any method
```
neSpecc, line 89: add description

src/components/modules/readonly.ts
```diff
@@ -83,4 +84,19 @@ export default class ReadOnly extends Module {
...
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  public decorator(method: Function): any {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    return (...args: any[]): any => {
+      if (!this.isEnabled) {
+        return method.call(this, ...args);
+      }
+
+      _.log('Read only method is enabled, you can\' call this method', 'warn');
```
neSpecc, line 99:
```suggestion
      _.logLabeled('Read-only mode is enabled, you can\'t call ${methodGroup ? methodGroup + '.' : ''} ${methodName} method', 'warn');
```

src/components/modules/readonly.ts
neSpecc, line 92:
```suggestion
  public decorator(method: Function, methodGroup: string = ''): any {
```

where `methodGroup` is an api module name — 'blocks', 'caret', etc

src/components/modules/toolbar/toolbox.ts
```diff
@@ -26,6 +26,11 @@ interface ToolboxNodes {
  *
  */
 export default class Toolbox extends Module<ToolboxNodes> {
+  public nodes = {
```
neSpecc, line 29: describe what is it and why it is here

src/components/modules/api/blocks.ts
gohabereg: Then we need to disable Block API methods in read-only as well

src/components/__module.ts
khaydarov: Yes, I also wanted to do this way. Thanks

gohabereg:
@neSpecc @khaydarov 
It seems we need to disable some BlockAPI methods in read-only mode as well

Example:
1. Load editor
2. Toggle read-only on
3. Get block object by index and call any method:
```
const block = editor.blocks.getBlockByIndex(0);

block.stretched = true;
```

gohabereg:
Also found this bug:
1. Open editor
2. Stretch the image via tune
3. Toggle read-only
4. Image is not stretched

src/components/__module.ts
```diff
@@ -21,7 +21,8 @@ export default class Module<T extends ModuleNodes = {}> {
   /**
    * Each module can provide some UI elements that will be stored in this property
    */
-  public nodes: T;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  public nodes: T = {} as any;
```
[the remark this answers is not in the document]
neSpecc: ?

src/components/block/index.ts
```diff
@@ -51,6 +50,11 @@ interface BlockConstructorOptions {
    * Editor's API methods
    */
   api: ApiModule;
```
neSpecc, line 52: ApiModules

src/components/core.ts
```diff
@@ -300,6 +305,9 @@ export default class Core {
         try {
           await this.moduleInstances[module].prepare();
         } catch (e) {
+          if (e instanceof CriticalError) {
```
neSpecc, line 308: describe example cases pls

src/components/errors/critical.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * This type of exception fill destroy the Editor! Be careful when use it
```
neSpecc, line 2:
```suggestion
 * This type of exception will destroy the Editor! Be careful when using it
```

src/components/modules/api/base.ts
```diff
@@ -0,0 +1,11 @@
+import Module from '../../__module';
+
+/**
+ * @class BaseApiModule
+ */
+export default class BaseApiModule extends Module {
```
neSpecc, line 6:
```suggestion
export default class ApiModule extends Module {
```

src/components/modules/api/base.ts
```diff
@@ -0,0 +1,11 @@
+import Module from '../../__module';
+
+/**
+ * @class BaseApiModule
```
neSpecc, line 4: Description missed

src/components/modules/api/index.ts
```diff
@@ -17,21 +17,50 @@ export default class API extends Module {
...
+      toolbar: this.Editor.ToolbarAPI,
+      inlineToolbar: this.Editor.InlineToolbarAPI,
+      tooltip: this.Editor.TooltipAPI,
+      i18n: this.Editor.I18nAPI,
+      readOnly: this.Editor.ReadOnlyAPI,
+    };
+
+    const allMethods = {} as APIInterfaces;
+
+    for (const apiComponent in apiComponents) {
```
neSpecc, line 39: maybe we can place this code to the ApiModule?

src/components/modules/api/index.ts
```diff
@@ -50,4 +79,19 @@ export default class API extends Module {
...
   }
+
+  /**
+   * Returns decorated list of methods that have read-only restriction
+   *
+   * @param methodGroup - sss
+   * @param methods - api module methods
+   * @param list - the list of methods that should be decorated
+   */
+  private decorateWithReadOnlyFunction(methodGroup, methods, list): void {
```
neSpecc, line 90: and this

neSpecc (changes requested):

src/components/__module.ts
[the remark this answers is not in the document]
khaydarov: Without any I got the following

image

src/components/modules/api/index.ts
khaydarov: ApiModule is an abstraction for single class, so in that function I have to call decoration process in each ApiModule extension. It was easier to all for "ApiModules" on methods getter so that it is in one place

src/components/core.ts
```diff
@@ -208,6 +206,9 @@ export default class Core {
       }
     }
 
+    this.config.readOnly = this.config.readOnly as boolean || false;
+    this.config.i18n = {};
```
neSpecc, line 210: already fixed in the `next` branch, see above

src/components/modules/api/index.ts
neSpecc: can we move this method to the ApiModule class?

src/components/modules/readonly.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  /**
+   * Function wraps passed method and calls if it is not in read-only mode
+   *
+   * @param {Function} method - decorated function
+   * @param {string} methodGroup - an api module name — 'blocks', 'caret', etc
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  public offDecorator(method: Function, methodGroup = ''): any {
```
neSpecc, line 95:
```suggestion
  public offDecorator(method: Function, methodGroup = ''): unknown {
```

src/codex.ts
```diff
@@ -106,7 +106,38 @@ export default class EditorJS {
 
     this.destroy = destroy;
 
-    Object.setPrototypeOf(this, editor.moduleInstances.API.methods);
+    const apiComponents = editor.moduleInstances.API.components;
```
neSpecc, line 109: move to separate method

src/components/block/api.ts
```diff
@@ -2,15 +2,20 @@ import Block from './index';
...
 import { BlockAPI as BlockAPIInterface } from '../../../types/api';
+import * as _ from '../utils';
 
 /**
  * Constructs new BlockAPI object
  *
  * @class
  *
  * @param {Block} block - Block to expose
+ * @param {Function} isReadOnlyEnabled - method to check if read-only mode is enabled
```
neSpecc, line 13: describe why it is function

src/components/modules/api/index.ts
```diff
@@ -8,11 +8,34 @@
...
+      inlineToolbar: this.Editor.InlineToolbarAPI,
+      tooltip: this.Editor.TooltipAPI,
+      i18n: this.Editor.I18nAPI,
+      readOnly: this.Editor.ReadOnlyAPI,
+    };
+  }
+
   /**
    * Editor.js Core API modules
    */
```
neSpecc, line 41: iterate `components` and automatically expose methods or classes

src/components/modules/api/index.ts
```diff
@@ -50,4 +74,23 @@ export default class API extends Module {
       }
     ) as APIInterfaces;
   }
+
+  /**
+   * Returns decorated list of methods that have read-only restriction
+   *
+   * @param methodGroup - sss
```
neSpecc, line 81: write normal description

src/components/modules/api/listeners.ts
neSpecc: still not fixed

src/components/modules/crossBlockSelection.ts
neSpecc: @gohabereg answer here pls

neSpecc (commented):
- remove code disabling API in read-only
- add warning in saver

talyguryn (approved):

src/components/modules/ui.ts
```diff
@@ -708,8 +734,16 @@ export default class UI extends Module {
    * @param {Event} event - selection event
    */
   private selectionChanged(event: Event): void {
+    const { CrossBlockSelection, BlockSelection } = this.Editor;
     const focusedElement = Selection.anchorElement as Element;
 
+    if (CrossBlockSelection.isCrossBlockSelectionStarted) {
+      // Removes all ranges when any Block is selected
+      if (BlockSelection.anyBlockSelected) {
```
neSpecc, line 742: Hard computations on every selection moving. Should be optimized

neSpecc (changes requested):

src/components/modules/blockSelection.ts
```diff
@@ -17,6 +17,14 @@ import { SanitizerConfig } from '../../../types/configs';
  *
  */
 export default class BlockSelection extends Module {
+
+  /**
+   * Saves any block selected cache
```
neSpecc, line 22:
```suggestion
   * Sometimes .anyBlockSelected can be called frequently, 
   * for example at ui@selectionChange (to clear native browser selection in CBS)
   * We use cache to prevent multiple iterations through all the blocks
```

neSpecc (approved):

PR #1048, opened 2020-03-01
Fix: bind editor when calling onChange

### LINKED ISSUE #999: onChange should have the editor instance available
The current behaviour of the onChange event mandates that the callback needs to be in a closure on the config object because the editor instance isn't available from the onChange event any other way.

e.g
```
let editor = new EditorJS({
  ...
  onChange: changeHandler
})
changeHandler = () => {  editor.save()  } 
```

I don't always have access to the editor instance in some abstractions of my tooling and it would be great if I could do something like the following

```
let editor = new EditorJS({
  ...
  onChange: somemodule.changeHanlder
});

/* Another File */
changeHandler = () => {  this.save()  } 
/// OR
changeHanlder = (ed) => { ed.save(); } 
```

Changing this means that I don't have to have the actual instance of my editor available in any scope and would simplify the models I have.

https://github.com/codex-team/editor.js/blob/ac93017c706f091fd3014c3edb248db4e03f55dd/src/components/modules/modificationsObserver.ts#L36 

You could .bind(editorInstance), or pass it as a parameter.

melbahja:
fixes #999


neSpecc:
@melbahja thanks for the PR. I've made some little improvements: 
- it would better to pass only external API rather than all internal editor modules.
- I think, passing and API as a parameter would better solution because some users can specify the onChange callback with arrow function inside own class, so we don't want to rewrite the context.

PR #1052, opened 2020-03-09
Add option for block plugins to specify if they are empty

xel-x:
I was writing some block-tools, which have a slightly more fancy UI (well, they don't do much more than counting characters and displaying the count for now), and the + sign you get on empty paragraphs happend to be unreachable.

I found out, that it depends on the tool being "empty" in DOM -- which mine wasn't and coudn't ever be.

I added the option to define a getter in the tool, to check if it is empty, falling back to the old way if that getter isn't present.


neSpecc:
Seems outdated.

PR #1061, opened 2020-03-14
Improvements: new Block Settings button style

neSpecc:
Also, focused block highlighting was removed. Now it will be highlighted only when settings is opened.

image


neSpecc:
See #1064

PR #1075, opened 2020-03-21
[Feature] BlockAPI Interface

gohabereg:


natterstefan:
Cool! I can’t wait to test the new version.

src/components/block/api.ts
```diff
@@ -0,0 +1,152 @@
...
+    get holder(): HTMLElement {
+      return block.holder;
+    },
+
+    /**
+     * Tool contents holder, returned from Tool's render() method
+     *
+     * @return {HTMLElement}
+     */
+    get pluginsContent(): HTMLElement {
```
gohabereg: @neSpecc @khaydarov I think we need to rename this as `plugin` is never used for any Editor.js entity

src/components/block/api.ts
```diff
@@ -0,0 +1,152 @@
...
+    get name(): string {
+      return block.name;
+    },
+
+    /**
+     * Tool config passed on Editor's initialization
+     *
+     * @return {ToolConfig}
+     */
+    get settings(): ToolConfig {
```
neSpecc, line 22: I think it is not required by outside

src/components/block/api.ts
```diff
@@ -0,0 +1,152 @@
...
+    get data(): Promise<BlockToolData> {
+      return block.data;
+    },
+
+    /**
+     * True if Tool has merge method
+     *
+     * @return {boolean}
+     */
+    get mergeable(): boolean {
```
neSpecc, line 58: I think it is not required by outside

src/components/block/api.ts
```diff
@@ -0,0 +1,152 @@
...
+    get isEmpty(): boolean {
+      return block.isEmpty;
+    },
+
+    /**
+     * True if Block has media
+     *
+     * @return {boolean}
+     */
+    get hasMedia(): boolean {
```
neSpecc, line 76: I think it is not required by outside

src/components/block/api.ts
```diff
@@ -0,0 +1,152 @@
...
+    get selected(): boolean {
+      return block.selected;
+    },
+
+    /**
+     * True if caret at Block
+     *
+     * @return {boolean}
+     */
+    get focused(): boolean {
```
neSpecc, line 94: I think it is not required by outside

src/components/block/api.ts
```diff
@@ -0,0 +1,152 @@
...
+    get pluginsContent(): HTMLElement {
+      return block.pluginsContent;
+    },
+
+    /**
+     * Saved Block data
+     *
+     * @return {BlockToolData}
+     */
+    get data(): Promise<BlockToolData> {
```
neSpecc, line 49: I think it is not required by outside

src/components/block/index.ts
```diff
@@ -176,35 +176,11 @@ export default class Block {
...
-        if (!$.isExtensionNode(contentNode)) {
-          return contentNode as HTMLElement;
-        }
-      }
-    }
-
-    return null;
-  }
-
   /**
```
neSpecc, line 179: Plugin can change its template element, like Header

src/components/modules/paste.ts
```diff
@@ -709,11 +710,11 @@ export default class Paste extends Module {
   /**
    * Insert data passed as application/x-editor-js JSON
    *
-   * @param {object} blocks — Blocks' data to insert
+   * @param {Array<Omit<SavedData, 'time'>>} blocks — Blocks' data to insert
```
neSpecc, line 713: there can be more omitted properties in a future, for example, `layout`

src/components/block/api.ts
gohabereg: WIll rename it to `config` as some BlockTunes might rely on Tool's configuration

src/components/block/api.ts
```diff
@@ -0,0 +1,110 @@
...
+    get name(): string {
+      return block.name;
+    },
+
+    /**
+     * Tool config passed on Editor's initialization
+     *
+     * @returns {ToolConfig}
+     */
+    get config(): ToolConfig {
```
neSpecc, line 25: There can be a misunderstanding with existed [Tool config](https://editorjs.io/provide-custom-configuration) term.

gohabereg: That's exactly the ToolConfig object

docs/api.md
```diff
@@ -1,9 +1,33 @@
...
 
+## Block API
+
+API for certain Block methods and properties. You can access it through `editor.api.block.getBlockByIndex` method or get it form `block` property of [Tool constructor](../types/tools/block-tool.d.ts) argument.
+
+`name: string` — Block's Tool name
+
+`config: ToolConfig` — Tool config passed on Editor initialization
+
+`holder: HTMLElement` — HTML Element that wraps Tool's HTML content
```
neSpecc, line 15: is this an element returned by tools render method? The description is not clear.

docs/api.md
```diff
@@ -1,9 +1,33 @@
...
 
-Blocks have access to the public methods provided by Editor.js API Module. Plugin and Tune Developers
+Tools have access to the public methods provided by Editor.js API Module. Plugin and Tune Developers
 can use Editor\`s API as they want.
 
+## Block API
+
+API for certain Block methods and properties. You can access it through `editor.api.block.getBlockByIndex` method or get it form `block` property of [Tool constructor](../types/tools/block-tool.d.ts) argument.
+
+`name: string` — Block's Tool name
```
neSpecc, line 11: what is a tool name?

docs/api.md
```diff
@@ -1,9 +1,33 @@
...
+
+`isEmpty: boolean` — `true` if Block has any editable content
+
+`selected: boolean` - `true` if Block is selected with Cross-Block Selection
+
+`set stretched(state: boolean)` — set Block's stretch state
+
+`stretched: boolean` — `true` if Block is stretched
+
+`call(methodName: string, param?: object): void` — method to call Tool's methods with checks and error handlers under-the-hood
```
neSpecc, line 25: what methods?

docs/api.md
```diff
@@ -1,9 +1,33 @@
...
+
+`save(): Promise<void|SavedData>` — returns data saved from current Block's state, including Tool name and saving exec time
```
neSpecc, line 27: - what is the difference between `SavedData` and `BlockToolData`? 
- ` saving exec time` - we include it only for common saving, not for every block.

docs/api.md
```diff
@@ -36,13 +60,13 @@ use 'move' instead)
 
 `getCurrentBlockIndex()` - current Block index
 
-`getBlockByIndex(index: Number)` - returns Block with passed index
+`getBlockByIndex(index: Number)` - returns Block API object by passed index
 
 `getBlocksCount()` - returns Blocks count
 
-`stretchBlock(index: number, status: boolean)` - make Block stretched
+`stretchBlock(index: number, status: boolean)` - *Deprecated* make Block stretched
```
neSpecc, line 67: what to use instead?

src/components/block/api.ts
```diff
@@ -0,0 +1,114 @@
...
+import { BlockAPI as BlockAPIInterface } from '../../../types/api';
+
+/**
+ * Constructs new BlockAPI object
+ *
+ * @class
+ *
+ * @param {Block} block - Block to expose
+ */
+function BlockAPI(block: Block): void {
```
neSpecc, line 13: why not a class?

src/components/block/index.ts
```diff
@@ -23,10 +25,19 @@ import * as _ from './utils';
...
-import DeleteTune from './block-tunes/block-tune-delete';
-import MoveDownTune from './block-tunes/block-tune-move-down';
-import SelectionUtils from './selection';
+import MoveUpTune from '../block-tunes/block-tune-move-up';
+import DeleteTune from '../block-tunes/block-tune-delete';
+import MoveDownTune from '../block-tunes/block-tune-move-down';
+import SelectionUtils from '../selection';
+import BlockAPI from './api';
+
+interface BlockConstructorOptions {
```
neSpecc, line 34: add a documentation

src/components/block/api.ts
gohabereg: Because `block` var would be accessible

docs/api.md
gohabereg: We used this definition earlier for the same thing, is it better?
image

docs/api.md
gohabereg: `BlockToolData` is an object, returned from `save` method of Tool instance. `Block` instance `save` wraps `BlockToolData` with `SavedData` object.

`BlockToolData`
```
{
  "text": "Some text"
}
``` 

`SavedData`
```
{
  "tool": "paragraph",
  "data": {
    "text": "Some text"
  },
  time: 0.1923
}
```

General `editor.save` method uses time from `SavedData` from each Block to calculate final exec time.

I agree that we have some inconsistency here, but I think it's out of the scope of this feature.

src/components/block/api.ts
neSpecc: You can note that via comment.

src/components/modules/api/blocks.ts
```diff
@@ -155,6 +156,8 @@ export default class BlocksAPI extends Module {
    *
    * @param {number} index - index of Block to stretch
    * @param {boolean} status - true to enable, false to disable
+   *
+   * @deprecated Use BlockAPI interface to stretch Blocks
```
neSpecc, line 160: also, we can add a console warning

src/components/modules/blockManager.ts
```diff
@@ -225,30 +230,60 @@ export default class BlockManager extends Module {
...
-  public insert(
-    toolName: string = this.config.initialBlock,
-    data: BlockToolData = {},
-    settings: ToolConfig = {},
-    index: number = this.currentBlockIndex + 1,
-    needToFocus = true
-  ): Block {
-    const block = this.composeBlock(toolName, data, settings);
+  public insert({
+    tool = this.config.initialBlock,
```
neSpecc, line 239: fix jsdoc with new naming

src/components/modules/blockManager.ts
```diff
@@ -225,30 +230,60 @@ export default class BlockManager extends Module {
...
+      this.currentBlockIndex++;
     }
 
     return block;
   }
 
+  /**
+   * Replace current working block
+   *
+   * @param {string} toolName — plugin name
```
neSpecc, line 270: tool

src/components/modules/paste.ts
```diff
@@ -753,11 +751,11 @@ export default class Paste extends Module {
         needToReplaceCurrentBlock = isCurrentBlockInitial && BlockManager.currentBlock.isEmpty;
       }
 
-      if (needToReplaceCurrentBlock) {
-        BlockManager.replace(tool, data, settings);
-      } else {
-        BlockManager.insert(tool, data, settings);
-      }
+      BlockManager.insert({
```
khaydarov, line 754: insert and replace are different actions, why they are joined?

hughfenghen:
Is it better to keep the naming of [SavedData.tool](https://github.com/codex-team/editor.js/blob/fc3e1467645a88dc491b634a654471e1f6f59638/src/types-internal/block-data.d.ts#L7) and [OutputBlockData.type](https://github.com/codex-team/editor.js/blob/f172ca3ebfeeab3216dfcba5bed726bf0696a7ad/types/data-formats/output-data.d.ts#L10) consistent?

src/components/modules/paste.ts
gohabereg: In several cases when we want to insert block, we actually replace it.
For example, if you paste into empty paragraph it should be replaced. So instead of adding if statement you can now pass the `replace` parameter

```
if (shouldReplace) {
  BlockManager.replace(...);
} else {
  BlockManager.insert(...);
}

// vs

BlockManager.insert({..., replace: shouldReplace});
```

gohabereg:
> Is it better to keep the naming of [SavedData.tool](https://github.com/codex-team/editor.js/blob/fc3e1467645a88dc491b634a654471e1f6f59638/src/types-internal/block-data.d.ts#L7) and [OutputBlockData.type](https://github.com/codex-team/editor.js/blob/f172ca3ebfeeab3216dfcba5bed726bf0696a7ad/types/data-formats/output-data.d.ts#L10) consistent?

Sure, but it's out of scope of this PR. We will work on it in future releases

Issue #1093, opened 2020-04-07
'codex.tooltips' submodule is missing


tasuku-s:
When I cloned this repository and execute command `git submodule update --init --recursive`,
the error has been occurred that is `fatal: No url found for submodule path 'src/components/external/codex.tooltips' in .gitmodules`.

Probably, the module is removed from `.gitmodules`, but `src/components/external/codex.tooltips' directory is still remained.

PR #1094, opened 2020-04-07
fix `fatal: No url found for submodule path 'src/components/external/codex.tooltips' in .gitmodules`  error

tasuku-s:
#1093


gohabereg:
Fixed by #1105

PR #1106, opened 2020-04-18
[Feature] i18n

### LINKED ISSUE #751: Support for i18n
I'm trying to translate the editor tooltips and names to Portuguese, but the docs don't specify a way to do that, and in fact, it doesn't seem possible at the moment.

Can this be done?

neSpecc:
- ability to change all internal UI texts
- ability to change titles of the Inline and Block tools
- ability to pass translations to the tools that can be accessed by tools developers by API

Resolves #751


src/components/core.ts
```diff
@@ -3,6 +3,7 @@ import * as _ from './utils';
 import { EditorConfig, OutputData, SanitizerConfig } from '../../types';
 import { EditorModules } from '../types-internal/editor-modules';
 import { LogLevels } from './utils';
+import I18n from "./i18n";
```
gohabereg, line 6:
```suggestion
import I18n from './i18n';
```

src/components/i18n/index.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    return section[dictKey];
+  }
+
+  /**
+   * Adjust module for using external dictionary
+   *
+   * @param dictionary - new messages list to override default
+   */
+  public static setDictionary(dictionary: {[key: string]: any}): void {
```
gohabereg, line 39: Type should be moved to `type` variable or interface.
Here it's `{[key: string]: any}`, however I18n.currentDictionary is `{[key: string]: object}`

Also, it's better to specify the type is recurrent:
```
type DictValue = {[string: key]: Dictionary | string} | string;
type Dictionary = {[string: key]: DictValue};
```

gohabereg: However, if the root level of the dictionary has  known properties, it's better to specify them:

```
interface I18nDict {
  tools: Dictionary;
  blockTunes: Dictionary;
  ...
}
```

src/components/i18n/index.ts
```diff
@@ -0,0 +1,59 @@
...
+    I18n.currentDictionary = dictionary;
+  }
+
+  /**
+   * Find messages section by namespace path
+   *
+   * @param namespace - path to section
+   */
+  private static getSection(namespace: string): object|undefined {
```
gohabereg, line 48: Maybe better to name it `getNamespace` as the arg name is 'namespace'?

src/components/modules/api/i18n.ts
```diff
@@ -0,0 +1,54 @@
...
+        return `blockTunes.${toolName}`;
+    }
+  }
+
+  /**
+   * Return I18n API methods with global dictionary access
+   */
+  public get methods(): I18n {
+    return {
+      t: (namespace: string, dictKey: string): string => I18nInternal.t(namespace, dictKey),
```
gohabereg, line 32: Don't think a tool will need access to the whole i18n dict. Maybe just left `tn` here and rename it to `t`?

gohabereg: I mean, pass only `t` method with the namespace just for tool

src/components/modules/toolbar/conversion.ts
```diff
@@ -68,7 +68,7 @@ export default class ConversionToolbar extends Module {
     this.nodes.tools = $.make('div', ConversionToolbar.CSS.conversionToolbarTools);
 
     const label = $.make('div', ConversionToolbar.CSS.conversionToolbarLabel, {
-      textContent: 'Convert to',
+      textContent: I18n.t('ui.inlineToolbar.converter', 'Convert to'),
```
gohabereg, line 71: I think we should move all internal namespaces and messages to typed objects or enums

src/components/modules/toolbar/inline.ts
```diff
@@ -588,7 +591,10 @@ export default class InlineToolbar extends Module {
      * Enable tooltip module on button
      */
     const tooltipContent = $.make('div');
-    const toolTitle = Tools.toolsClasses[toolName][Tools.INTERNAL_SETTINGS.TITLE] || _.capitalize(toolName);
+    const toolTitle = I18n.t(
+      `toolNames`,
```
gohabereg, line 595: Unnecessary literal

src/components/modules/tools.ts
```diff
@@ -472,3 +473,22 @@ export default class Tools extends Module {
     };
   }
 }
+
+/**
+ * What kind of plugins developers can create
+ */
+export enum toolTypes {
```
gohabereg, line 480: Enums should be in PascalCase
```suggestion
export enum ToolTypes {
```

gohabereg: Also, usually, enums go in singular form
```suggestion
export enum ToolType {
```

src/components/modules/tools.ts
```diff
@@ -472,3 +473,22 @@ export default class Tools extends Module {
...
+  /**
+   * Block tool
+   */
+  BLOCK,
```
gohabereg, line 484: Enum values also usually go in PascalCase
```suggestion
  Block,
```

src/components/i18n/index.ts
```diff
@@ -0,0 +1,60 @@
...
+   * @param dictKey - dictionary key. Better to use default locale original text
+   */
+  public static t(namespace: string, dictKey: string): string {
+    const section = I18n.getNamespace(namespace);
+
+    if (_.isEmpty(section)) {
+      _.log('I18n: section %o was not found in current dictionary', 'warn', namespace);
+    }
+
+    if (!section || !section[dictKey]) {
```
gohabereg, line 28: Need to check here if `section[dictKey]` is an object (or not a string)

src/components/i18n/index.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * Find messages section by namespace path
+   *
+   * @param namespace - path to section
+   */
+  private static getNamespace(namespace: string): Dictionary {
+    const parts = namespace.split('.');
+
+    return parts.reduce((key, part) => {
```
gohabereg, line 52: Don't think `key` is a good name here

types/configs/i18n-dictionary.d.ts
```diff
@@ -0,0 +1,93 @@
+/**
+ * Structure of the i18n dictionary
+ */
+export interface I18nDictionary {
+  /**
+   * Translation of internal UI components of the editor.js core
+   */
+  ui: Dictionary;
```
gohabereg, line 8: I think all fields should be optional (if we should be able to translate some parts and don't translate other) or required (if a dictionary is required to be applied to all parts)

types/index.d.ts
```diff
@@ -18,6 +18,7 @@ import {
   Styles,
   Toolbar,
   Tooltip,
+  I18n,
```
gohabereg, line 21: Other types also should be exported from index.d.ts to be accessible by import from `@editorjs/editorjs` package

src/components/i18n/index.ts
neSpecc: Section is always an object

types/index.d.ts
neSpecc: done

src/components/i18n/index.ts
gohabereg: As calls are not typesafe, you can pass path to value, not section

neSpecc: so the section will be undefined, and this `if` cover this case

src/components/i18n/namespace-internal.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  return result as DictNamespaces<typeof defaultDictionary>;
+}
+
+/**
+ * Type safe access to the internal messages dictionary sections
+ *
+ * @example I18n.ui(I18nInternalNS.ui.blockTunes.toggler, 'Click to tune');
+ */
+export const I18nInternalNS = getNamespaces(defaultDictionary);
```
neSpecc, line 52: image

src/components/i18n/namespace-internal.ts
```diff
@@ -0,0 +1,52 @@
...
+import { DictNamespaces } from '../../../types/configs/i18n-internal-namespace';
+import { typeOf } from '../utils';
+
+/**
+ * Evaluate messages dictionary and return object for namespace chaining
+ *
+ * @param dict - Messages dictionary
+ * @param [keyPath] - subsection path (used in recursive call)
+ */
+function getNamespaces(dict, keyPath?): DictNamespaces<typeof defaultDictionary> {
```
gohabereg, line 11: Args types are not defined

src/components/modules/api/i18n.ts
```diff
@@ -0,0 +1,55 @@
...
+   * Return namespace section for tool or block tune
+   *
+   * @param toolName - name of tool. Used to provide dictionary only for this tool
+   * @param toolType - 'block' for Block Tool, 'inline' for Inline Tool, 'tune' for Block Tunes
+   */
+  private static getNamespace(toolName: string, toolType: ToolType): string {
+    switch (toolType) {
+      case ToolType.Block:
+      case ToolType.Inline:
+        return `tools.${toolName}`;
```
gohabereg, line 21: Maybe put `tools` and `blockTunes` to constant?

gohabereg: Somewhere next by ToolType enum

types/api/i18n.d.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Describes Editor`s I18n API
+ */
+export interface I18n {
+  /**
+   * Perform translation with automatically added namespace like `tools.${toolName}` or `blockTunes.${tuneName}`
+   *
+   * @param dictKey - what to translate
```
gohabereg, line 8: ESLint should throw a warning that param type is not defined

types/configs/i18n-internal-namespace.d.ts
```diff
@@ -0,0 +1,35 @@
+/**
```
gohabereg, line 1: I think it's better to put this file into `src/types-internal` folder

Issue #1090, opened 2020-04-01
Feature Request (Tool-Setting): add option to not delete selected block on backspace or enter keydown


natterstefan:
## Scenario

A block's tool setting opens a modal (or any other new view outside of editorjs). Whenever the user hits backspace _or_ enter in the new view though, the block - which is still selected (`.ce-block--selected`) - gets deleted.

There is currently _no way_ to remove the selection from the block, when a tune was toggled/clicked.

## Example

![Screen Capture on 2020-04-01 at 16-40-17](https://user-images.githubusercontent.com/1043668/78150397-887ae480-7437-11ea-810f-d27563ea9b55.gif)

## Example Codepen

https://codepen.io/natterstefan/pen/PoqLVJr?editors=0110

## Research

There is a `enableLineBreaks` setting, managing how blocks behave when `enter` is pressed. In [the example](https://codepen.io/natterstefan/pen/PoqLVJr?editors=0110) it still looks like this is not working properly.

https://github.com/codex-team/editor.js/blob/be6f9b78f265c3d553c799c99e59c9916bf29e17/src/components/modules/blockEvents.ts#L257-L263

Same goes for `backspace`, but here we never reach this part of the code:

https://github.com/codex-team/editor.js/blob/be6f9b78f265c3d553c799c99e59c9916bf29e17/src/components/modules/blockEvents.ts#L353-L361

Because the block is still selected and [results to true in this if](https://github.com/codex-team/editor.js/blob/be6f9b78f265c3d553c799c99e59c9916bf29e17/src/components/modules/blockEvents.ts#L327-L338
).

## Feature request

Enable developers to unselect a block when a tune was toggled. Means, a new method is added to the blocks API (https://editorjs.io/api#blocks).

## Current Workaround (not 100% tested tough)

When the toggle is clicked, we remove the [selected class](https://github.com/codex-team/editor.js/blob/be6f9b78f265c3d553c799c99e59c9916bf29e17/src/components/block.ts#L67) from the block. This should help us to pass the if`s in the keydown handlers.

But this is kinda hacky IMHO.

## Additional questions

[...]

PR #1109, opened 2020-04-18
Fix block deleted when settings are opened #1090

m-alzam:
Prevent block removal when settings are opened #1090


neSpecc:
I think it would better to leave an ability to delete blocks this way. So it's better to close Block Settings in such a case.

m-alzam:
> I think it would better to leave an ability to delete blocks this way. So it's better to close Block Settings in such a case.

But there's a problem with input elements inside settings. If you press the backspace key while editing text in a settings' input, the block is deleted instead of doing the default behavior (remove a character).

neSpecc:
> > I think it would better to leave an ability to delete blocks this way. So it's better to close Block Settings in such a case.
> 
> But there's a problem with input elements inside settings. If you press the backspace key while editing text in a settings' input, the block is deleted instead of doing the default behavior (remove a character).

Ok, got this. We need to think about how to detect if the Caret inside some kind of an editable element at the settings. Or maybe allow an ability to prevent default behavior from the plugin side.

gohabereg:
I think we can use `document.activeElement` for that

m-alzam:
Thanks @neSpecc and @gohabereg for the suggestions. I'l try to implement the check if we're in an editable element tomorrow.

m-alzam:
PR updated.

src/components/modules/ui.ts
```diff
@@ -390,13 +390,23 @@ export default class UI extends Module {
   private backspacePressed(event: KeyboardEvent): void {
     const {BlockManager, BlockSelection, Caret} = this.Editor;
 
+    /** Prevent block removal while editing an input inside settings */
+    if (this.Editor.BlockSettings.opened) {
+      if (document.activeElement.tagName === 'INPUT') {
```
neSpecc, line 395: we have a useful method for that, see `dom.canSetCaret()`

m-alzam: Ok, thanks. It has more features. I'll update the PR this evening of I have time.

neSpecc:
The solution works, but it only allows to press backspace at the input. There are also some problems with Flipper - class responsible for tab/arrow navigation between buttons:
- you can't navigate by arrows inside the `input` at the block tunes
- you can't focus the `input` with Flipper
- the input has no auto focus when block tunes are opened

m-alzam:
Closing in favor of #1123

Issue #1073, opened 2020-03-20
Embed tool position in toolbar


mikekucherov:
Embed tool is always goes to the end of tools list

PR #1113, opened 2020-04-18
Save Tools' order in the Toolbox

gohabereg:


gohabereg:
Resolves #1073

src/components/modules/tools.ts
```diff
@@ -313,6 +313,7 @@ export default class Tools extends Module {
    * @param {object} data - append tool to available list
    */
   public success(data: {toolName: string}): void {
+    console.log(data);
```
neSpecc, line 316: log

src/components/modules/tools.ts
```diff
@@ -429,20 +430,30 @@ export default class Tools extends Module {
     for (const toolName in this.toolsClasses) {
       if (Object.prototype.hasOwnProperty.call(this.toolsClasses, toolName)) {
         const toolClass = this.toolsClasses[toolName];
+        const toolConfig = this.toolsSettings[toolName][this.USER_SETTINGS.CONFIG];
 
         if (typeof toolClass.prepare === 'function') {
```
neSpecc, line 435: so this condition is no longer needed because both cases are similar.

PR #1120, opened 2020-04-23
Fix position for the unlink button

ch3rn1k:
Found, that after adding link to the text and selecting text again is placing unlink button to the bottom of the element. Removing `margin-bottom: -1px;` fixes that.

изображение

### DIFF
--- a/src/styles/inline-toolbar.css
+++ b/src/styles/inline-toolbar.css
@@ -114,7 +114,6 @@
     }
     .icon--unlink {
       display: inline-block;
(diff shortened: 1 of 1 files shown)


neSpecc:
Please describe you browser version and the OS.

neSpecc:
Found the bug in Firefox, fixed by https://github.com/codex-team/editor.js/commit/f4c40513b9236634856bf800c3f5fcc9e86d2da5

neSpecc:
Will be released with 2.18

PR #1123, opened 2020-04-25
Fix behaviour of inputs editing in block settings

### LINKED ISSUE #1090: Feature Request (Tool-Setting): add option to not delete selected block on backspace or enter keydown

gohabereg:
Resolves #1090


gohabereg:
Hi @natterstefan @m-alzam,

This PR should resolve problems with `enter` and `backspace` keys. Could you please check it in your use-cases?

src/components/domIterator.ts
```diff
@@ -156,7 +156,8 @@ export default class DomIterator {
       focusedButtonIndex = (this.items.length + focusedButtonIndex - 1) % this.items.length;
     }
 
-    if (Dom.isNativeInput(this.items[focusedButtonIndex])) {
+    if (Dom.canSetCaret(this.items[focusedButtonIndex])) {
+
```
neSpecc, line 160: there can be contenteditable element, so you need to use custom `focus()` method

src/components/flipper.ts
```diff
@@ -190,7 +213,7 @@ export default class Flipper {
       _.keyCodes.ENTER,
     ];
 
-    if (this.allowArrows) {
+    if (this.allowArrows && this.iterator.currentItem !== document.activeElement) {
```
neSpecc, line 216: need to describe this case or move the second condition to the variable

src/components/flipper.ts
```diff
@@ -199,11 +222,7 @@ export default class Flipper {
       );
     }
 
-    if (!this.activated || handlingKeyCodeList.indexOf(event.keyCode) === -1) {
-      return false;
-    }
-
-    return true;
+    return (this.activated && handlingKeyCodeList.indexOf(event.keyCode) !== -1);
```
neSpecc, line 225: extra parentheses

src/components/modules/blockSelection.ts
```diff
@@ -188,17 +188,20 @@ export default class BlockSelection extends Module {
...
 
+    const isKeyboard = reason && (reason instanceof KeyboardEvent);
+    const isPrintableKey = isKeyboard && _.isPrintableKey((reason as KeyboardEvent).keyCode);
+
     /**
      * If reason caused clear of the selection was printable key and any block is selected,
      * remove selected blocks and insert pressed key
      */
-    if (this.anyBlockSelected && reason && reason instanceof KeyboardEvent && _.isPrintableKey(reason.keyCode)) {
+    if (this.anyBlockSelected && isKeyboard && isPrintableKey && !SelectionUtils.exists) {
```
neSpecc, line 198: `SelectionUtils.exists` - bad naming. Exists what?

src/components/modules/toolbar/blockSettings.ts
```diff
@@ -158,7 +176,10 @@ export default class BlockSettings extends Module {
       return this.buttons;
     }
 
-    const toolSettings = this.nodes.toolSettings.querySelectorAll(`.${this.Editor.StylesAPI.classes.settingsButton}`);
+    const toolSettings = this.nodes.toolSettings.querySelectorAll(
+      // Select buttons and inputs
+      `.${StylesAPI.classes.settingsButton}, .${StylesAPI.classes.input}`
```
neSpecc, line 181: why not use `$.findAllInputs()` here?

src/components/modules/toolbar/blockSettings.ts
```diff
@@ -195,7 +216,18 @@ export default class BlockSettings extends Module {
   private enableFlipper(): void {
     this.flipper = new Flipper({
       focusedItemClass: this.CSS.focusedButton,
-      activateCallback: () => {
+      activateCallback: (activatedItem) => {
```
neSpecc, line 219: the parameter is not described. maybe name it `focusedItem`?

src/components/modules/ui.ts
```diff
@@ -410,7 +416,7 @@ export default class UI extends Module {
   private backspacePressed(event: KeyboardEvent): void {
     const { BlockManager, BlockSelection, Caret } = this.Editor;
 
-    if (BlockSelection.anyBlockSelected) {
+    if (BlockSelection.anyBlockSelected && !Selection.exists) {
```
neSpecc, line 419: need to describe why the second part was added

src/components/modules/ui.ts
```diff
@@ -438,7 +470,7 @@ export default class UI extends Module {
     const { BlockManager, BlockSelection, Caret } = this.Editor;
     const hasPointerToBlock = BlockManager.currentBlockIndex >= 0;
 
-    if (BlockSelection.anyBlockSelected) {
+    if (BlockSelection.anyBlockSelected && !Selection.exists) {
```
neSpecc, line 473: need to describe why the second part was added

src/components/modules/toolbar/blockSettings.ts
```diff
@@ -195,7 +216,18 @@ export default class BlockSettings extends Module {
...
+
+        /**
+         * If activated item is editable element, restore selection to block and flip to the next item
```
neSpecc, line 222: can't understand this moment. Why do we need to restore selection to block, if, for example, an input at block settings is focused

gohabereg: I thought about this scenario:
1. Open Block settings via tab
2. Type something to an editable element there
3. Press enter

What should happened after that?
For sure, Tune should do something with the typed text, but what about settings block. Should it be closed? At the moment caret is returned to the Block and next Tune is selected

src/components/dom.ts
```diff
@@ -573,4 +599,31 @@ export default class Dom {
   public static isAnchor(element: Element): element is HTMLAnchorElement {
     return element.tagName.toLowerCase() === 'a';
   }
+
+  /**
+   * Set focus to contenteditable or native input element
+   *
+   * @param element - element where to set focus
+   */
+  public static focus(element: HTMLElement): void {
```
neSpecc, line 608: can we use this method in module/caret@set ?

neSpecc: and maybe it's better to place it to the selectionUtis?

m-alzam:
Hi @gohabereg, I'll try test your patch tomorrow with my plugin. Thanks for it. Should I abandon my PR?

m-alzam:
Hi @gohabereg: the patch works fine for me. Thanks!

src/components/modules/toolbar/blockSettings.ts
```diff
@@ -130,6 +141,15 @@ export default class BlockSettings extends Module {
   public close(): void {
     this.nodes.wrapper.classList.remove(this.CSS.wrapperOpened);
 
+    /**
+     * Restore selection to Block
```
neSpecc, line 145: need to describe current condition: why we need to check if the selection is at editor

src/components/modules/toolbar/blockSettings.ts
neSpecc: can we write a method for accepting `cdx-buttons` and every-inputs with saving their order at dom?

src/components/selection.ts
```diff
@@ -266,6 +275,37 @@ export default class SelectionUtils {
     return window.getSelection();
   }
 
+  /**
+   * Set focus to contenteditable or native input element
+   *
+   * @param element - element where to set focus
+   * @param offset - offset of cursor
+   */
```
neSpecc, line 283: describe what it returns

PR #1131, opened 2020-04-28
preserve block focus after keyboard events

### LINKED ISSUE #1130: focus not preserved after keyboard or block select events
The block focus is not preserved after keyboard events or if a different block is selected.   (This will cause issues with apps that use CSS to highlight the currently focused block).

This is similar to [issue#1124](https://github.com/codex-team/editor.js/issues/1124), but occurs under several other circumstances.

Steps to reproduce:
1. display any doc in EditorJS
2. click or tab to any block, or enter text via keyboard

Expected behavior:
The currently selected block should have a style of "ce-block--focused" but it doesn't.

Device, Browser, OS:
Firefox 75.0, Ubuntu 20.04 (doesn't matter)

Editor.js version:
2.17.0/2.18.0

The fix is to set the focus again after mouse or keyboard events are processed or a new block is selected.

djulien:
Set focus again after keyboard event or block selection changed.
fixes #1130 

(I'm not too good with all the git branching stuff - hopefully I did it correctly)

### DIFF
--- a/src/components/modules/blockEvents.ts
+++ b/src/components/modules/blockEvents.ts
@@ -51,6 +51,7 @@ export default class BlockEvents extends Module {
         this.escapePressed(event);
         break;
     }
+    this.Editor.BlockManager.highlightCurrentNode(); //set focus on current block after key processed -DJ
   }
 
   /**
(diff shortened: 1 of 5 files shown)


src/components/modules/blockEvents.ts
```diff
@@ -455,8 +456,9 @@ export default class BlockEvents extends Module {
       return;
     }
 
-    if (this.Editor.Caret.navigateNext()) {
-      /**
+//    if (this.Editor.Caret.navigateNext()) {
```
chapterjason, line 459: Just remove it, git has it in history.

neSpecc:
Seems outdated.

PR #1140, opened 2020-05-01
Hotfix/issue1133 selection shortcut removed on editor destroy

### LINKED ISSUE #1133: [Bug] CMD+A shortcut not removed on destroying the editor
During editor initialisation, a module called blockSelection is initialised which registers a shortcut handler for CMD+A (src/components/modules/blockSelection.ts:132). The shortcut is added but it doesn't seem to be removed on destroying the editor. If I destroy the editor and initialise it, I get two handlers registered for CMD+A. One for the old DOM node which isn't really there in the DOM now and one for the current DOM node. So when I press CMD+A, the old one crashes in handleCommandA function's getBlock call in blockSelection.ts.

Steps to reproduce:
1. Initialize the editor with some data.
2. Destroy the editor with setTimeout after initialization.
3. After destroy initialize the editor with the same data again.
4. Click in a text block and press CMD+A.
5. The handleCommandA function will be called twice. One for the old one and one for the new.

Expected behavior:
De-register shortcuts on destroying the editor instance.

Screenshots:
None

Editor.js version: 2.17.0

sis-dk:
Solves the problem of CMD+A shortcut not being removed when editor is destroyed.
So basically, now every module can have a destroy method and if one is provided in a module, it will automatically be called when editor.destroy is called. There is a loop that goes through every module and checks if it has a destroy method. If one is present, it'll be called.
Closes #1133


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@editorjs/editorjs",
-  "version": "2.18.0",
+  "version": "2.18.1",
```
gohabereg, line 3: 2.18 is not released yet, so you can include patch to 2.18 instead of bumping patch version

src/codex.ts
```diff
@@ -81,9 +82,14 @@ export default class EditorJS {
   public exportAPI(editor: Core): void {
     const fieldsToExport = [ 'configuration' ];
     const destroy = (): void => {
-      editor.moduleInstances.Listeners.removeAll();
-      editor.moduleInstances.UI.destroy();
-      editor.moduleInstances.ModificationsObserver.destroy();
+      for(const moduleName in editor.moduleInstances) {
```
gohabereg, line 85: Why not?
```
Object.values(editor).forEach(...)
```

src/codex.ts
```diff
@@ -81,9 +82,14 @@ export default class EditorJS {
...
+        if (Object.prototype.hasOwnProperty.call(editor.moduleInstances, moduleName)) {
+          const moduleInstance = editor.moduleInstances[moduleName];
+          if(moduleInstance.destroy && _.isFunction(moduleInstance.destroy)) {
```
gohabereg, line 88: I think its better to add check for `undefined` inside the `isFunction`

package.json
sis-dk: Oh ok

src/codex.ts
sis-dk: No specific reason. Object.values looks cleaner. Will change.

src/codex.ts
sis-dk: Will remove the 'moduleInstance.destroy' check from the if condition. Is a separate check in isFunction required? I can leave the isFunction code as it is and the typeof check should take care even if I pass undefined?

Issue #1153, opened 2020-05-13
enableLineBreaks does not work [Bug]


flaming-cl:
Hi, under certain circumstances, enableLineBreaks doesn't seem to work. 

Steps to reproduce:
1. within a block, select a few sentences with your mouse like this, 

2. press Backspace to delete your selected content
3. press Enter, no line breaks inserted

Thanks for replying 😊

sis-dk:
You set enableLineBreaks to true?

sis-dk:
You mean Shift+Enter right? It looks like it happens when the inline toolbar is open. After you press backspace, if the inline toolbar is still open, shift+enter doesn't seem to be working. Just adding my observation.

flaming-cl:
> You mean Shift+Enter right? It looks like it happens when the inline toolbar is open. After you press backspace, if the inline toolbar is still open, shift+enter doesn't seem to be working. Just adding my observation.

Hey, you are right! It does not work because inline toolbar is still open. Thanks a lot!!

PR #1154, opened 2020-05-14
<fix> enableLineBreaks opened, inline toolbar was not closed as expected

flaming-cl:
issue: https://github.com/codex-team/editor.js/issues/1153

Hi, under certain circumstances (when or immediately after mouse selection), enableLineBreaks doesn't seem to work.

This happened when inline toolbar was not closed properly. As the inline toolbar was still open, the Enter event could not be fired and no line breaks being inserted. 

It seems that, when selecting a few sentences within a block, 「BlockSelection.anyBlockSelected」was not true. This may be because such selection will not set the CSS style「ce-block--selected」to the current blocks.


sis-dk:
Line breaks when Shift+Enter is pressed is done by the browser on a contenteditable region. When the inline toolbar is opened, something called a flipper is activated (on inline toolbar close, its deactivated). There are two preventDefault calls in the flipper code which stops the event from reaching the event listeners that the browser has for Shift+Enter when the flipper is active and thereby stopping the browser from handling the line break.

preventDefault calls:
https://github.com/codex-team/editor.js/blob/release/2.18/src/components/flipper.ts#L94
and
https://github.com/codex-team/editor.js/blob/release/2.18/src/components/flipper.ts#L261
Try commenting them out or avoid those preventDefaults when event.shiftKey is true. The line breaks will start working.

Ideally, the toolbar should be closed by https://github.com/codex-team/editor.js/blob/release/2.18/src/components/modules/ui.ts#L335 when you press backspace, but the 'selectionchange' event sometimes doesn't fire on backspace and sometimes does. Its related to this bug i guess https://bugs.chromium.org/p/chromium/issues/detail?id=725890 on chrome. Tried in Safari and it always fires and always the inline toolbar closes on backspace.

Just telling the possible reasons why it doesn't work when inline toolbar is open.

neSpecc:
Seems outdated. Thanks for the investigation.

PR #1164, opened 2020-05-23
Feature/disable tab event config

flaming-cl:
Added an internal setting disableDefaultTabEvent to disable Editor/browser Tab Event
If users want to add their own tab event like inserting tab indents, they can disable the original tab event by using this config.


neSpecc:
Great, will check it soon.

neSpecc:
@flaming-cl Now you can add own keydown handler in your tool and override Tab behaviour

flaming-cl:
> @flaming-cl Now you can add own keydown handler in your tool and override Tab behaviour

Great! This is much better than adding another internal setting. Thanks!!

mkeida:
Hi! How exactly can I override default tab key behaviour? I need TAB for indents in my code block.

PR #1201, opened 2020-06-14
<fix> toolbar--opened overlap with certain text [issue 1196]

### LINKED ISSUE #1196: [Bug] Toolbar is penetrated by text
About the input text is Chinese, the add button in front of the block does not disappear, and then click the add button will appear behind the text penetrated the toolbar.

Steps to reproduce:
1. Start a new line by pressing Enter
2. Start typing Chinese characters (note: the add button in front does not disappear at this time)
3. Here, click the add button in front with the mouse, and the toolbar will be penetrated by the text.

Expected behavior:
When entering Chinese, the front add button should disappear.

Screenshots:
image

Device, Browser, OS:
PC, Chrome 83.0.4103.97,  MacOS 10.15.4
Editor.js version:
2.18

flaming-cl:
Resolves #1196 

### steps to reproduce:
1. Press "Tab" in an empty block to make toolbar opened 
2. type in something with input sources like Chinese, Japanese, etc.
### solution
add an extra condition for function **isPrintableKey**
### reasons
Unlike English, the languages mentioned above need a second processing after simply press a bunch of keys.
image
image
And when processing such languages and the event is keydown, the event Keycode will be 229.
https://lists.w3.org/Archives/Public/www-dom/2010JulSep/att-0182/keyCode-spec.html


neSpecc:
> I think fixing the issue for only Chinese or any particular language is too specific, maybe we should try to find a more general solution which will cover all the languages with which the issue exists. For example, inserting the range of all these keyCode values associated with languages which has the issue.

The 229 keycode covers not only Chinese. It means that the key was handled by Input processor.

https://lists.w3.org/Archives/Public/www-dom/2010JulSep/att-0182/keyCode-spec.html

hyun-park:
Including Korea. 🙁

PR #1205, opened 2020-06-17
Return the result of block.call

jacobsmith:
This change allows blocks to return the result of `call` methods, thus allowing them to expose arbitrary data as needed.

My particular use case is I am using Vue to mount components inside of the larger editorjs framework. One of the components that we are developing can be thought of as a nested agenda, where labels need to be in an order like:

```
I. Top level
  a. second level
    i. third level
```

My plan is to have an orchestrator query all blocks, filter those that need labels prepended, and then programmatically tell each block (with another `call` method) to set its depth to the desired level. At that point, Vue can reactively update any labels, etc. that are needed, allowing us to use a more declarative form of programming.

I believe this change will allow for other such uses, and I imagine it should not break any existing code since it was returning `null` before.


jacobsmith:
My current workaround for this involves a currying function:

In the orchestrator:
```
            const blockCount = this.editorJsInstance.blocks.getBlocksCount();
            for (let i = 0; i < blockCount; i++) {
                const block = this.editorJsInstance.blocks.getBlockByIndex(i);

                let component;
                // have to do some weird currying because `block.call` does not currently return the value of the function
                block.call('setComponent', (vueComponent) => component = vueComponent);

                // do whatever with the component
```

In the tool:
```
    setComponent(setter) {
        setter(this.component);
    }
```

It's a bit unwieldy, and definitely not super obvious, but does expose the variables as needed.

ranemihir:
@jacobsmith We are not understanding what is that you are actually trying to implement exactly with this PR. Can you provide an detailed and understandable explaination about the PR, maybe provide a small GIF, image or video. 
Also, create an issue and assign it to the PR.

src/components/block/api.ts
```diff
@@ -81,10 +81,11 @@ function BlockAPI(block: Block): void {
      * @param {string} methodName - method to call
      * @param {object} param - object with parameters
      *
-     * @returns {void}
+     * @returns {any}
      */
-    call(methodName: string, param?: object): void {
-      block.call(methodName, param);
+    // eslint-disable-next-line
```
neSpecc, line 86: I don't like an idea to ignore this rule. Why not use the `unkown` type?

ranemihir:
@jacobsmith Assign an issue to this PR.

src/components/block/api.ts
ranemihir: How do I ignore `eslint(jsdoc/no-undefined-types)` for `unknown` type in js-doc comment?

ranemihir: Also, there is no issue assigned to this PR. So I just added the link to this PR in the CHANGELOG.md file, is that ok?

PR #1207, opened 2020-06-18
Add a Fixed Title Block at the top

### LINKED ISSUE #1189: 💡Title block
Could you consider to implement title block that can take only the first position? Another option is to limit the H1 header one per editor.

ranemihir:
Fixes #1189 

Add a fixed title block which cannot be deleted or removed by the end user.
This functionality can be enabled by setting the **'fixedTitleBlock'** property to true.
It can also be intialized with an Object containing the **placeholde**r property which
will be used as the placeholder text for the title block.
The block uses a textarea with autosize module to grow as the text size increase.
The **'OutputData'** class changes and adds another property called 'title'
which contains the title provided by the user.
Screenshot 2020-06-18 at 11 05 37 PM

It cannot be added to the 'blocks' property itself because then it would have not been
at a fixed position and removable by the end user.
There is not way found to sanitize the data as EditorJS.Sanitizer was not a property and
in order to intialize Sanitizer module Editor config object requires which is only present in
core.ts and codex.ts and cannot be sent to saver making it not possible to save the data.
The documentation will be added once the change is approved.


azaitsev:
@ranemihir thank you so much for implementing that! Could I expect it will be merged or it doesn't fit project vision?

ranemihir:
@azaitsev I appreciate that you like the feature! :) But the I am not the official contributor of the project. Also, there is not way that I could find to sanitize the data. I did send them a personal email regarding this and the CodeX team told me that they would review the PR ASAP.

azaitsev:
Is there any chance it will be merged?

gohabereg:
Hi @ranemihir @azaitsev 
This is cool feature, but hardcoding title is too specific. User might need to fix any block, so it's better to pass data for any tool user plugged in to the config.

Suggestion from me: add `fixFirstBlock` property to the config and disable move and delete tunes for the first block. 
Usage example:
```
new EditorJS({
  tools: {
    header: HeaderTool
  },
  data: {
    time: 1595875684516,
    blocks: [
      {
        type: 'header',
        data: {
          text: ''
        }
      }
    ]
  },
  fixFirstBlock: true
})
```

ranemihir:
@gohabereg Yes, you are right. Peter actually told me about this in today's meeting. I totally agree.
Can you tell me how to disable move and delete tunes. That would be a big help. 
I will try to send the new PR asap.

gohabereg:
@ranemihir tunes are set up in Block class. But Block class doesn't know if it first or not. So you need to pass tunes to Block from BlockManager as constructor paremeters (just like data and settings). If block is the first one, just pass an empty array

azaitsev:
Maybe we wanna make a set of fixed blocks in the desired order? Something like:
```new EditorJS({
  tools: {
    header: HeaderTool
  },
  data: {
    time: 1595875684516,
    blocks: [
      {
            "type" : "image",
            "data" : {
                "caption" : "Select post cover",
            },
            fixedPosition: 0,
        },
      {
        type: 'header',
        data: {
          text: ''
        },
        fixedPosition: 1,
      }
    ]
  },
})```

**Comment by gohabereg** [comment]
Specifying order is redundant as array already sets it. Also, fixed blocks is a property of the editor itself, not tools.

However, we just can set amount of fixed blocks we want
```
New EditorJS({
  ...
  fixedBlocks: 3,
  ...
});
```

ranemihir:
@gohabereg , I think @azaitsev is right here. If fixed block property does not take the tools in consideration then it is pretty much useless. 
Think about the use cases, people want a fixed block for a fixed title for their content or for making say Google forms like structures (as @neSpecc suggested). In all of these use cases, the order and the tools which are being used have to be fixed. 
Setting just the number of fixed blocks isn't gonna help in these scenarios but it requires providing the tools and their fixed positions.

gohabereg:
You can provide order and needed tools  in the `blocks` array

ranemihir:
@gohabereg Oh yeah, right! The number of blocks will represent the number blocks in the blocks property. Ok I get it now.

neSpecc:
This PR is too specific for merging, so I'll close it.

This feature should be designed and described via text at the related issue.

PR #1208, opened 2020-06-19
Change cursor on placeholder to text from default

### LINKED ISSUE #1165: 💡Use 'text' cursor when there is the only placeholder
1. Describe a problem.
When placeholder is displaying it is not intuitive that you can click and edit that content because of the cursor is in standard style
telegram-cloud-photo-size-2-5352945091536334306-x

Make cursor: text for editor block when there is the only placeholder
telegram-cloud-photo-size-2-5352945091536334308-x

ranemihir:
Fixes  #1165 
Previously, the cursor on placeholder was default which is changed to text.


src/styles/export.css
```diff
@@ -30,6 +30,13 @@
   }
 }
 
+/**
+ * Add text cursor on placeholder 
+ */
+.ce-header[contentEditable=true][data-placeholder]:empty::before {
```
neSpecc, line 36: `ce-header` is a hardcode. You need to add `cursor` to the editor's element, not a plugin

ranemihir: I think the problem only exists with the Header tool itself. I think @gohabereg is right that it should be moved to the Header tool itself.

ranemihir:
@gohabereg I have created the PR in the header repo. Check it out at [https://github.com/editor-js/header/pull/51](https://github.com/editor-js/header/pull/51).

Issue #920, opened 2019-10-18
Bug with inputs


Obzhigalov:
Hi there! I created my own block tool for Editor. In my tool you have to input some text, then click "save" button, then it will render. So everything is ok, besides input. When I press Enter on the keyboard when input is focused - it creates a new block with this input. Maybe it's my bad, maybe its a bug in Editor. I have the same issues with every input in my custom block tools. Here is the video https://recordit.co/bvGKcGuRav

hata6502:
I encounted same problem too!
But this bug doesn't appear with editor-js/image's caption input. 
https://github.com/editor-js/image

editor-js/image uses div for caption input. 
Modified div to input, this bug appeared. 

![Screenshot from 2019-12-07 11-56-05](https://user-images.githubusercontent.com/7702653/70367962-e3b52400-18e8-11ea-899a-695920508fca.png)

So, I think this bug can be avoided to use &lt;div contenteditable="true"&gt; !

Dacesilian:
I confirm this bug, there is my video of it in my plugin - https://youtu.be/A1-C6L4-tkw .
Is it planned to fix this, please? I've tried to bind keydown/keyup/keypress listeners but preventDefault is not working.

karpovsystems:
`event.stopPropagation()` worked for me

```javascript
input.addEventListener('keydown', event => {
    if (event.keyCode === 13) event.stopPropagation();
})
```

For creating a block after pressing enter I used

```javascript
input.addEventListener('keydown', event => {
    if (event.keyCode !== 13) return;
    event.stopPropagation();
    this.api.blocks.insert('paragraph');
    this.api.caret.setToBlock(this.api.blocks.getCurrentBlockIndex());
})
```

sethmichaelbrown:
I am still seeing this bug in v2.22.2. I am creating a multiline MUI input via a custom plugin. The plugin renders a target `div` then I use `ReactDOM.render` to append a React component to the target. 

Any time I hit enter within the MUI multiline text area, I get a new block. I have tried listening for `keydown` and using `e.stopPropagation()` as well as `e.preventDefault()` and `return`. None of these seem to work. 

Any thoughts on how I might keep focus within the `textarea` to create a new line? Thanks.

PR #1214, opened 2020-06-22
Fix for input and textarea bug

PavlovicWorkCo:
This PR should fix Issue #920 . 
A bit more detail about the issue:
Inputs and text areas are not handled properly.
In cases in which we want to have multiple text inputs in one custom plugin, some of them with formatable text and some of them without, we need to use inputs and/or text areas (for those which we don't want to have formatable text). In the current version this causes multiple bugs when hitting enter in inputs and textareas, because the `currentInput` is not updated properly and because the text in inputs and textareas is not innerHTML (it's in their value attribute).
Bugs:
1) It copy pastes the whole `<input>` element and puts it in the next paragraph outside of the block instead of only cutting out the text from the caret to the end. 
2) It cuts out the `<input>` element (without the text in it) and copies the whole content editable element.
input-bugs


hata6502:
I support this PR!

PavlovicWorkCo:
@neSpecc @khaydarov @gohabereg Hi. Could you take a look at the issue and PR, and let me know if anything needs changing and if and when you would consider merging it. This bug is a blocker for a project that is soon going into production, so we would very much appreciate if it could be resolved without having to fork the repo. 🙏

neSpecc:
> @neSpecc @khaydarov @gohabereg Hi. Could you take a look at the issue and PR, and let me know if anything needs changing and if and when you would consider merging it. This bug is a blocker for a project that is soon going into production, so we would very much appreciate if it could be resolved without having to fork the repo. 🙏

we will check it soon. The next release is planned for the end of July.

src/components/utils.ts
```diff
@@ -556,3 +556,12 @@ export function getValidUrl(url: string): string {
 export function openTab(url: string): void {
   window.open(url, '_blank');
 }
+
+/**
+ * Checks if element is "<input />" or "<textarea />"
+ *
+ * @param element - Element to check
+ */
+export function isNativeTextInput(element: Node | HTMLElement): boolean {
```
neSpecc, line 565: we already have such util, see `dom.ts@isNativeInput`

neSpecc (changes requested):

gohabereg (approved):

neSpecc (approved):

Issue #1182, opened 2020-06-03
[Bug] -  Error during delete a block by index


nicolapagni:
I've installed the v2.18 to try the new feature about deleting specific block by index.
I can notice that now I see a warning if I try to delete a block by an invalid index..ok, but now I see this error:

Instead, if I try to delete though the same code the firt element It seems work fine..

` Cannot read property 'lastInput' of undefined
`
Steps to reproduce:
1. Create a new editor instance
2. As first block (for example), insert a paragraph with something write
3. As second one insert a paragraph too
4. call after adding the second one the following code: `this.api.blocks.getBlockByIndex(1);`

Expected behavior:
The block should be removed.

Screenshots:
<img width="421" alt="editorjs-error" src="https://user-images.githubusercontent.com/17281895/83647219-f4eb9e80-a5b4-11ea-92a2-953b7ca8cbaa.png">

main.js is the file where I call the delete function.

Device, Browser, OS: 
Chrome v83.0.4103.61, Windows 10

Editor.js version: 2.18

stephenrichard:
Hello @nicolapagni , I encountered the same issue working with the new `delete` feature.

After digging into the code, I found that after deleting a block this line is called : `this.Editor.Caret.setToBlock(this.Editor.BlockManager.currentBlock, this.Editor.Caret.positions.END);` 

It's supposed to move the cursor focus after the current block. **The issue is that `this.Editor.BlockManager.currentBlock` is `undefined`**. In my cas I'm not really surprised because I am adding/removing blocks without the need to interact with the editor manually.

2 options I see to fix this issue : 
- Quickest -> Add a condition around that line if `this.Editor.BlockManager.currentBlock` exists
- Quickest *(but possible unwanted behaviour)* -> same as above but also define a `else` condition that moves the cursor after the last block using `this.Editor.Caret.setToBlock(this.Editor.BlockManager.lastBlock, this.Editor.Caret.positions.END);`

I can submit a PR with those changes any time but I would love some feedbacks from maintainers or other users ✌️

ranemihir:
@nicolapagni We followed the steps as you have mentioned but we couldn't able to reproduce the issue. Can you provide any further details regarding this issue, maybe a GIF or a sall video to show the exact process of reproducing the issue.

stephenrichard:
@ranemihir 
I reproduce the issue with those steps :

1. Create a new editor instance (with sample data, 2 blocks is enough)
2. Do not edit, focus or add a block in the editor
[...]

PR #1218, opened 2020-07-03
Fix blocks.delete with undefined index (#1182)

stephenrichard:
In some situations `this.Editor.BlockManager.currentBlock` is not defined. (For example if the user didn't interact with the editor yet but can add/remove blocks outside via buttons using the API)

Capture d’écran 2020-07-03 à 18 22 36

When deleting a block, the Caret (cursor head in my understanding) tries to move to the currentBlock (that can be `undefined`) and trigger a legitimate error.

This PR simply adds a condition around the Caret that checks for a currentBlock :
- If it's defined, move to the currentBlock
- If not, just do not move


ranemihir:
@stephenrichard can you create an issue regarding this so that we can test if the issue actually exists?

stephenrichard:
@ranemihir Yes, sorry I forgot to link an issue to this but it refers to #1182

ranemihir:
@stephenrichard Thank you. I will test it and let you know.

stephenrichard:
@neSpecc Sure, how should I proceed ? I checkout this PR, edit the CHANGELOG.md and push the changes ? (It won't revoke the approvals ?)

PR #1222, opened 2020-07-06
Add debounce time for the editor onChange callback

Teebo:
Hi, I trust you are doing well.
Thank you for EditorJS, it is such a great tool.
I was not sure where to find guides for contributing.

### Requirement case
I would like to have a property on the configuration to configure the debounce time for the Editor's onChange callback.

I would like to handle the debounce because I am saving the Editor data to Firebase, so I want to limit the number of times the update is made to a document when the user interacts with the editor.

I have tried an approach on the pull request.
Please let me know if the changes qualify and how I can improve them.

Thanks.


hata6502:
May be related: https://github.com/codex-team/editor.js/issues/1225

src/components/core.ts
```diff
@@ -194,6 +194,7 @@ export default class Core {
     this.config.onReady = this.config.onReady || ((): void => {});
     // eslint-disable-next-line @typescript-eslint/no-empty-function
     this.config.onChange = this.config.onChange || ((): void => {});
+    this.config.onChangeDebounceTime = this.config.onChangeDebounceTime || undefined;
```
gohabereg, line 197: If `onChangeDebounceTime === 0`, result of the expression id undefined, so there is no way to set 0 as value

src/components/modules/modificationsObserver.ts
```diff
@@ -18,7 +18,7 @@ export default class ModificationsObserver extends Module {
    *
    * @type {number}
    */
-  public static readonly DebounceTimer = 450;
+  public DebounceTimer = this.config.onChangeDebounceTime || 450;
```
gohabereg, line 21: The same as above

src/components/core.ts
Teebo: Thank you @gohabereg, the `undefined` here is handled on the changes I have made in the modificationsObserver.ts file

src/components/modules/modificationsObserver.ts
Teebo: @gohabereg please refer to commit b8137a1c312e837eb44eb08a35415dc99e841a81

neSpecc:
@Teebo please, attach an issue describing a problem you resolve.

Anyway, why not just to add own debounce on the your callback? Adding a special option for this case looks redundant.

Teebo:
@neSpecc my issue is as follows:

I have tried the suggested approach but it does not work.
I have used lodash's debounce, but for every change fired the debounced operation will be registered and thus be called, even though debounced, see below.

```
const saveOperation = () => {
  console.log('in operation');
 }
...
onChange: () => {
    console.log('the on change');
    _.debounce(saveOperation, 2000)();
}
...
```
editor-log

sis-dk:
@Teebo You are using the onChange and _.debounce wrong. You don't want a new debounced function every time onChange is called. You want it only once. I've given you an example [here](https://github.com/codex-team/editor.js/issues/1238#issuecomment-669197466)

Teebo:
@sis-dk the `saveOperation` is defined outside the onChnage scope.
Please try it

Teebo:
@neSpecc @sis-dk don't you guys think that as a user I'd want to control how often or when the onChange fires without having to use an external function like lodash? It is not entirely redundant if you look at it

neSpecc:
I don't think that adding new config options for every such small case because that way the Editor's config will come large and hard-to-learn.

You can increase the `onChange` debounce by adding it on your own side. 

I'll close this PR. If the problem will actual for more people, please create a separate issue. We can reopen the PR in that case.

Teebo:
@neSpecc no problem, thank you.
@sis-dk did you succeed with the example? I'd like to know where I am lacking

sis-dk:
> @sis-dk did you succeed with the example? I'd like to know where I am lacking
Hi @Teebo IDK if its too late to help. I didn't notice your comment. If you still have the problem, reply to this comment and I'll help you out.

Teebo:
@sis-dk its not too late, please help me out with a working example, maybe put it on stackblitz?

Teebo:
@neSpecc you still don't see a need for that debounce property?
It would be really great to have it, I'm begging you or whoever owns this project

hata6502:
I've tried using lodash.debounce: https://github.com/codex-team/editor.js/compare/next...hata6502:extend-onchange-debounce?expand=1

I think ... the debounce time in Editorjs is a magic number. 
So fixing Editorjs's debounce time to 450 is a safe way currently.

marek-miotelka:
Hi, what if I want to lower the debounce time?

I want to implement character limit feature where user shouldn't be able to add text over the specified limit.

WIth current debounce time he is able to do it, and then the content he added is removed. 
Also the character counter is lagging, cause it's picking up the changes 450ms after a user stopped typing.

PR #1224, opened 2020-07-07
Add onBlur event

hata6502:
https://github.com/codex-team/editor.js/issues/1225


neSpecc:
The `onBlur` does not resolve the attached issue, because the content of the editor can be modified not only through the typing from the keyboard. All other changes like blocks moving, changing block tunes, and others won't fire the `onBlur` event. It means that you can't use it for triggering a saving.

I'll close the PR for now. If someone will have a use-case for the `onBlur` callback, please raise the issue describing your case. Maybe we will reopen this PR later.

tom2strobl:
@neSpecc While I agree with you that it doesn't solve the issues some people might believe it would, it still gives power to the people. In my case in a react scenario I re-initialize the editor when I have new data from the server. If I use onChange (even debounced) as soon as a successful write happens my user loses editor focus because the editor is re-rendered. Which is fine if blocks were moved, but sucks when, well, when the editor is focused. If the debounce is too short I could be interrupted even mid sentence, if its too long I risk navigational side effects.

If I knew that the editor is focused/blurred, I can write React logic to not re-initialize the editor until the editor is blurred and things would be solved rather cleanly without the use of a dedicated save button.

PR #1227, opened 2020-07-08
Fix #1183 / Expose src/types-internal definitions to npm.

### LINKED ISSUE #1183: Error when trying to integrate Editorjs with Angular
This is the error I get: 
---------------------------------------------------------------------
ERROR in node_modules/@editorjs/editorjs/types/api/block.d.ts:2:25 - error TS2307: Cannot find module '../../src/types-internal/block-data'.

import {SavedData} from '../../src/types-internal/block-data';
---------------------------------------------------------------------

So basically it's trying to import from src folder, but there is no src folder in the @editorjs/editorjs when i use the 'npm install @editorjs/editorjs --save-dev'. 
I am aware that there is an ngx-editorjs package, but I'm really frustrated by this issue.
Any help would be highly appreciated.

RicoBrase:
Should fix codex-team/editor.js#1183

I am aware of the fact, that the directory is literally called `types-internal` and therefore probably not meant to be available via npm.

BUT - you are including at least one of those type definitions in your published type definitions, e.g. `/src/types-internal/block-data.d.ts` in `/types/api/block.d.ts`, so at least the `block-data.d.ts` has to be published via npm as well, otherwise it will lead to errors like #1183.

This fix is meant as an easy workaround without modifying too much of the project or it's folder structure.


gohabereg:
Hi @RicoBrase,
Thanks for contribution! However, we don't want anything from src folder to be in the package.

I move these types to `types` folder here — https://github.com/codex-team/editor.js/pull/1251/files

RicoBrase:
@gohabereg Fair enough and quite possibly the best solution.

I'll close this PR for now.

PR #1248, opened 2020-07-25
Add RTL support

### LINKED ISSUE #670: Add RTL support

ImangazalievM:
Resolves #670

Copied from https://github.com/codex-team/editor.js/pull/909


talyguryn:
Link for testing: https://deploy-preview-1248--editor-js.netlify.app/example-dev-rtl.html

ImangazalievM:
1.I fixed the positioning of icons in the toolbar.
2. Unfortunately, there is nothing I can do about text selection, this is the standard behavior when mixing RTL and LTR text

talyguryn (approved):

ranemihir (approved):

example/example-dev-rtl.html
```diff
@@ -0,0 +1,233 @@
...
+  var editor = new EditorJS({
+    /**
+     * Wrapper of Editor
+     */
+    holder: 'editorjs',
+
+    /**
+     * Text direction
+     */
+    direction: 'rtl',
```
neSpecc, line 92: let's move this option to the `i18n` section, near `messages`

neSpecc (changes requested):

example/example-dev-rtl.html
```diff
@@ -0,0 +1,233 @@
+<!--
```
neSpecc, line 1: let's name this file `example-rtl.html`

neSpecc (changes requested):

neSpecc (changes requested):
There is an issue with arrow navigation between blocks. And there is no ability to move caret by arrows if it is at the start or at the end of the block.

For now, it checks the end of the block for `right` arrow and the start of the block for `left` arrow.

See blockEvents.ts `arrowRightAndDown` and `arrowLeftAndUp`

src/components/__module.ts
```diff
@@ -46,4 +46,12 @@ export default class Module {
   public set state(Editor: EditorModules) {
     this.Editor = Editor;
   }
+
+  /**
+   * Returns true if current direction is RTL (Right-To-Left)
+   */
+  protected isRtl(): boolean {
```
neSpecc, line 53: maybe it would better to use getter instead of method?

src/styles/rtl.css
```diff
@@ -0,0 +1,74 @@
+.codex-editor.rtl-fix {
```
gohabereg, line 1: Why is it called `fix`? Maybe just add a modifier: `.codex-editor--rtl`

neSpecc (commented):
Almost ready, but it still has some issues:

In the "narrow mode" (which enables when editor inserted in small containers) there are some issues. Try to remove underscore from the `ce-example__content--small` class at example-rtl.html.

The narrow mode is used in small containers, where is no place for the Plus button and for the Block Settings button. It have two features:
- the plus button is placed inside the editor column

image

- additional padding-right added for the block settings button

image

We need to adopt these moments for the RTL

neSpecc (approved):

gohabereg (approved):

talyguryn (approved):

javadbat:
can we use it now?

PR #1264, opened 2020-08-04
Tools destroy called when the editor is destroyed

### LINKED ISSUE #1047: Tools-API destroy is not invoked when editor.destroy() is triggered
## The Problem

According to the docs, when I do the following:

```js
/**
* Create an instance
*/
const editor = new EditorJS();

/**
* Destroy editor.
*/
editor.destroy();
```

all `destroy` methods of the tools should be invoked.

> destroy
> Method will be fired when Editor's instance is destroying with destroy API method. Clear your Tool's stuff here: remove event listeners, clear nodes cache and null variables and properties.
Docs: https://editorjs.io/tools-api#destroy

But, this does not happen.

Not sure if this is what was reported here already: https://github.com/codex-team/editor.js/issues/855.

## Reproduce

1. Add the following to `example/tools/header/src/index.js`

```js
  static get toolbox() {
    // ...
  }

  destroy() {
    console.log('I was destroyed');
  }
```

2. Now run `yarn build:dev` and execute `editor.destroy()` in the console

3. Nothing happend, even though the console should now say `I was destroyed`.

sis-dk:
When the editor instance is destroyed, it calls the destroy of the blockManager. blockManager in-turn calls destroy of all the blocks that it manages.

fixes #1047


src/components/modules/blockManager.ts
```diff
@@ -667,6 +667,18 @@ export default class BlockManager extends Module {
     this.Editor.UI.checkEmptiness();
   }
 
+  /**
+   * Cleans up all the block tools' resources
+   * This is called when editor is destroyed
+   */
+  public destroy(): void {
+    this.blocks.forEach((block) => {
```
ranemihir, line 675: Using `await Promise.all(this.blocks.map(block) => { /*further code*/ })` would be more accurate as it will cover async tasks if present in any of the `destroy` functions.

ranemihir:
@sis-dk Please add your changes to the CHANGELOG.md file with the link to the issue.

src/components/modules/blockManager.ts
sis-dk: Sure I'll do it tonight (Saturday night IST). That'll require a chain of async functions I rekon. I'll have a look at it and get back to you.

ranemihir: Please also add your changes to the CHANGELOG.md file with the link to the issue.

gohabereg:
@neSpecc @khaydarov @ranemihir @sis-dk 
This is essential feature. However, it covers only Tools instances. So I think it's better to just call `REMOVED` hook for this case.
But we still need some way to tell tools that Editor is destroyed (to remove external scripts for example). For that we should use `static destroy()` in tools and call it for each tool in the Tools module.
What do you think guys?

gohabereg (approved):

docs/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - `Improvements` - The `initialBlock` property of Editor config is deprecated. Use the `defaultBlock` instead. [#993](https://github.com/codex-team/editor.js/issues/993)
 - `Fix` - Fixed the issue of toolbar not disappearing on entering input in Chinese, Hindi and some other languages. [#1196](https://github.com/codex-team/editor.js/issues/1196)
 - `Improvements` - BlockAPI `call()` method now returns the result of calling method, thus allowing it to expose arbitrary data as needed [#1205](https://github.com/codex-team/editor.js/pull/1205)
+- `Fix` - Tool's `destroy` method is not invoked when `editor.destroy()` is called. [#1047](https://github.com/codex-team/editor.js/issues/1047)
```
neSpecc, line 10: also, we need to add a line about the new `reset` method

docs/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
...
+- `Improvements` - Tool's `reset` static method added to the API to clean up any data added by Tool on initialization
```
neSpecc, line 11:
```suggestion
- `New` - Tool's `reset` static method added to the API to clean up any data added by Tool on initialization
```

neSpecc (approved):

khaydarov (approved):

PR #1078, opened 2020-03-23
[Feature] Block Tunes API

gohabereg:
Things left:

- [ ] Refactor
- [ ] Write docs
- [ ] Add common tunes support
- [ ] Render common and Block tunes in separate containers

PR #1272, opened 2020-08-11
feat: added alignment tune

semoal:
Hello guys, i've introduced a starter for block aligment. 

Actually the classes are just `text-align`, probably this should work for 90% of cases, but for some images, probably we'll need better styles like flex, or align-self. 

I'm open to criticism, first time contributing and I've been a bit lost at the start.


gohabereg:
Hi @semoal !
Thank you for your contribution! Do you think alignment should be on the core level? Isn't it related to a Tool itself?

semoal:
> Hi @semoal !
> Thank you for your contribution! Do you think alignment should be on the core level? Isn't it related to a Tool itself?

Well actually, for me the best approach would be a tune extension that can be loaded for x tool, for ex:

I load the alignment tune to -> paragraph, text, and list tools.

But i'm not sure if the api is allowed to do that with a lot of change, that's why I tough if we implement this on the core, but customizable, like a config option:

```
var editor = new EditorJS({
      holder: 'editorjs',
      tunes: {
       aligment: {
           enabled: [header, paragraph, list, table, ....] // any tool that can be aligned.
         }
     },
      tools: {
        header: {
          class: Header,
          inlineToolbar: ['link'],
          config: {
            placeholder: 'Header'
          },
          shortcut: 'CMD+SHIFT+H'
        },

```

neSpecc:
I think, these tools should be implemented as separate Block Tunes plugins when the [Block Tunes API](https://github.com/codex-team/editor.js/pull/1078) will be released

semoal:
> I think, these tools should be implemented as separate Block Tunes plugins when the [Block Tunes API](https://github.com/codex-team/editor.js/pull/1078) will be released

Should i help on opening the API?

gohabereg:
Hi @semoal 
Your help is welcome! 
Work on Tunes API was freezed so it should be reviewed and designed considering the changes. If you are ready to contribute I'm glad to help.

I'm closing this PR as this functionality definitely should be implemented by plugins or via Tunes API

Issue #1280, opened 2020-08-24
Prevents blocks sanitizing for undo plugin


sos-productions:
I don't konw how to desactivate sanitizing

I described the problem https://github.com/kommitters/editorjs-undo/issues/16

Any idea welcome. Thanks

gohabereg:
Hi @sos-productions 
It seems, you need an ability to customize tools sanitizer configuration.

At the moment for Paragraph Tool (or any other) you can override it's `sanitize` property and pass to the editor:

```
import Paragraph from '@editorjs/paragraph';

Object.defineProperty(Paragraph, 'sanitize',  {
  get() {
    return {
      text: { br: true, strong: true, ... }
    };
  }
});

new EditorJS({
  tools: {
    paragraph: Paragraph
  }
});
```

Or fork a tool and change `static get sanitize()` return value in sources.  

The right way to overcome this issue in general — add ability to pass `sanitize` property to each Tool on editor initialization:

```
new EditorJS({
  tools: {
    paragraph: {
      sanitize: customSanitizeConfig
    }
  }
});
```

Disabling sanitize at all because only one plugin is unsafe.

sos-productions:
hi gohabereg, I need to take a snapshot of the blocks for a history step whatever the content, at the beginiing it is done with undo.initialize call , later is done with .save() in registerChange() inside the Undo plugin.

 so I need to disable sanitize temporarely to avoid crunching the tags to be able to go back on previous state on undo.undo() or redo.redo() smootthly.

Now my init of the Undo plugin is quite simple:

`    /**
        Initialize undo stack
    **/
    function initUndo(editor) {
    
        const undo = new Undo({editor});
 
        /*true is to avoid sanitizing Blocks, requires to patch editorjs*/
        editor.save(true).then((initialData) => {
            undo.initialize(initialData);
        }).catch((error) => {
            console.log('Saving failed: ', error)
        });
    
        editor.undo=undo;
    }`

This problem occurs everytime a need to export throughout the editor.save() process whatever the tool. This is thus also the case for load ans save blocks to a .js local file by example.

block.save() is not affected by this sanitizing issue as there is no sanitize on the block content. So I tried  a loop from first to editor.blocks.getBlocksCount() with block.save() to would have solved this. 

[...]

Issue #1191, opened 2020-06-08
💡Render thousands of blocks freezes the app


mqtik:
1. Describe a problem.
Hi.
First of all, great editor, very powerfull across platforms.

We have a situation, where we need to render thousands of blocks, like so:
![image](https://user-images.githubusercontent.com/42002579/84062423-a283f680-a995-11ea-8edd-3f672a454ae4.png)

That one has a bit over a thousand blocks, but we have several that has much morer than that.
Loading that freezes the app for a few seconds, making impossible to click anything even outside the Editor. When the memory used to load that kind of data by the Editor is released, then the app works again.

This is because the Editor does a .map(), which loads the whole data, instead of (maybe) load it in chunks relative to the scroll.

2. Describe the solution you'd like. Mockups are welcome.
Render blocks depending on the scroll. Like FlatList on React Native, which renders content relative to the actual scroll, and there's an option to even remove sub clipped views.

3. Are there any alternatives?
I can't think of any yet. Or, at least, none that can be done without modyfing the Editor's Core.

mqtik:
This would be a start:
https://github.com/codex-team/editor.js/issues/1192#issuecomment-672962053

PR #1281, opened 2020-08-25
Handles extra parameter sanitizeDisable for saving block(s)

sos-productions:
Fullfill my own request described here:https://github.com/codex-team/editor.js/issues/1280


build
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+#yarn has to be installed first
+#
+npm run-script build
```
hata6502, line 4: There are some unrelated files.

src/components/modules/api/saver.ts
```diff
@@ -14,14 +14,16 @@ export default class SaverAPI extends Module {
    */
   public get methods(): Saver {
     return {
-      save: (): Promise<OutputData> => this.save(),
+      save: (sanitizeDisable = true): Promise<OutputData> => this.save(sanitizeDisable),
```
hata6502, line 17: I think default sanitizeDisable should be `false` for safety.

hata6502 (commented):
I think the undo function is important for Editor.js :+1:

build
sos-productions: Helper file to build optional but I submitted an issue about github forking not complete where I completed this script to be able to build editorjs. see https://github.com/codex-team/editor.js/issues/1283

src/components/modules/api/saver.ts
sos-productions: yes you are right

build
hata6502: I read https://github.com/codex-team/editor.js/issues/1283, and commented. 
Editor.js uses git sub modules. 
https://github.com/codex-team/editor.js/issues/1283#issuecomment-680714370

src/components/modules/api/saver.ts
sos-productions: @hata6502 As I end up discussion https://github.com/codex-team/editor.js/issues/1280 sanitizing should better be focus on input and on server side, it is good however that a first stage of sanitizing on pasting, rendering. and on inserting blocks has been done but for large number of blocks this overload on client side gives this https://github.com/codex-team/editor.js/issues/1191 . However, this patch that imho has no security issue because we work not input but on output, readonly for block content .

neSpecc:
Outdated

PR #1290, opened 2020-08-28
Fix i18n default configuration

khaydarov:


ranemihir (approved):
I just tested it. It fixed all the errors that were showing previously.

robonetphy (approved):
Can you build and Commit here?

khaydarov:
> Can you build and Commit here?

done

neSpecc (approved):

Issue #1056, opened 2020-03-10
inline toolbar order


flaming-cl:
Hi, if we hope to change the order of the inline tools, how can we implement it?

For example, if we want Marker to be placed before all the other tools:
![image](https://user-images.githubusercontent.com/51183663/76331514-3825d080-632a-11ea-8dd6-b2903ab75606.png)

Thanks!

khaydarov:
reference #862

neSpecc:
1. Add `inlineToolbar` option to the root of EditorConfig. This common option will allow removing some tools from all blocks. Or disable the Inline Toolbar for all blocks by passing `false`. Or specify a sorting of tools.

2. If there is no root `inlineToolbar` option passed, get the order of the inline tools by their order in the `tools` config option appended to the internal (link, bold, italic).

3. The `inlineToolbar` option of specific tool (got from `tools<tool>` sections) will override the root setting.

4. Inside the `show` method of the Inline Toolbar perform buttons ordering corresponding by the current (tool or root) setting.

PR #1293, opened 2020-08-31
Add inlineToolbar property

ranemihir:
Tried implementing the `inlineToolbar` property according the instructions given by @neSpecc in this [comment](https://github.com/codex-team/editor.js/issues/1056#issuecomment-615901667) related to issue #1056.

I still don't understand how to set the tools in the inline toolbar according to the order given in the `inlineToolbar` property by the user. Can someone tell me how to do that? @neSpecc @gohabereg


src/components/modules/toolbar/inline.ts
```diff
@@ -363,9 +363,31 @@ export default class InlineToolbar extends Module {
 
     const toolSettings = this.Editor.Tools.getToolSettings(currentBlock.name);
 
-    return toolSettings && toolSettings[this.Editor.Tools.USER_SETTINGS.ENABLED_INLINE_TOOLS];
+    return toolSettings && (toolSettings[this.Editor.Tools.USER_SETTINGS.ENABLED_INLINE_TOOLS] == undefined || Array.isArray(toolSettings[this.Editor.Tools.USER_SETTINGS.ENABLED_INLINE_TOOLS]) || toolSettings[this.Editor.Tools.USER_SETTINGS.ENABLED_INLINE_TOOLS] == true);
```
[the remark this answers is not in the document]
ranemihir: ok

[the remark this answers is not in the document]
ranemihir: but what about sorting tools, how do i do that?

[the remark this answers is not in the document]
neSpecc: reorder buttons on each toolbar opening

src/components/modules/toolbar/inline.ts
```diff
@@ -393,11 +414,16 @@ export default class InlineToolbar extends Module {
 
     /**
      * Filter buttons if Block Tool pass config like inlineToolbar=['link']
+     * Else filter them according to the default inlineTOllbar property.
```
khaydarov, line 417:
```suggestion
     * Else filter them according to the default inlineToolbar property.
```

src/components/modules/toolbar/inline.ts
```diff
@@ -393,11 +414,16 @@ export default class InlineToolbar extends Module {
...
      */
     if (Array.isArray(inlineToolbarSettings)) {
       buttons.forEach((button) => {
         button.hidden = !inlineToolbarSettings.includes(button.dataset.tool);
       });
+    } else if (Array.isArray(this.config.inlineToolbar)) {
+      buttons.forEach((button) => {
+        button.hidden = !this.config.inlineToolbar.includes(button.dataset.tool);
+      });
     }
```
khaydarov, line 427: I'm not sure, but the lines above could be simplified. You're repeating yourself: the difference are condition and iterable object. Maybe create a new private function and pass buttons and hiddenList

khaydarov (changes requested):

src/components/modules/toolbar/inline.ts
ranemihir: Yes, to be honest, your thoughts did cross my mind but I couldn't think of a way to implement it. I will try to implement it as you have suggested.

ranemihir:
Ordering of tools is working fine but there is this error which keeps coming even though the code is working fine Beacause of which I can't build the code. 

Error: `Property 'dataset' does not exist on type 'ChildNode'.` 

Also, I know the code can also be optimised further. But can someone help me with this error? @neSpecc @khaydarov @gohabereg

gohabereg:
@ranemihir 
You need to cast variable to `HTMLElement` type

ranemihir:
@gohabereg Thanks man.

src/components/modules/toolbar/inline.ts
ranemihir: Can you check the recent commit and suggest more optimizations?

src/components/core.ts
```diff
@@ -223,6 +223,11 @@ export default class Core {
         direction: config.i18n?.direction || 'ltr',
       };
     }
+
+    /**
+     * InlineToolbar is disabled if set to false.
```
neSpecc, line 228:
```suggestion
     * The default order of Inline tools. Will be applied if user will pass the `InlineToolbar: true` through the tool options.
     * If it is false, the Inline Toolbar will be disabled for all the tools. 
```

src/components/modules/toolbar/inline.ts
```diff
@@ -393,11 +441,26 @@ export default class InlineToolbar extends Module {
...
-      buttons.forEach((button) => {
-        button.hidden = !inlineToolbarSettings.includes(button.dataset.tool);
-      });
+      inlineToolbar = inlineToolbarSettings;
+    } else if (Array.isArray(this.config.inlineToolbar)) {
+      inlineToolbar = this.config.inlineToolbar;
+    }
+
+    buttons.forEach((button) => {
+      button.hidden = !inlineToolbar.includes(button.dataset.tool);
```
gohabereg, line 456: This line will throw an error if `inlineToolbarSettings` or `config.inlineToolbat` is not defined or defined but not an array.

src/components/modules/toolbar/inline.ts
```diff
@@ -362,20 +362,68 @@ export default class InlineToolbar extends Module {
...
+          buttons[j] = temp;
+        }
+      });
+    });
+
+    /**
+     * appends all the button nodes in sorted order
+     */
+    buttons.forEach(async (btn) => {
+      this.nodes.buttons.appendChild(btn);
```
gohabereg, line 414: I thinks it would be more logical to return sorted buttons order from this function and append them to container somewhere else

gohabereg (changes requested):

src/components/modules/toolbar/inline.ts
ranemihir: check the latest commit

src/components/modules/toolbar/inline.ts
ranemihir: check latest commit

src/components/modules/toolbar/inline.ts
```diff
@@ -362,20 +362,63 @@ export default class InlineToolbar extends Module {
...
     const toolSettings = this.Editor.Tools.getToolSettings(currentBlock.name);
+    const inlineToolbarSettings = toolSettings[this.Editor.Tools.USER_SETTINGS.ENABLED_INLINE_TOOLS];
 
-    return toolSettings && toolSettings[this.Editor.Tools.USER_SETTINGS.ENABLED_INLINE_TOOLS];
+    /**
+     * Checks if either tool's own inlineToolbar property or config.inlineToolbar property is is an array.
+     * If both are nanything other than arrays, then no inlineToolbar will be displayed.
+     * As there is no way to know which tools or order of tools user wants.
+     */
+    const isInlineToolbarArray = Array.isArray(inlineToolbarSettings) || Array.isArray(this.config.inlineToolbar);
```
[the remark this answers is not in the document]
neSpecc: So, if the user does not specify the **common** order through the root `inlineToolbar` property, the **default** order — is the order of tools from the `tools` property.

src/components/modules/toolbar/inline.ts
```diff
@@ -362,20 +362,63 @@ export default class InlineToolbar extends Module {
...
-    const currentSelection = SelectionUtils.get(),
-        currentBlock = this.Editor.BlockManager.getBlock(currentSelection.anchorNode as HTMLElement);
-
-    const toolSettings = this.Editor.Tools.getToolSettings(currentBlock.name),
-        inlineToolbarSettings = toolSettings && toolSettings[this.Editor.Tools.USER_SETTINGS.ENABLED_INLINE_TOOLS];
+    const currentSelection = SelectionUtils.get();
+    const currentBlock = this.Editor.BlockManager.getBlock(currentSelection.anchorNode as HTMLElement);
 
+    const toolSettings = this.Editor.Tools.getToolSettings(currentBlock.name);
+    const inlineToolbarSettings = toolSettings && toolSettings[this.Editor.Tools.USER_SETTINGS.ENABLED_INLINE_TOOLS];
```
neSpecc, line 421:
```suggestion
    const inlineToolbarSettingsForTool = toolSettings && toolSettings[this.Editor.Tools.USER_SETTINGS.ENABLED_INLINE_TOOLS];
```

src/components/modules/toolbar/inline.ts
```diff
@@ -393,10 +436,32 @@ export default class InlineToolbar extends Module {
 
     /**
      * Filter buttons if Block Tool pass config like inlineToolbar=['link']
+     * Else filter them according to the default inlineToolbar property.
      */
+
+    let inlineToolbar;
```
neSpecc, line 442:
```suggestion
    let inlineToolbarOrder;
```

types/configs/editor-config.d.ts
```diff
@@ -85,4 +85,9 @@ export interface EditorConfig {
    * @param {API} api - editor.js api
    */
   onChange?(api: API): void;
+
+  /**
+   * Defines default toolbar for all tools.
+   */
+  inlineToolbar?: string[];
```
neSpecc, line 92: it also can be `false`

neSpecc (changes requested):
- the tools are not ordered on the first selection

src/components/modules/toolbar/inline.ts
```diff
@@ -362,20 +378,76 @@ export default class InlineToolbar extends Module {
...
 
     const toolSettings = this.Editor.Tools.getToolSettings(currentBlock.name);
+    const inlineToolbarSettingsForTool = toolSettings && toolSettings[this.Editor.Tools.USER_SETTINGS.ENABLED_INLINE_TOOLS];
 
-    return toolSettings && toolSettings[this.Editor.Tools.USER_SETTINGS.ENABLED_INLINE_TOOLS];
+    /**
+     * Condition 1 : Returns true, if the inlineToolbar property of the block is an array or true;
+     * false otherwise.
+     */
+    const isInlineToolbarSettingsForToolProvided = inlineToolbarSettingsForTool == true || Array.isArray(inlineToolbarSettingsForTool);
```
gohabereg, line 387:
```suggestion
    const isInlineToolbarSettingsForToolProvided = inlineToolbarSettingsForTool === true || Array.isArray(inlineToolbarSettingsForTool);
```

src/components/modules/toolbar/inline.ts
```diff
@@ -362,20 +378,76 @@ export default class InlineToolbar extends Module {
...
+
+    /**
+     * Condition 2 : Returns true, if the inlineToolbar property of the block is not set,
+     * but the default inlineToolbar property is either an array or true.
+     */
+    const isDefaultInlineToolbarProvided = inlineToolbarSettingsForTool == undefined && this.config.inlineToolbar;
```
gohabereg, line 393:
```suggestion
    const isDefaultInlineToolbarProvided = inlineToolbarSettingsForTool === undefined && this.config.inlineToolbar;
```

gohabereg: Also var name does not actually express its value

src/components/modules/toolbar/inline.ts
```diff
@@ -362,20 +378,76 @@ export default class InlineToolbar extends Module {
...
+
+    /**
+     * If one of the above condition is true, then the inlineToolbar will be displayed.
+     */
+    const showInlineToolbar = isInlineToolbarSettingsForToolProvided || isDefaultInlineToolbarProvided;
+
+    /**
+     * condition 2 could return an array, so casting it to Boolean.
+     */
+    return Boolean(showInlineToolbar);
```
gohabereg, line 403: I think would be better to move all new lines here to separate method called smth like `getInlineToolbarConfig`

src/components/modules/toolbar/inline.ts
```diff
@@ -362,20 +378,76 @@ export default class InlineToolbar extends Module {
...
+
+    /**
+     * creates an array of all the button nodes
+     */
+    const buttons = Array.from(buttonNodes);
+
+    /**
+     * sorts the buttons array according to the inlineToolbar property provided by user
+     */
+    inlineToolbar.forEach((toolName, i) => {
```
gohabereg, line 425: Can't we use builtin `sort` method here?

src/components/modules/toolbar/inline.ts
```diff
@@ -393,10 +465,32 @@ export default class InlineToolbar extends Module {
...
      * Filter buttons if Block Tool pass config like inlineToolbar=['link']
+     * Else filter them according to the default inlineToolbar property.
      */
-    if (Array.isArray(inlineToolbarSettings)) {
+
+    let inlineToolbarOrder;
+
+    if (Array.isArray(inlineToolbarSettingsForTool)) {
+      inlineToolbarOrder = inlineToolbarSettingsForTool;
+    } else if (inlineToolbarSettingsForTool == undefined && Array.isArray(this.config.inlineToolbar)) {
```
gohabereg, line 475: This case should run if `inlineToolbarSettingsForTool === true` as well as if it is not defined

gohabereg, line 475:
```suggestion
    } else if (inlineToolbarSettingsForTool === undefined && Array.isArray(this.config.inlineToolbar)) {
```

gohabereg (changes requested):

src/components/modules/toolbar/inline.ts
ranemihir: We have to sort it according to the given array. Is there a built in sort method present for this?

src/components/modules/toolbar/inline.ts
ranemihir: `isDefaultInlineToolbarProvided` => Is an array provided for default inline toolbar property.
This is how i thought about it. Can u suggest a better name?

src/components/modules/toolbar/inline.ts
ranemihir: Sorry, I didn't understand properly what lines are you referring to.

src/components/modules/toolbar/inline.ts
neSpecc: @gohabereg answer pls

docs/CHANGELOG.md
```diff
@@ -14,8 +14,10 @@
 - `Improvements` - Unuseful log about missed i18n section has been removed  [#1269](https://github.com/codex-team/editor.js/issues/1269)
 - `Fix` - Fixed issue with enter key in inputs and textareas [#920](https://github.com/codex-team/editor.js/issues/920)
 - `Improvements` - Allowed to set `false` as `toolbox` config in order to hide Toolbox button [#1221](https://github.com/codex-team/editor.js/issues/1221)
+- `Improvements` - Allows users to provide a default `inlineToolbar` property which will be used for all tools whose `inlineToolbar` property is set to `true`. It can be overriden by the tools' own `inlineToolbar` property. Also, the inline tools will be ordered according to the order of the inline tools in array provided in the `inlineToolbar` property. [#1056](https://github.com/codex-team/editor.js/issues/1056)
```
khaydarov, line 17:
```suggestion
- `Improvements` - Allows users to provide default `inlineToolbar` property which will be used for all tools whose `inlineToolbar` property is set to `true`. It can be overriden by the tool's own `inlineToolbar` property. Also, inline tools will be ordered according to the order of the inline tools in array provided in the `inlineToolbar` property. [#1056](https://github.com/codex-team/editor.js/issues/1056)
```

src/components/modules/toolbar/inline.ts
```diff
@@ -362,20 +378,76 @@ export default class InlineToolbar extends Module {
...
+
+    /**
+     * If one of the above condition is true, then the inlineToolbar will be displayed.
+     */
+    const showInlineToolbar = isInlineToolbarSettingsForToolProvided || isDefaultInlineToolbarProvided;
+
+    /**
+     * condition 2 could return an array, so casting it to Boolean.
+     */
+    return Boolean(showInlineToolbar);
```
khaydarov, line 403:
```suggestion
    return showInlineToolbar;
```

khaydarov: but be sure that code above is safe

src/components/modules/toolbar/inline.ts
khaydarov: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

khaydarov (changes requested):

src/components/modules/toolbar/inline.ts
ranemihir: No its actually not, one of the property returns the inlineToolbar array itself, that's why I used Boolean(). WHat sould I do?

src/components/modules/toolbar/inline.ts
```diff
@@ -317,6 +333,47 @@ export default class InlineToolbar extends Module {
     return this.nodes.wrapper.contains(node);
   }
 
+  /**
+   * @param toolName
+   * @param tool
+   */
+  private inlineToolbarSettings(toolName): string[]|boolean {
```
gohabereg, line 340: Better name it `getInlineToolbarSettings`

gohabereg (changes requested):

ranemihir:
@gohabereg Done. Check latest commit.

gohabereg (approved):

neSpecc (changes requested):
If I won't pass the root `inlineToolbar` property, the Inline Toolbar won't be showed near any block.

This property should be optional.

src/components/modules/toolbar/inline.ts
```diff
@@ -230,21 +232,21 @@ export default class InlineToolbar extends Module<InlineToolbarNodes> {
...
 
-    /**
-     * Call 'clear' method for Inline Tools (for example, 'link' want to clear input)
-     */
-    this.tools.forEach((toolInstance: InlineTool) => {
-      if (typeof toolInstance.clear === 'function') {
-        toolInstance.clear();
-      }
-    });
+    // /**
```
neSpecc, line 242: moved to addToolsFiltered

neSpecc (approved):

khaydarov (approved):

gohabereg (approved):

PR #1326, opened 2020-09-24
Add generics to OutputBlockData

hata6502:
https://github.com/codex-team/editor.js/issues/1551


types/data-formats/output-data.d.ts
```diff
@@ -3,15 +3,15 @@ import {BlockToolData} from '../tools';
 /**
  * Output of one Tool
  */
-export interface OutputBlockData {
+export interface OutputBlockData<S extends string = string, T extends object = any> {
   /**
    * Too type
```
sosie-js, line 8: Tool type

hata6502: Thanks!

hata6502:
I create a repository to introduce the better way to use Editor.js with TypeScript. 
Please see: https://github.com/hata6502/editorjs-examples/blob/main/examples/EditorJSData.ts

types/data-formats/output-data.d.ts
```diff
@@ -3,15 +3,15 @@ import {BlockToolData} from '../tools';
...
  * Output of one Tool
  */
-export interface OutputBlockData {
+export interface OutputBlockData<S extends string = string, T extends object = any> {
   /**
-   * Too type
+   * Tool type
    */
-  type: string;
+  type: S;
```
neSpecc, line 10: The type is always a string. Are you sure we need a generic here?

neSpecc (changes requested):

types/data-formats/output-data.d.ts
```diff
@@ -3,15 +3,15 @@ import {BlockToolData} from '../tools';
 /**
  * Output of one Tool
  */
-export interface OutputBlockData {
+export interface OutputBlockData<S extends string = string, T extends object = any> {
```
neSpecc, line 6:
```suggestion
export interface OutputBlockData<S extends string = string, Data extends object = any> {
```

neSpecc: and describe it via `@template` please

neSpecc (changes requested):

types/data-formats/output-data.d.ts
hata6502: By using Generics, Editor.js users can describe string literal types. 

https://www.typescriptlang.org/docs/handbook/literal-types.html#string-literal-types

By using string literal types in `OutputBlockData['type']`, Editor.js users can describe more type-safe `OutputData` in each application. 

See also: 
https://github.com/hata6502/editorjs-examples/blob/main/examples/ExampleEditorJSData.ts#L17

neSpecc (approved):

gohabereg (approved):

Issue #873, opened 2019-08-25
Unique id for each block


floatas:
Hey, It would be great if editor assigned unique id for each block. This would help to keep references, add links to other blocks.
This probably would introduce breaking changes, discussion would be needed before implementing.

I have fork where I plan to do more changes, one of them is uids https://github.com/floatas/editor.js/commit/36c1f61ed66689bf9237ae3f37cc7e76e0658cb9

jamiemchale:
This is also mentioned in #727

christoph-kluge:
@floatas did you discovered any issues so far with your own implementation? 

I'm also looking into this right now because it makes it more easy to post-process removed/updated blocks on the server-side.

Can you elaborate more on the breaking changes?

vincent-herlemont:
@floatas I have the same issue and I have added [fasiha/mudderjs](https://github.com/fasiha/mudderjs) for generate unique ordered key. [#819/#506733820] also mentioned in (https://github.com/codex-team/editor.js/issues/819#issuecomment-506733820)

floatas:
@christoph-kluge  I haven't found any issues. When loading document without id's, they are generated. Seems to work with all cases, if you found any issues, let me know (also my modifications are available as npm). As for breaking changes, structure changed (added new property), I guess it should work fine with original documents, haven't tested that much.

@vincent-herlemont I'm thinking of adding ordered key to blocks, but for now I just take order of appearance and assign numbered key in backed to each block. Will definitely take a look at your implementation

christoph-kluge:
@floatas thank you sharing the the code. It helped me a lot. 

I discovered one issue: It always generates new uuids even if you pass an existing dataset through `data` while initilizing a new editor. I have extended your code with a small fix. You can find it here: https://github.com/christoph-kluge/editor.js/commit/be048d79116ce86e897c75428786c50d05b1b730

alibail:
Appreciate this is an old thread but it comes up when looking for this feature - I implemented this through introducing a "metadata" object on each block as I required to be able to assign a unique ID to every block to work with it in my backend system as well as pass additional metadata for each block through my backend system. Anyone looking for similar I have the sample code here, which can be easily customised to meet your own needs; https://github.com/alibail/editor.js

pie6k:
This is quite an important feature in my case.

My use case:

[...]

PR #1333, opened 2020-09-25
Add unique ids for each block

christoph-kluge:
I'm adressing #873 with this PR.

Tested Features:

* Existing data: If your previous block does not have an "id" property, it will create a new unique id
* Existing data: If your previous block does have an "id" property, it will keep the existing unique id
* Existing data: If a block gets changed the "id" property stays the same
* Insert New Block: It will create a new unique id
* Copy & Paste blocks: It will create a new unique ids per copied block
* Move Blocks (Up/Down): It will keep the existing unique id

Important Note:
* If you "transform/transition" an existing block to another block it will receive a NEW unique id


dsaralaya (approved):

sis-dk:
Does this handle creating a new id when copying a block and pasting it as a new block? Initialise empty editor, type a para block, CMD+A (selects all the blocks. In our case the para block), CMD+V. Would the newly pasted block have a new id?

christoph-kluge:
@sis-dk thanks for spotting the copy & paste issue. I have adressed this in my latest commit

mqtik:
Why don't use shortid for id generation?
It's what I'm about to use in a module I'm working on (native side), but this one uses shortid to generate ID's.

Just want to keep consistent.

christoph-kluge:
I have used this feature on a custom build since 2.15.2 with UUIDs. Now I wanted to update for latest >2.18+ version and simply ported this feature with UUIDs. I think we could also add some custom configuration/callback to generate the IDs in a different way. Since One editor will not have quadrillions of entries I could imageine that shortids are pretty valid too 👍 

Maybe the maintainers (@neSpecc, @khaydarov, @gohabereg) could give a statement about this. I'm fine to add some custom configuration/callback to support custom-ids instead of hardcoded uuids.

src/components/modules/renderer.ts
```diff
@@ -63,12 +65,14 @@ export default class Renderer extends Module {
    */
   public async insertBlock(item: OutputBlockData): Promise<void> {
     const { Tools, BlockManager } = this.Editor;
+    const id = item.id || _.generateUuidv4();
```
hata6502, line 68:
```suggestion
    const id = item.id;
```

When `const id` is undefined, a new ID is generated in `BlockManager.insert()`.

hata6502 (changes requested):

ChasLui:
I need this feature pls merge @christoph-kluge 🙏

neSpecc:
Resolved by #1667

PR #1336, opened 2020-09-30
BugFix 1302 undefined name

### LINKED ISSUE #1302: [Bug] Uncaught TypeError: Cannot read property 'name' of undefined
**Describe a bug:**
When user clicks the image and press tab and hit enter, user gets `Uncaught TypeError: Cannot read property 'name' of undefined` on the console.

**Steps to reproduce:**
1. Go to editor.js sample document. ([https://editorjs.io/](https://editorjs.io/))
2. Click sample image.
3. Hit Tab.
4. Hit Enter.
5. You get an error

**Expected behavior:**
You shouldn't get an error.

**Screenshots:**
ezgif com-optimize

robonetphy:
Resolve #1302 
Remaining:
- Cursor Removal form next Element in case of Delimiter Tool.
- Update Change Log


src/components/modules/ui.ts
```diff
@@ -392,8 +392,12 @@ export default class UI extends Module {
    * @param {KeyboardEvent} event - keyboard event
    */
   private defaultBehaviour(event: KeyboardEvent): void {
-    const keyDownOnEditor = (event.target as HTMLElement).closest(`.${this.CSS.editorWrapper}`);
     const { currentBlock } = this.Editor.BlockManager;
+    /**
+     *  currentBlock !=undefined
+     *  Ignore when KeyDown Element is null
+     */
+    const keyDownOnEditor = (event.target as HTMLElement).closest(`.${this.CSS.editorWrapper}`) || currentBlock != undefined;
```
gohabereg, line 400:
```suggestion
    const keyDownOnEditor = (event.target as HTMLElement).closest(`.${this.CSS.editorWrapper}`) || currentBlock !== undefined;
```

gohabereg (changes requested):

src/components/modules/ui.ts
neSpecc, line 400: let's move the `currentBlock !== undefined` statement to the separate value and describe this case (why it is counts as keydown on editor)

neSpecc (changes requested):

talyguryn (approved):

src/components/modules/ui.ts
robonetphy: Done

robonetphy: Done

src/components/modules/ui.ts
```diff
@@ -392,10 +392,21 @@ export default class UI extends Module {
...
+     *  currentBlock !=undefined
+     *  Ignore when KeyDown Element is null
+     */
+    const keyDownOnEditor = (event.target as HTMLElement).closest(`.${this.CSS.editorWrapper}`);
     const isMetaKey = event.altKey || event.ctrlKey || event.metaKey || event.shiftKey;
 
+    /**
+     * Transfer event when currentBlock defined and keyDownOnEditor is null
+     */
+    if(currentBlock!== undefined && keyDownOnEditor===null){
```
gohabereg, line 406:
```suggestion
    if(currentBlock !== undefined && keyDownOnEditor === null){
```

src/components/modules/ui.ts
```diff
@@ -392,10 +392,21 @@ export default class UI extends Module {
...
+      this.Editor.BlockEvents.keydown(event);
+      return;
```
gohabereg, line 408:
```suggestion

      return;
```

gohabereg (changes requested):

PR #1337, opened 2020-10-01
Throw error only if read-only is enabled from the start

gohabereg:


src/components/modules/readonly.ts
```diff
@@ -13,6 +13,11 @@ import { CriticalError } from '../errors/critical';
  * @property {boolean} readOnlyEnabled - read-only state
  */
 export default class ReadOnly extends Module {
+  /**
+   * Array of tools name which don't support read-only mode
+   */
+  private toolsDontSupportReadOnly: string[] = [];
```
neSpecc, line 19:
```suggestion
  private toolsDoesntSupportReadOnly: string[] = [];
```

neSpecc (approved):

khaydarov (approved):

PR #1356, opened 2020-10-09
Fixes for 2.19

gohabereg:


example/example.html
```diff
@@ -26,6 +26,21 @@
       <div class="ce-example__button" id="saveButton">
         editor.save()
       </div>
+
+      <div class="ce-example__statusbar">
+
+        @neSpecc neSpecc yesterday Author Member
```
neSpecc, line 32: ༼ʘ̚ل͜ʘ̚༽

types/api/blocks.d.ts
```diff
@@ -19,7 +19,7 @@ export interface Blocks {
    *
```
neSpecc, line 19: remove param from jsdoc

neSpecc (changes requested):

example/example.html
```diff
@@ -26,6 +26,21 @@
...
+        i think, all the changes from example-dev.html should be included to the example.html as well.
+
+        @gohabereg	Reply…
```
khaydarov, line 35: what is that

neSpecc (approved):

khaydarov (approved):

PR #1366, opened 2020-10-13
Fixes for release:

neSpecc:
- Fix paddings in InlineToolbar
- Fix double-firing of shortcuts
- Fix double firing of InlineToolbar `make` method on the editor initialization
- Found an old bug when internal tools `shortcut` property was ignored


src/components/modules/toolbar/toolbox.ts
```diff
@@ -232,32 +233,48 @@ export default class Toolbox extends Module<ToolboxNodes> {
...
 
     /** Increment Tools count */
     this.displayedToolsCount++;
   }
 
+  /**
+   * Returns tool's shortcut
+   * It can be specified via internal 'shortcut' static getter or by user settings for tool
+   *
+   * @param {string} toolName - name of tool
```
khaydarov, line 250:
```suggestion
   * @param {string} toolName - tool's name
```

src/components/modules/toolbar/toolbox.ts
```diff
@@ -232,32 +233,48 @@ export default class Toolbox extends Module<ToolboxNodes> {
...
+   * @param {ToolConstructable} tool - name of tool
```
khaydarov, line 251:
```suggestion
   * @param {ToolConstructable} tool - tool's instance
```

gohabereg (approved):

khaydarov (approved):

Issue #1382, opened 2020-10-15
[Bug] Wrong caret moving and block merging when there are spaces at block edge


hata6502:
Wrong caret moving and block merging when there are spaces at block edge. 

Steps to reproduce:
Please see GIF animations. 

Expected behavior:
Don't ignore spaces at block edge when caret moving and block merging. 

Screenshots:

- Wrong caret moving in right direction
![ezgif com-gif-maker (4)](https://user-images.githubusercontent.com/7702653/96152224-46367200-0f47-11eb-889d-58f6e8c5c259.gif)

- Wrong caret moving in left direction
![ezgif com-gif-maker (5)](https://user-images.githubusercontent.com/7702653/96152787-e2607900-0f47-11eb-9ae3-281d86226711.gif)

- Wrong block merging
![ezgif com-gif-maker (6)](https://user-images.githubusercontent.com/7702653/96153099-3f5c2f00-0f48-11eb-8df5-a5182e0163ad.gif)

Device, Browser, OS: PC, Chrome, Ubuntu 20.04

Editor.js version: 2.19.0

Plugins you use with their versions: none

PR #1385, opened 2020-10-16
Fixed block merging and caret moving when there are spaces at the block edge

hata6502:
https://github.com/codex-team/editor.js/issues/1382


hata6502:
TODO: fix Block.isEmpty

src/components/block/index.ts
```diff
@@ -367,8 +367,8 @@ export default class Block {
    *
    * @returns {boolean}
    */
-  public get isEmpty(): boolean {
-    const emptyText = $.isEmpty(this.pluginsContent);
+  public isEmpty(ignoreSpaces: boolean): boolean {
```
hata6502, line 370: I think getter/setter should not be used for the future maintainability. 
See: https://github.com/airbnb/javascript#accessors--no-getters-setters

src/components/modules/caret.ts
```diff
@@ -205,13 +209,13 @@ export default class Caret extends Module {
      * Why not regular .trim():
      *  in case of ' hello |' trim() will also remove space at the beginning, so length will be lower than anchorOffset
      */
-    const rightTrimmedText = lastNode.textContent.replace(/\s+$/, '');
+    const text = ignoreSpaces ? lastNode.textContent.replace(/\s+$/, '') : lastNode.textContent;
```
christoph-kluge, line 212: I have discovered a bug on 2.22.2 and I'm looking for hints how to solve it properly and understand the code a bit better. During my research I found this PR and I just want to mention:

In certain scenarios there is a chance that `lastNode` might be `undefined`. In L190 you're checking it correctly but here `if (lastnode && ...` you're not. 

In my scenario I have custom block which which uses onPaste configurations. The Block itself uses `ReactDOM.render()` to offload it into a react component - just in case it's important. 

If my caret is in the middle of the text and I paste my "pattern" then after moving left,up,down,right then I receive several errors related to `isAtEnd()` and `$.isEmpty()` which uses a `node.normalize()` while `node` is `undefined`. Dunno yet why. Trying to identify it on my side.

christoph-kluge: Just an informational comment: After debugging further I realised: After I paste my "onPaste" inside a paragraph and the block is successfully added, then my caret is still at the paragraph-block. But when I start moving up,down,left,right then the result of `this.Editor.BlockManager.currentBlock` returns the new inserted block instead of the position of the caret.

Need to debug further.

asfixia: Why not just stop doing trim or similar to discover if the caret is at the end or start?
These changes looks like an wrong work around.

Issue #1393, opened 2020-10-22
[Bug] When entering Editor.js by Tab key, keydown event listener occurs many errors


hata6502:
When entering Editor.js by Tab key, keydown event listener occurs many errors!
Please see the following video. 

Screenshots:
![ezgif com-gif-maker](https://user-images.githubusercontent.com/7702653/96825244-a196c180-146b-11eb-9d14-4528447f1eaf.gif)

Editor.js version: v2.19.0

PR #1397, opened 2020-10-24
fix(interaction): problem with entering to Editor.js by Tab key

hata6502:
https://github.com/codex-team/editor.js/issues/1393


robonetphy:
@hata6502 Can you test this bug on current version of the Editor.js.
I guess this Bug also resolve in last release.

hata6502:
@robonetphy Hello! I've tried in `next` branch, but this bug occurs in example/example.html
And it also can be reproduced in v2.19.0 https://editorj
ezgif com-gif-maker (1)
s.io/

robonetphy (approved):

src/components/modules/ui.ts
```diff
@@ -750,6 +750,8 @@ export default class UI extends Module<UINodes> {
       return;
     }
 
+    this.Editor.BlockManager.setCurrentBlockByChildNode(focusedElement);
```
neSpecc, line 753: This method will be called many times when user will do selecting some text fragment. Maybe we can provide some optimizations?

neSpecc (approved):

Issue #1398, opened 2020-10-24
[Bug] Remove Shortcut listners and codex-tooltip when destroy()


hata6502:
Since forget to release Shortcut listners and codex-tooltip when destroy(), 
they may occur performance issue with long time using in same browser window. 

Expected behavior:
Release Shortcut listners and codex-tooltip by destroy()

Editor.js version: v2.19.0

UmangRobro:
@hata6502 So Can we close This issue?

robonetphy:
@hata6502  So Can we close this issue?

PR #1399, opened 2020-10-24
Remove Shortcut listeners when destroy()

hata6502:
https://github.com/codex-team/editor.js/issues/1398


src/components/modules/shortcuts.ts
```diff
@@ -71,4 +71,14 @@ export default class Shortcuts extends Module {
     this.registeredShortcuts[index].remove();
     this.registeredShortcuts.splice(index, 1);
   }
+
+  /**
+   * Module destruction
+   * De-registers all Shortcuts
+   */
+  public destroy(): void {
+    while (this.registeredShortcuts.length > 0) {
```
sis-dk, line 80: Wouldn't this while loop run infinitely if for some reason the remove function returns from line 68? There is no reason for that to happen but just in case. A for of loop would be safer?

hata6502: Sorry for my delay. 

I think when line 68 removed occurs more other bugs. Even use for loop instead of while, it is not a resolution of true problem.

robonetphy:
@hata6502 Please remove the conflict.
So we can include this PR in our next patch.

hata6502:
Resolved conflicts.

robonetphy:
@hata6502 
Are we Talking about these (bottom right side contain shortcut.js key up listener ) shortcut listener ? 
image
Because after running `this.editor.destroy()` on `next` .it will be removed .
image

hata6502:
@robonetphy 
Thank you!
The shortcut listeners can be really removed in `next` branch!
This bug occurs only v2.18.0, and it will be resolved in next version.

PR #1404, opened 2020-10-28
Resolve editor.destroy() Problem

### LINKED ISSUE #1380: [Bug] Problem with editor.destroy() in 2.19
```js
const editor = new EditorJS()

editor.destroy(); // TypeError: null is not an object (evaluating 'this.flipper.deactivate')
```

Steps to reproduce:
1. Go to example/example.html
2. Call `editor.destroy()` from the console
3. You'll see `TypeError: null is not an object (evaluating 'this.flipper.deactivate')`

Expected behavior:

The editor destroys correctly.

Device, Browser, OS:

Safari

Editor.js version:

2.19

robonetphy:
Resolve #1380 
@neSpecc let me know Changes I have to make in CHANGELOG.md (Create new version or new patch) ?
and Also, `yarn lint` giving me as the following warning, 
Screenshot 2020-10-29 005601
How to solve that?


src/components/modules/toolbar/index.ts
```diff
@@ -381,8 +381,6 @@ export default class Toolbar extends Module<ToolbarNodes> {
    * It is used in Read-Only mode
    */
   private destroy(): void {
-    this.Editor.Toolbox.destroy();
```
sis-dk, line 384: This destroy is also called when we [toggle read only mode](https://github.com/codex-team/editor.js/blob/75d6592cd7510e68cf99634992412ced764de260/src/components/modules/toolbar/index.ts#L152). Would that expect this destroy to take care of destroying toolbox and blockSettings? Just curious as we use this in production.

https://github.com/codex-team/editor.js/blob/75d6592cd7510e68cf99634992412ced764de260/src/components/modules/toolbar/index.ts#L152

robonetphy: @sis-dk 
Ya that would be handle with toggle button. We just removing this because it called twice when we trying to destroy the editor.

neSpecc: @sis-dk @robonetphy thanks. I've moved the Toolbox and BlockSettinga destroy methods to the `toggleReadOnly`, because it is the logical opposite to `this.drawUI()`.

1. On the read-only mode toggling all the modules destroys correctly, no memory leak here. 
2. On `editor.destroy()` all modules` destroy fired once.

neSpecc (approved):

gohabereg (approved):

Issue #1458, opened 2020-12-03
[Bug] `removed` lifecycle hook doesn't dispatch when `EditorJS.destroy()`


hata6502:
`removed` lifecycle hook doesn't dispatch when `EditorJS.destroy()`. 
https://editorjs.io/tools-api#lifecycle-hooks

Steps to reproduce:
1. Embed test `removed` hook in paragraph block

```
  removed() {
    console.log('removed');
  }
```

2. When delete paragraph block, the `removed` hook dispatches. 

![Screenshot from 2020-12-04 08-45-37](https://user-images.githubusercontent.com/7702653/101103849-2b11e580-360d-11eb-8208-47810796b3dc.png)

3. When `EditorJS.destroy()` is executed, the `removed` hook doesn't dispatch. 

`example-dev.html`

```
    setTimeout(() => {
      editor.destroy();
    }, 3000);
```

![Screenshot from 2020-12-04 08-48-31](https://user-images.githubusercontent.com/7702653/101104004-87750500-360d-11eb-8ba6-6fc30c153174.png)

Expected behavior:
`removed` lifecycle hook dispatches with also `EditorJS.destroy()`. 

Device, Browser, OS: Ubuntu 20.04, Chrome

Editor.js version: 2.19.0

hata6502:
Alternative solution: use `EditorJS.clear()` before `EditorJS.destory()`

PR #1459, opened 2020-12-04
removed lifecycle hooks are dispatched with also Editor JS.destroy()

hata6502:
https://github.com/codex-team/editor.js/issues/1458


hata6502:
Sorry! I've found many dangerous BREAKING CHANGES in this PR.

PR #1480, opened 2020-12-16
Add data property into InlineToolConstructorOptions

Danue1:
```typescript
// before
import { BaseTool, BlockTool, BlockToolConstructorOptions, InlineToolConstructorOptions } from "@editorjs/editorjs";

export class Example implements BaseTool, BlockTool {
  ...
  constructor({ data }: BlockToolConstructorOptions | (InlineToolConstructorOptions & { data?: undefined })) {
    console.log(data);
  }
  ...
}

// after
import { BaseTool, BlockTool, BlockToolConstructorOptions, InlineToolConstructorOptions } from "@editorjs/editorjs";

export class Example implements BaseTool, BlockTool {
  ...
  constructor({ data }: BlockToolConstructorOptions | InlineToolConstructorOptions) {
    console.log(data);
  }
  ...
}
```

I am using EditorJS based on TypeScript, but I want to add it because there is an inconvenient part of the type definition.

### DIFF
--- a/types/tools/inline-tool.d.ts
+++ b/types/tools/inline-tool.d.ts
@@ -43,6 +43,7 @@ export interface InlineTool extends BaseTool {
  */
 export interface InlineToolConstructorOptions {
   api: API;
+  data?: undefined;
   config?: ToolConfig;
 }
 


hata6502:
I think, `BaseTool` and `InlineToolConstructorOptions` is not needed to implement Block Tool. 
Because `BlockTool` extends `BaseTool`, and `InlineToolConstructorOptions` is not for Block Tool. 

Can you use the following type definition?

```
export class Example implements BlockTool {
  constructor({ data }: BlockToolConstructorOptions ) {
    console.log(data);
  }
}
```

Danue1:
image
image

I thought I had to make an inline tool and a block tool separately,
This error occurred and I thought it was intended.
I am solving this type error with this PR, is there a better way?

hata6502:
Awesome! I've experienced the same issue. 
I resolved this issue by another way in my PR. 
https://github.com/codex-team/editor.js/pull/1484

hata6502:
@neSpecc 
I wants to merge this PR, too.

types/tools/inline-tool.d.ts
```diff
@@ -43,6 +43,7 @@ export interface InlineTool extends BaseTool {
  */
 export interface InlineToolConstructorOptions {
   api: API;
+  data?: undefined;
```
neSpecc, line 46: Inline Tools never accept `data` at the constructor. So I think it is not a solution.

Danue1: I also think it is good solution, but it works on current design without several changes.
I can't imagine better solution on current design.

hata6502: Me, too. 
This is the easiest way to fix the type definition.

Danue1: Or... We have plan to API changes for fixing about this case?

gohabereg:
Hey there,
I've made a small research and here is results:
1. Editor's config accepts `ToolConstructable`:
```
new EditorJS({
  tools: {
    myTool: MyTool // typeof MyTool should implement ToolConstructable
  }
})
``` 
2. By signature `export type ToolConstructable = BlockToolConstructable | InlineToolConstructable`
3. So the error appears because there is no way to connect `BlockTool` and `BlockToolConstructable` interfaces:
```
class MyTool implements BlockTool {
  constructor(data: BlockToolConstructorOptions) {}
}
```
Code above states that `MyTool` implements `BlockTools`, but doesn't state `typeof MyTool` implements `BlockToolConstructable`
4. So when we pass MyTool to EditorJS constructor compiler doesn't know if `typeof MyTool` is `BlockToolConstructable` or `InlineToolConstructable`, that is the reason of the error.


Considering above the only __correct__ way to avoid the error is to explicitly specify what is `typeof MyTool`:

```
new EditorJS({
  tools: {
    myTool: myTool as BlockToolConstructable,
    myInlineTool: myInlineTool as InlineToolConstructable
  }
})
```

gohabereg:
The other way is to describe `BlockTool` and `InlineTool` as classes and set these classes as a type for EditorJS constructor, but for now we don't have these classes exported from the package

Issue #1432, opened 2020-11-21
Fix ToolConfig type definition


hata6502:
```
export type ToolConfig<T extends object = any> = T;
```

suggestion: 
```
export type ToolConfig<T extends object = any> = T | {};
```

Each tools should expect empty object as config. 
When developing the tools, it is difficult to expect an error by empty object. 

https://github.com/codex-team/editor.js/blob/next/src/components/modules/tools.ts#L345
https://github.com/codex-team/editor.js/blob/next/src/components/block/index.ts#L220

PR #1484, opened 2020-12-21
Fix type definition for tool constructor

hata6502:
https://github.com/codex-team/editor.js/issues/1432

### DIFF
--- a/src/types-internal/editor-modules.d.ts
+++ b/src/types-internal/editor-modules.d.ts
@@ -27,7 +27,7 @@ import NotifierAPI from '../components/modules/api/notifier';
 import SaverAPI from '../components/modules/api/saver';
 import Saver from '../components/modules/saver';
 import BlockSelection from '../components/modules/blockSelection';
-import RectangleSelection from '../components/modules/RectangleSelection';
+import RectangleSelection from '../components/modules/rectangleSelection';
 import InlineToolbarAPI from '../components/modules/api/inlineToolbar';
 import CrossBlockSelection from '../components/modules/crossBlockSelection';
 import ConversionToolbar from '../components/modules/toolbar/conversion';
--- a/types/tools/block-tool.d.ts
+++ b/types/tools/block-tool.d.ts
@@ -19,9 +19,9 @@ export interface BlockTool extends BaseTool {
   /**
    * Process Tool's element in DOM and return raw data
    * @param {HTMLElement} block - element created by {@link BlockTool#render} function
-   * @return {BlockToolData}
+   * @return {BlockToolData | Promise<BlockToolData>}
    */
-  save(block: HTMLElement): BlockToolData;
+  save(block: HTMLElement): BlockToolData | Promise<BlockToolData>;
 
   /**
    * Create Block's settings block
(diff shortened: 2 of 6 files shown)


types/tools/block-tool.d.ts
```diff
@@ -85,7 +85,7 @@ export interface BlockTool extends BaseTool {
  */
 export interface BlockToolConstructorOptions<D extends object = any, C extends object = any> {
   api: API;
-  data: BlockToolData<D>;
+  data?: BlockToolData<D>;
```
hata6502: By marking to undefinable, resolved https://github.com/codex-team/editor.js/pull/1480.

types/tools/block-tool.d.ts
```diff
@@ -19,9 +19,9 @@ export interface BlockTool extends BaseTool {
   /**
    * Process Tool's element in DOM and return raw data
    * @param {HTMLElement} block - element created by {@link BlockTool#render} function
-   * @return {BlockToolData}
+   * @return {BlockToolData | Promise<BlockToolData>}
    */
-  save(block: HTMLElement): BlockToolData;
+  save(block: HTMLElement): BlockToolData | Promise<BlockToolData>;
```
hata6502, line 24: BlockTool.save() supports async/await, so support Promise also in type definition.

types/tools/block-tool.d.ts
Danue1: It seems better to modify InlineToolConstructorOptions.
If you change BlockToolConstructorOptions.data to be optional, the user of this library asks "It might not pass data?" And can get confused.

Danue1: Like the PR I requested, InlineToolConstructorOptions.data can be added as an optional undefined type. "Why did this library have a type specified here?" As users would think, there is no need to significantly change the API design of this library, and as a result, I think there will be nothing wrong.

hata6502: 👍
I think the solution in https://github.com/codex-team/editor.js/pull/1480 is better! 
In my solution, each BlockTool needs to check `undefined`. 
https://github.com/hata6502/editorjs-examples/blob/main/examples/ExampleBlockTool.ts#L43

@neSpecc 
I changed this PR draft, and I want to merge https://github.com/codex-team/editor.js/pull/1480 first.

Danue1: Thank you for understanding!

hata6502:
Conflict resolved.

src/types-internal/editor-modules.d.ts
```diff
@@ -27,7 +27,7 @@ import NotifierAPI from '../components/modules/api/notifier';
 import SaverAPI from '../components/modules/api/saver';
 import Saver from '../components/modules/saver';
 import BlockSelection from '../components/modules/blockSelection';
-import RectangleSelection from '../components/modules/RectangleSelection';
+import RectangleSelection from '../components/modules/rectangleSelection';
```
hata6502, line 30:
```
$ yarn tsc --noEmit
yarn run v1.22.10
$ /home/hata6502/editor.js/node_modules/.bin/tsc --noEmit
src/types-internal/editor-modules.d.ts:30:32 - error TS2307: Cannot find module '../components/modules/RectangleSelection'.

30 import RectangleSelection from '../components/modules/RectangleSelection';
                                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error.

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

```

PR #1456, opened 2020-12-03
Fix SanitizerConfig type definition

hata6502:
https://github.com/codex-team/editor.js/issues/1513

PR #1491, opened 2020-12-30
Hotfix: sanitizer config type definition

phanletrunghieu:
When I use sanitizer with typescript.

I wanna use a function with `el` param.

```js
const sanitizerConfig = {
  b: true,
  p: function(el) {
    return !el.textContent.includes('bad text')
  }
}
```

(https://editorjs.io/sanitizer)

But sanitizer is definited with `func(): any` (missing `el` parram).


neSpecc (changes requested):
No description provided.

types/configs/sanitizer-config.d.ts
```diff
@@ -31,5 +31,5 @@ export interface SanitizerConfig {
    *   }
    * }
    */
-  [key: string]: boolean|{[attr: string]: boolean|string}|(() => any);
+  [key: string]: boolean|{[attr: string]: boolean|string}|((el?: any) => any);
```
neSpecc, line 34: why do you use `any` type instead of `Element`?

neSpecc (changes requested):

hata6502:
https://github.com/codex-team/editor.js/pull/1456

khaydarov:
@hata6502 could you resolve Peters' comment please?

types/configs/sanitizer-config.d.ts
phanletrunghieu: Reasonable

neSpecc (approved):

khaydarov (approved):

PR #1493, opened 2020-12-30
feat: Add unique ids for each block

cobbcheng:
When editing documents collaboratively, we usually need a unique id to save and identify the current block. This PR is modified and resubmitted based on this #1333 , because this PR is no longer maintained

I'm adressing #873 with this PR.

Tested Features:

- Existing data: If your previous block does not have an "id" property, it will create a new unique id
- Existing data: If your previous block does have an "id" property, it will keep the existing unique id
- Existing data: If a block gets changed the "id" property stays the same
- Insert New Block: It will create a new unique id
- Copy & Paste blocks: It will create a new unique ids per copied block
- Move Blocks (Up/Down): It will keep the existing unique id
- 

Important Note:

If you "transform/transition" an existing block to another block it will receive a NEW unique id


neSpecc:
Please, add a description to the PR. What case you're trying to cover?

cobbcheng:
> Please, add a description to the PR. What case you're trying to cover?

I'm sorry this is my negligence, I have added a description

mahono:
Awesome. This is exactly what we need to finally use EditorJs for our system. We also need an ID per block for collaboration features and in some special cases to map (save) each block to a separate database record (which is not possible without an identifier). UUIDv4 perfectly fits. Maybe someone needs another "id generator" but this could be added/opened in a later step. So I can't wait to see this feature in EditorJS. 👍

chumbalayaa:
Yes, we'd love to have uids assigned for each block. Really important for our commenting use case!

src/components/block/api.ts
```diff
@@ -14,6 +14,14 @@ function BlockAPI(
   block: Block
 ): void {
   const blockAPI: BlockAPIInterface = {
+    /**
+     * Tool id
```
khaydarov, line 18:
```suggestion
     * Block id
```

src/components/block/index.ts
```diff
@@ -27,6 +27,11 @@ import SelectionUtils from '../selection';
  * Interface describes Block class constructor argument
  */
 interface BlockConstructorOptions {
+  /**
+   * Tool's id
```
khaydarov, line 31:
```suggestion
   * Block's id
```

src/components/block/index.ts
```diff
@@ -107,6 +112,11 @@ export default class Block {
     };
   }
 
+  /**
+   * unique identifier
```
khaydarov, line 116:
```suggestion
   * Block unique identifier
```

types/api/block.d.ts
```diff
@@ -5,6 +5,11 @@ import {SavedData} from '../data-formats';
  * @interface BlockAPI Describes Block API methods and properties
  */
 export interface BlockAPI {
+  /**
+   * Tool name
```
khaydarov, line 9:
```suggestion
   * Block unique identifier
```

src/components/utils.ts
```diff
@@ -607,6 +607,21 @@ export function getValidUrl(url: string): string {
   }
 }
 
+/**
+ * Creates a UUID v4 version
+ *
+ * @returns {string}
+ */
+export function generateUuidv4(): string {
```
khaydarov, line 615: I think it would be better to name this method just as `generateId()` and make `UUIDv4` "private". E.g.
```js
export function generateId(): string {
    return uuidv4();
}

function uuidv4(): string {
  // your code
}
```
so that we could change the id generator algorithm

khaydarov (changes requested):

src/components/modules/paste.ts
```diff
@@ -759,7 +759,7 @@ export default class Paste extends Module {
    *
    * @returns {void}
    */
-  private insertEditorJSData(blocks: Array<Pick<SavedData, 'data' | 'tool'>>): void {
+  private insertEditorJSData(blocks: Array<Pick<SavedData, 'id' | 'data' | 'tool'>>): void {
```
hata6502, line 762: `id` seems unused. 

```suggestion
  private insertEditorJSData(blocks: Array<Pick<SavedData, 'data' | 'tool'>>): void {
```

neSpecc:
I think it's better to use `nanoid(10)` instead of `UUIDv4`

https://github.com/ai/nanoid

hata6502:
Should we close the ancestor PR https://github.com/codex-team/editor.js/pull/1333 with respects?

hata6502:
> I think it's better to use `nanoid(10)` instead of `UUIDv4`
> 
> https://github.com/ai/nanoid

`nanoid(10)` is not recommended because of the collision rate. 
The default behavior `nanoid()` is as safe as uuid. 

https://zelark.github.io/nano-id-cc/

neSpecc:
> > I think it's better to use `nanoid(10)` instead of `UUIDv4`
> > https://github.com/ai/nanoid
> 
> `nanoid(10)` is not recommended because of the collision rate.
> The default behavior `nanoid()` is as safe as uuid.
> 
> https://zelark.github.io/nano-id-cc/

The uniqueness of block ids will be scoped by the single entry. So I think that collision of ids of blocks inside one entry will be rare.

cobbcheng:
Thanks for your suggestions, you are right, I will modify and submit again, thanks!

src/components/utils.ts
```diff
@@ -607,6 +607,30 @@ export function getValidUrl(url: string): string {
...
+  });
+}
+
+/**
+ * Create a block id
+ *
+ * @returns {string}
+ */
+export function generateBlockId(): string {
+  return uuidv4();
```
neSpecc, line 631: Check this comment please https://github.com/codex-team/editor.js/pull/1493#issuecomment-773294522

neSpecc (changes requested):

src/components/utils.ts
cobbcheng: Ok i will change to nano(10)

evanjmg: @cobbcheng need any help on this?

cypress/plugins/index.js
```diff
@@ -0,0 +1,21 @@
...
+//
+// You can read more here:
+// https://on.cypress.io/plugins-guide
+// ***********************************************************
+
+// This function is called when a project is opened or re-opened (e.g. due to
+// the project's config changing)
+
+/**
+ * @type {Cypress.PluginConfig}
```
evanjmg, line 16: This supposed to be here?

marcoancona:
Why is this considered "not ready"? Something I can help with? We would also like to use Editor.js but we are stuck because we cannot merge edits from different users.

mqtik:
I'd like to help too.

mqtik:
Question:
Does this respect blocks already containing an ID?

gohabereg:
Closing due more relevant #1667

Issue #1516, opened 2021-01-27
Add a generic type to ToolSettings


hata6502:
By adding a generic type to ToolSettings, `EdtorConfig["tools"]` can be more TypeScript-firendly. 

See an example: [examples/ExampleEditorConfigTools.ts](https://github.com/hata6502/editorjs-examples/blob/main/examples/ExampleEditorConfigTools.ts): A type definition improved EditorConfig\["tools"\].

`ExampleEditorConfigTools` is one of `EdtorConfig["tools"]` mapped between tool name and tool config.

PR #1517, opened 2021-01-27
Add a generic type to ToolSettings

hata6502:
https://github.com/codex-team/editor.js/issues/1516


khaydarov (approved):

types/tools/tool-settings.d.ts
```diff
@@ -19,7 +19,7 @@ export interface ToolboxConfig {
 /**
  * Object passed to the Tool's constructor by {@link EditorConfig#tools}
  */
-export interface ToolSettings {
+export interface ToolSettings <T extends object = any> {
```
neSpecc, line 22: `T` should be named as `Config`. And this type variable should be documented via `@template` tag ([example](https://github.com/codex-team/ctproto/blob/main/types/message.ts))

gohabereg (approved):

neSpecc (approved):

Issue #1520, opened 2021-01-28
💡 Add DefaultBlockData type to develop customized paragraph blocks


hata6502:
1. Describe a problem.
When splitting a paragraph block, Editor.js makes block data based on [editor-js/paragraph](https://github.com/editor-js/paragraph). 

https://github.com/codex-team/editor.js/blob/next/src/components/modules/blockManager.ts#L488

This behavior may occur errors with customized default block. 

2. Describe the solution you'd like. Mockups are welcome.
Add `DefaultBlockData`  type in Editor.js. 

```
interface DefaultBlockData {
  text: string;
}
```

See also: https://github.com/hata6502/editorjs-examples/blob/main/examples/ExampleDefaultBlockTool.ts

neSpecc:
Yeah, there is a `todo` for that

hata6502:
test

PR #1521, opened 2021-01-28
chore(types): add DefaultBlockToolData type for insert/split of the default tool

hata6502:
https://github.com/codex-team/editor.js/issues/1520


neSpecc:
What if we will use the default tool with other data structure?

hata6502:
> What if we will use the default tool with different data structure?

Then the DefaultBlockToolData should be converted in customized default block constructor.

src/components/core.ts
```diff
@@ -167,15 +167,19 @@ export default class Core {
      */
     this.config.minHeight = this.config.minHeight !== undefined ? this.config.minHeight : 300;
 
+    const defaultBlockToolData: DefaultBlockToolData = {
+      text: '',
```
neSpecc, line 177: I think it is not a solution because the `text` is still hardcoded

neSpecc (changes requested):

src/components/core.ts
hata6502: Even `text` data is hard-coded, it can be controled in the constructor of customized default block tool. 
See: https://github.com/hata6502/editorjs-examples/blob/main/examples/ExampleDefaultBlockTool.ts#L28

gohabereg: I'd say that's OK as temporary solution

gohabereg (approved):

hata6502:
Resolved conflicts.

neSpecc:
I still think that this is not a solution. We need to support any default tool strucuture.

gohabereg:
This is not a solution for the original issue, but it is still good improvement for current version

hata6502:
How would you like it?

hata6502:
Resolved conflicts.

neSpecc (approved):

PR #1524, opened 2021-01-29
Adding `configuration` field-type to EditorJS class

alirezakay:
`configuration` object is a part of EditorJS instance, therefore, we must have its type definition in this file.
So that we won't get any typescript errors on accessing this object on other projects.

i.e. 
> Property 'configuration' does not exist on type 'EditorJS'. ts(2339)


khaydarov:
@alirezakay Thank you for PR, but can you give an example of this property usage? You have configuration object before Editor instantiation so you can use that

PR #1541, opened 2021-02-17
Feature: remove bundles

talyguryn:
Remove bundles for easy merging.


gohabereg (approved):

neSpecc (changes requested):
Can not be merged without budling github actions

example/example-rtl.html
```diff
@@ -27,6 +27,9 @@
   </div>
   <div class="ce-example__content _ce-example__content--small">
     <div id="editorjs"></div>
+    <div id="hint-core" style="text-align: center;">
+      No core bundle file found. Run <code class="inline-code">yarn build</code>
+    </div>
     <div id="hint" style="text-align: center;">
```
neSpecc, line 33:
```suggestion
    <div id="hint-tools" style="text-align: center;">
```

example/example-i18n.html
```diff
@@ -58,7 +58,7 @@
 <script src="https://cdn.jsdelivr.net/npm/@editorjs/inline-code@latest"></script><!-- Inline Code -->
 
 <!-- Load Editor.js's Core -->
-<script src="../dist/editor.js"></script>
+<script src="../dist/editor.js" onload="document.getElementById('hint-core').hidden = true"></script>
```
neSpecc, line 61: where is the `hint-core` on this page?

neSpecc (changes requested):

example/example-rtl.html
talyguryn: thanks

example/example-i18n.html
talyguryn: sorry. missed

neSpecc (approved):

neSpecc (approved):

Issue #1552, opened 2021-02-20
[Bug] Propagated config data will be mutated by Editor.js initialization


hata6502:
Describe a bug.
Propagated config data will be mutated by Editor.js initialization. 

Steps to reproduce:
1. Use the following `example/example-dev.html`

```
<!--
 Use this page for debugging purposes.

 Editor Tools are loaded as git-submodules.
 You can pull modules by running `yarn pull_tools` and start experimenting.
 -->
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <title>Editor.js 🤩🧦🤨 example</title>
  <link href="https://fonts.googleapis.com/css?family=PT+Mono" rel="stylesheet">
  <link href="assets/demo.css" rel="stylesheet">
  <script src="assets/json-preview.js"></script>
  <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no" />
</head>
<body>
  <div class="ce-example">
    <div class="ce-example__header">
      <a class="ce-example__header-logo" href="https://codex.so/editor">Editor.js 🤩🧦🤨</a>

      <div class="ce-example__header-menu">
        <a href="https://github.com/editor-js" target="_blank">Plugins</a>
        <a href="https://editorjs.io/usage" target="_blank">Usage</a>
        <a href="https://editorjs.io/configuration" target="_blank">Configuration</a>
        <a href="https://editorjs.io/creating-a-block-tool" target="_blank">API</a>
      </div>
    </div>
    <div class="ce-example__content _ce-example__content--small">
      <div id="editorjs"></div>
      <div id="hint-core" style="text-align: center;">
        No core bundle file found. Run <code class="inline-code">yarn build</code>
      </div>
      <div id="hint-tools" style="text-align: center;">
        No submodules found. Run <code class="inline-code">yarn pull_tools</code>
      </div>
      <div class="ce-example__button" id="saveButton">
        editor.save()
      </div>
      <div class="ce-example__statusbar">
        Readonly:
        <b id="readonly-state">
          Off
        </b>
        <div class="ce-example__statusbar-button" id="toggleReadOnlyButton">
          toggle
        </div>
      </div>
    </div>
    <div class="ce-example__output">
      <pre class="ce-example__output-content" id="output"></pre>

      <div class="ce-example__output-footer">
        <a href="https://codex.so" style="font-weight: bold;">Made by CodeX</a>
      </div>
    </div>
  </div>

[...]

PR #1553, opened 2021-02-20
Make passed config immutable

hata6502:
https://github.com/codex-team/editor.js/issues/1552


src/components/core.ts
```diff
@@ -116,31 +116,27 @@ export default class Core {
...
      * Make config object
+     *
+     * @type {EditorConfig}
      */
-    if (!_.isObject(config)) {
-      config = {
-        holder: config,
-      };
-    }
+    this.config = _.isObject(config) ? {
```
neSpecc, line 125: let's separate zero-config adjustment and a storing it in the class property into the two parts of code with their own comments. That will be more maintainable.

neSpecc (approved):

gohabereg (approved):

PR #1574, opened 2021-02-27
refactoring(modules): sanitizer module is util now

gohabereg:


neSpecc (changes requested):
At example-dev: 

image

src/components/utils/sanitizer.ts
```diff
@@ -0,0 +1,192 @@
...
+  sanitizeConfig: (toolName: string) => SanitizerConfig | SanitizerConfig
+): Array<Pick<SavedData, 'data' | 'tool'>> {
+  return blocksData.map((block) => {
+    const toolConfig = _.isFunction(sanitizeConfig) ? sanitizeConfig(block.tool) : sanitizeConfig;
+
+    if (_.isEmpty(toolConfig)) {
+      return block;
+    }
+
+    block.data = this.deepSanitize(block.data, toolConfig) as BlockToolData;
```
neSpecc: `this.deepSanitize`  -> `deepSanitize`   ?

src/components/utils/sanitizer.ts
```diff
@@ -0,0 +1,192 @@
...
+
+/**
+ * Sanitize Blocks
+ *
+ * Enumerate blocks and clean data
+ *
+ * @param blocksData - blocks' data to sanitize
+ * @param sanitizeConfig — sanitize config to use or function to get config for Tool
+ */
+export function sanitizeBlocks(
```
neSpecc, line 45: It is the only method that uses the `Block` term and the `SavedData` type. Maybe it is not util? Maybe it should be moved to the BlockManager, for example? What do you think?

gohabereg: I think Sanitizer is a still Editor.js only util, so it's ok to leave it here to not overcrowd BlockManager

khaydarov (approved):

docs/CHANGELOG.md
```diff
@@ -25,9 +25,11 @@
 - `Refactoring` - The Listeners module now is a util.
 - `Refactoring` - The Events module now is a util.
 - `Fix` - Editor Config now immutable [#1552](https://github.com/codex-team/editor.js/issues/1552).
+- `Refactoring` - The Sanitizer module is util now.
```
neSpecc: wrong section

src/components/utils/sanitizer.ts
```diff
@@ -0,0 +1,192 @@
...
+ * CodeX Sanitizer
+ *
+ * @module Sanitizer
+ * Clears HTML from taint tags
+ *
+ * @version 2.0.0
+ *
+ * @example
+ *
+ * Sanitizer.clean(yourTaintString, yourConfig);
```
neSpecc: actualize example

src/components/utils/sanitizer.ts
```diff
@@ -0,0 +1,192 @@
...
+ * Sanitize Blocks
+ *
+ * Enumerate blocks and clean data
+ *
+ * @param blocksData - blocks' data to sanitize
+ * @param sanitizeConfig — sanitize config to use or function to get config for Tool
+ */
+export function sanitizeBlocks(
+  blocksData: Array<Pick<SavedData, 'data' | 'tool'>>,
+  sanitizeConfig: (toolName: string) => SanitizerConfig | SanitizerConfig
```
neSpecc: why the `sanitizeconfig` is a function? maybe the config should be passed calculated?

src/components/modules/saver.ts
```diff
@@ -42,7 +43,9 @@ export default class Saver extends Module {
       });
```
neSpecc, line 43: can we just call `deepSanitize` here and remove the `sanitizeBlocks` method?

src/components/utils.ts
```diff
@@ -642,3 +642,40 @@ export function deprecationAssert(condition: boolean, oldProperty: string, newPr
...
+export function cacheable<Target, Value, Arguments extends any[] = any[]>(
+  target: Target,
+  propertyKey: string,
+  descriptor: PropertyDescriptor
+): PropertyDescriptor {
+  const propertyToOverride = descriptor.value ? 'value' : 'get';
+  const originalMethod = descriptor[propertyToOverride];
+  const cacheKey = `#${propertyKey}Cache`;
+
+  descriptor[propertyToOverride] = function (...args: Arguments): Value {
```
neSpecc, line 665: I think we need to add more comments here.

src/components/utils.ts
```diff
@@ -642,3 +642,40 @@ export function deprecationAssert(condition: boolean, oldProperty: string, newPr
...
+    if (this[cacheKey] === undefined) {
+      this[cacheKey] = originalMethod.apply(this, ...args);
+    }
+
+    return this[cacheKey];
+  };
+
+  if (propertyToOverride === 'get' && descriptor.set) {
```
neSpecc, line 679: and here

src/components/tools/block.ts
```diff
@@ -101,4 +114,47 @@ export default class BlockTool extends BaseTool<IBlockTool> {
...
+      return baseConfig;
+    }
+
+    const toolConfig = {} as SanitizerConfig;
+
+    for (const fieldName in toolRules) {
+      if (Object.prototype.hasOwnProperty.call(toolRules, fieldName)) {
+        const rule = toolRules[fieldName];
+
+        if (_.isObject(rule)) {
```
neSpecc, line 141: describe this statement, please

src/components/tools/base.ts
```diff
@@ -249,7 +249,7 @@ export default abstract class BaseTool<Type extends Tool = Tool> {
    * Returns Tool's sanitizer configuration
    */
   public get sanitizeConfig(): SanitizerConfig {
-    return this.constructable[CommonInternalSettings.SanitizeConfig];
+    return this.constructable[CommonInternalSettings.SanitizeConfig] || {};
```
neSpecc, line 252: why not `undefined` by default?

src/components/modules/toolbar/inline.ts
```diff
@@ -279,70 +279,6 @@ export default class InlineToolbar extends Module<InlineToolbarNodes> {
...
-    const settingsForTool = tool.enabledInlineTools;
-
-    /**
-     * Whether to enable IT for a particular tool is the decision of the editor user.
-     * He can enable it by the inlineToolbar settings for this tool. To enable, he should pass true or strings[]
-     */
-    const enabledForTool = settingsForTool === true || Array.isArray(settingsForTool);
-
-    /**
-     * Disabled by user
```
neSpecc, line 322: There were many useful comments describing all the cases. We need to move it to the new code.

neSpecc (changes requested):

src/components/utils/sanitizer.ts
gohabereg: Each block has own sanitize config. We can pass a map name => config but I think function is better. However you can still pass an object which will be applied for all blocks

src/components/modules/saver.ts
gohabereg: Each block has own sanitize config, so we can't

src/components/tools/base.ts
gohabereg: For Block Tools it will be still {} even if we return undefined here, for inline tools it helps us to get rid of unnecessary checks

src/components/modules/toolbar/inline.ts
```diff
@@ -279,70 +279,6 @@ export default class InlineToolbar extends Module<InlineToolbarNodes> {
...
-    /**
-     * InlineToolbar property of a particular tool
-     */
-    const settingsForTool = tool.enabledInlineTools;
-
-    /**
-     * Whether to enable IT for a particular tool is the decision of the editor user.
-     * He can enable it by the inlineToolbar settings for this tool. To enable, he should pass true or strings[]
-     */
-    const enabledForTool = settingsForTool === true || Array.isArray(settingsForTool);
```
neSpecc, line 319: can't find this statement in a new code. What if a user did not specify the `inlineToolbar`

gohabereg: Nothing will happen. New code checks for true or array, otherwise inline tools will be empty map

test/cypress/tests/tools/BlockTool.spec.ts
```diff
@@ -131,10 +135,87 @@ describe('BlockTool', () => {
...
+        api: {},
+        config: {},
+      } as any);
 
-    expect(tool.sanitizeConfig).to.be.deep.eq(options.constructable.sanitize);
+      const inlineTool2 = new InlineTool({
+        name: 'inlineTool',
+        constructable: class {
+          public static sanitize = {
+            b: true,
```
neSpecc:
```suggestion
            a: true,
```

neSpecc (approved):

khaydarov (approved):

PR #1595, opened 2021-03-14
[Refactoring] Tools

gohabereg:
I've added inner wrappers for tools so we can work with them as with abstract objects

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -16,7 +16,7 @@
 [submodule "example/tools/simple-image"]
 	path = example/tools/simple-image
 	url = https://github.com/editor-js/simple-image
-[submodule "src/components/tools/paragraph"]
+[submodule "src/tools/paragraph"]
 	path = src/tools/paragraph
 	url = https://github.com/editor-js/paragraph
 [submodule "example/tools/marker"]
--- a/.gitmodules
+++ b/.gitmodules
@@ -17,7 +17,7 @@
 	path = example/tools/simple-image
 	url = https://github.com/editor-js/simple-image
 [submodule "src/components/tools/paragraph"]
-	path = src/components/tools/paragraph
+	path = src/tools/paragraph
 	url = https://github.com/editor-js/paragraph
 [submodule "example/tools/marker"]
 	path = example/tools/marker
--- a/src/components/block/index.ts
+++ b/src/components/block/index.ts
@@ -1,6 +1,6 @@
 import {
   BlockAPI as BlockAPIInterface,
-  BlockTool,
+  BlockTool as IBlockTool,
   BlockToolConstructable,
   BlockToolData,
   BlockTune,
@@ -16,12 +16,13 @@ import * as _ from '../utils';
 import ApiModules from '../modules/api';
 import BlockAPI from './api';
 import { ToolType } from '../modules/tools';
+import SelectionUtils from '../selection';
+import BlockTool from '../tools/block';
 
 /** Import default tunes */
 import MoveUpTune from '../block-tunes/block-tune-move-up';
 import DeleteTune from '../block-tunes/block-tune-delete';
 import MoveDownTune from '../block-tunes/block-tune-move-down';
-import SelectionUtils from '../selection';
 
 /**
  * Interface describes Block class constructor argument
@@ -40,12 +41,7 @@ interface BlockConstructorOptions {
   /**
    * Tool's class or constructor function
    */
-  Tool: BlockToolConstructable;
-
-  /**
-   * Tool settings from initial config
-   */
-  settings: ToolSettings;
+  tool: BlockTool;
 
   /**
    * Editor's API methods
@@ -116,6 +112,8 @@ export default class Block {
    * Instance of the Tool Block represents
    */
   public tool: BlockTool;
+
+  public toolInstance: IBlockTool;
 
   /**
    * Class blueprint of the ool Block represents
@@ -209,27 +207,20 @@ export default class Block {
   constructor({
     name,
     data,
-    Tool,
-    settings,
+    tool,
     api,
     readOnly,
   }: BlockConstructorOptions) {
     this.name = name;
-    this.class = Tool;
-    this.settings = settings;
-    this.config = settings.config || {};
+    this.settings = tool.settings;
+    this.config = tool.settings.config || {};
     this.api = api;
     this.blockAPI = new BlockAPI(this);
 
     this.mutationObserver = new MutationObserver(this.didMutated);
 
-    this.tool = new Tool({
-      data,
-      config: this.config,
-      api: this.api.getMethodsForTool(name, ToolType.Block),
-      block: this.blockAPI,
-      readOnly,
-    });
+    this.tool = tool;
+    this.toolInstance = tool.instance(data, this.blockAPI, readOnly);
 
     this.holder = this.compose();
     /**
@@ -349,7 +340,7 @@ export default class Block {
    * @returns {object}
    */
   public get sanitize(): SanitizerConfig {
-    return this.tool.sanitize;
+    return this.tool.sanitizeConfig;
   }
 
   /**
@@ -359,7 +350,7 @@ export default class Block {
    * @returns {boolean}
    */
   public get mergeable(): boolean {
-    return _.isFunction(this.tool.merge);
+    return _.isFunction(this.toolInstance.merge);
   }
 
   /**
@@ -502,7 +493,7 @@ export default class Block {
     /**
      * call Tool's method with the instance context
      */
-    if (this.tool[methodName] && this.tool[methodName] instanceof Function) {
+    if (this.toolInstance[methodName] && this.toolInstance[methodName] instanceof Function) {
       if (methodName === BlockToolAPI.APPEND_CALLBACK) {
         _.log(
           '`appendCallback` hook is deprecated and will be removed in the next major release. ' +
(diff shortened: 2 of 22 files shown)


.gitmodules
```diff
@@ -17,7 +17,7 @@
 	path = example/tools/simple-image
 	url = https://github.com/editor-js/simple-image
 [submodule "src/components/tools/paragraph"]
```
neSpecc:
```suggestion
[submodule "src/tools/paragraph"]
```

src/components/block/index.ts
```diff
@@ -672,6 +661,24 @@ export default class Block {
     this.removeInputEvents();
   }
 
+  /**
+   * Call Tool instance destroy method
+   */
+  public destroy(): void {
+    if (_.isFunction(this.toolInstance.destroy)) {
+      this.toolInstance.destroy();
```
neSpecc, line 669: Don't forget to add it to the docs site.

src/components/tools/base.ts
```diff
@@ -0,0 +1,175 @@
...
+import * as _ from '../utils';
+
+export enum UserSettings {
+  Shortcut = 'shortcut',
+  Toolbox = 'toolbox',
+  EnabledInlineTools = 'inlineToolbar',
+  Config = 'config',
+}
+
+export enum InternalSettings {
```
neSpecc: Maybe we need to separate `UserSettings` and `InternalSettings` for Block Tools and Inline Tool and other tools?

src/components/tools/base.ts
```diff
@@ -0,0 +1,175 @@
...
+  Title = 'title', // for Inline Tools. Block Tools can pass title along with icon through the 'toolbox' static prop.
+  Shortcut = 'shortcut',
+  Toolbox = 'toolbox',
+  SanitizeConfig = 'sanitize',
+  ConversionConfig = 'conversionConfig',
+  IsReadOnlySupported = 'isReadOnlySupported',
+  PasteConfig = 'pasteConfig'
+}
+
+export type ToolConfig = Omit<ToolSettings, 'class'>
```
neSpecc: I would suggest naming it ToolOptions or something like that, to avoid conflicts with tools `config` property.

src/components/tools/base.ts
```diff
@@ -0,0 +1,175 @@
...
+   */
+  public type: ToolType;
+
+  /**
+   * Tool name specified in EditorJS config
+   */
+  public name: string;
+
+  /**
+   * Flag show is current Tool internal or not
```
neSpecc: need to clarify what does `internal tool` mean

src/components/tools/base.ts
```diff
@@ -0,0 +1,175 @@
...
+
+  /**
+   * Tool's constructable blueprint
+   */
+  protected constructable: ToolConstructable;
+
+  /**
+   * Editor default tool
+   */
+  protected defaultTool: string;
```
neSpecc: why not using `isDefault: boolean`?

src/components/tools/base.ts
```diff
@@ -0,0 +1,175 @@
...
+   * Returns Tool's sanitizer configuration
+   */
+  public get sanitizeConfig(): SanitizerConfig {
+    return this.constructable[InternalSettings.SanitizeConfig];
+  }
+
+  /**
+   * Returns true if current Tool is default
+   */
+  public get isDefault(): boolean {
```
neSpecc: Is it related only to Block Tools? If so, we need to move it from the abstract class to the `BlockTool`

src/components/modules/blockEvents.ts
```diff
@@ -5,6 +5,7 @@ import Module from '../__module';
 import * as _ from '../utils';
 import SelectionUtils from '../selection';
 import Flipper from '../flipper';
+import BlockTool from '../tools/block';
```
neSpecc: where this class is used?

src/components/modules/tools.ts
```diff
@@ -1,19 +1,21 @@
...
-  ToolConfig,
+  Tool,
   ToolConstructable,
   ToolSettings
 } from '../../../types';
 import BoldInlineTool from '../inline-tools/inline-tool-bold';
 import ItalicInlineTool from '../inline-tools/inline-tool-italic';
 import LinkInlineTool from '../inline-tools/inline-tool-link';
-import Stub from '../tools/stub';
+import Stub from '../../tools/stub';
```
neSpecc, line 13: The Stub tool is just an implementation of the Block Tool, so it should not be placed at the same level as abstract tools classes. It would better to move it to the `tools/block/stub`

src/components/modules/tools.ts
```diff
@@ -63,136 +65,93 @@ export default class Tools extends Module {
...
+   * Tools objects factory
    */
-  public readonly toolsClasses: { [name: string]: ToolConstructable } = {};
+  private factory: ToolsFactory;
 
   /**
    * Tools` classes available to use
    */
-  private readonly toolsAvailable: { [name: string]: ToolConstructable } = {};
+  private readonly toolsAvailable: Map<string, InlineTool | BlockTool | BlockTune> = new Map();
```
neSpecc:
```
type ToolClass = InlineTool | BlockTool | BlockTune;
```

src/components/block/index.ts
gohabereg: It should be already there

src/components/tools/base.ts
gohabereg: Yeah, I also thought about that

src/components/modules/tools.ts
gohabereg: There is a confusion about dir names. We have tools for paragraph and stub, I just moved it one level up

src/components/modules/paste.ts
```diff
@@ -20,7 +20,7 @@ interface TagSubstitute {
    *
    * @type {string}
```
neSpecc:
```suggestion
```

src/components/modules/paste.ts
```diff
@@ -46,7 +46,7 @@ interface PatternSubstitute {
...
```
neSpecc:
```suggestion
```

src/components/modules/readonly.ts
```diff
@@ -37,14 +37,16 @@ export default class ReadOnly extends Module {
...
+    const { block } = Tools;
     const toolsDontSupportReadOnly: string[] = [];
 
-    Object.entries(blockTools).forEach(([name, tool]) => {
-      if (!Tools.isReadOnlySupported(tool)) {
-        toolsDontSupportReadOnly.push(name);
-      }
-    });
+    Array
+      .from(block.entries())
```
neSpecc: I think, it's better to name Block Tools getter `blockTools` rather than `block` — it will be more readable

src/components/tools/base.ts
```diff
@@ -0,0 +1,178 @@
+import { ToolType } from '../modules/tools';
+import { Tool, ToolConstructable, ToolSettings } from '../../../types/tools';
+import { API, SanitizerConfig } from '../../../types';
+import * as _ from '../utils';
+
+export enum UserSettings {
```
neSpecc, line 9: all the enums should be documented

src/components/modules/readonly.ts
neSpecc: and `inlineTools`, `blockTunes`

src/components/modules/tools.ts
```diff
@@ -424,40 +263,49 @@ export default class Tools extends Module {
...
-    for (const toolName in this.toolsClasses) {
-      if (Object.prototype.hasOwnProperty.call(this.toolsClasses, toolName)) {
-        const toolClass = this.toolsClasses[toolName];
-        const toolConfig = this.toolsSettings[toolName][this.USER_SETTINGS.CONFIG];
+  /**
+   * Fail callback
+   *
+   * @param {object} data - append tool to unavailable list
+   */
+  private fallback(data: { toolName: string }): void {
```
neSpecc: naming is not clear.

src/components/modules/toolbar/blockSettings.ts
```diff
@@ -229,8 +229,10 @@ export default class BlockSettings extends Module<BlockSettingsNodes> {
    * Add Tool's settings
    */
   private addToolSettings(): void {
-    if (_.isFunction(this.Editor.BlockManager.currentBlock.tool.renderSettings)) {
-      $.append(this.nodes.toolSettings, this.Editor.BlockManager.currentBlock.tool.renderSettings());
+    const settingsElement = this.Editor.BlockManager.currentBlock.renderSettings();
```
khaydarov, line 232:
```suggestion
    const settingsElement = this.Editor.BlockManager.currentBlock?.renderSettings();
```

gohabereg: `currentBlock` must be defined there. I'd suggest to leave it without `?` because we will know something is wrong

neSpecc (approved):

src/components/modules/toolbar/blockSettings.ts
khaydarov: but it will throw an exception if currentBlock is null

khaydarov (approved):

src/components/modules/toolbar/blockSettings.ts
gohabereg: I meant it cannot be null here, but if it is we will got an exception — that means something is not correct in the code

PR #1596, opened 2021-03-15
[Feature] Block Tunes API

gohabereg:


christoph-kluge:
Is there any time estimation when this will go into the main branch?

src/components/block/index.ts
```diff
@@ -690,11 +685,47 @@ export default class Block {
         pluginsContent = this.toolInstance.render();
 
     contentNode.appendChild(pluginsContent);
-    wrapper.appendChild(contentNode);
+
+    let wrappedContentNode: HTMLElement = contentNode;
```
neSpecc, line 713: add a comment here describing tunes wrapping

src/components/block/index.ts
```diff
@@ -690,11 +685,47 @@ export default class Block {
...
+   *
+   * @param tunesData - current Block tunes data
+   * @private
+   */
+  private composeTunes(tunesData: {[name: string]: BlockTuneData}): void {
+    [
+      ...this.tunes.externalTools.values(),
+      ...this.tunes.internalTools.values(),
+    ].forEach((tune) => {
+      const collection = tune.isInternal ? this.defaultTunesInstances : this.tunesInstances;
```
neSpecc, line 739: What's the reason for separation for `this.defaultTunesInstances` and `this.tunesInstances`? As I can see, all the usages are merging them back.

src/components/block/index.ts
```diff
@@ -145,6 +148,10 @@ export default class Block {
    */
   private readonly toolInstance: IBlockTool;
 
+  private readonly tunesInstances: Map<string, IBlockTune> = new Map();
+  private readonly defaultTunesInstances: Map<string, IBlockTune> = new Map();
+  private unavailableTunesData: {[name: string]: BlockTuneData} = {};
```
neSpecc, line 164: describe how it will work in a comment, please.

src/components/block/index.ts
```diff
@@ -690,11 +685,47 @@ export default class Block {
...
 
     contentNode.appendChild(pluginsContent);
-    wrapper.appendChild(contentNode);
+
+    let wrappedContentNode: HTMLElement = contentNode;
+
+    [...this.tunesInstances.values(), ...this.defaultTunesInstances.values()]
+      .forEach((tune) => {
+        if (_.isFunction(tune.wrap)) {
+          wrappedContentNode = tune.wrap(wrappedContentNode);
```
neSpecc: do we need to add try-catch statement here? To avoid rendering problems in case of an error on the tune side?

neSpecc: same thing at the `save()` method

src/components/modules/saver.ts
```diff
@@ -104,10 +104,16 @@ export default class Saver extends Module {
...
 
-      blocks.push({
+      const output: any = {
         type: tool,
         data,
-      });
+      };
+
+      if (!_.isEmpty(tunes)) {
+        output.tunes = tunes;
```
neSpecc, line 113: maybe we need to preserve the `tunes` field event it is empty?

src/components/modules/tools.ts
```diff
@@ -248,9 +187,56 @@ export default class Tools extends Module {
...
+          .from(this.blockTunes.entries())
+          .filter(([, tune]) => names.includes(tune.name))
+          .concat([ ...this.blockTunes.internalTools.entries() ])
+      );
+    }
+
+    const defaultTuneNames = this.config.tunes;
+
+    if (Array.isArray(defaultTuneNames)) {
+      return new ToolsCollection<BlockTune>(
```
neSpecc, line 189: duplicated code fragments

src/components/block/index.ts
gohabereg: Tune instance doesn't provide any flag if it's internal or not, so we need to split them to know how to render in block settings

src/components/modules/saver.ts
gohabereg: I think better to omit it to decrease output's size and explicitly state that block doesn't have any stateful tunes

src/components/modules/tools.ts
gohabereg: Let's keep it for now, I have idea to refactor it later

src/components/modules/toolbar/blockSettings.ts
```diff
@@ -237,10 +237,13 @@ export default class BlockSettings extends Module<BlockSettingsNodes> {
   }
 
   /**
-   * Add default settings
+   * Add external tunes: provided by user and default ones
```
neSpecc: why `external`? Looks like it appends both internal and external

types/configs/editor-config.d.ts
```diff
@@ -95,4 +95,9 @@ export interface EditorConfig {
    * Defines default toolbar for all tools.
    */
   inlineToolbar?: string[]|boolean;
+
+  /**
+   * Common BlockTune names
```
neSpecc:
```suggestion
   * Common Block Tunes list. Will be added to all the blocks which do not specify their own 'tunes' set
```

types/data-formats/output-data.d.ts
```diff
@@ -15,6 +16,8 @@ export interface OutputBlockData<Type extends string = string, Data extends obje
    * Saved Block data
    */
   data: BlockToolData<Data>;
+
+  tunes?: {[name: string]: BlockTuneData};
```
neSpecc, line 23:
```suggestion
  /**
   * Block Tunes data 
   */
  tunes?: {[name: string]: BlockTuneData};
```

src/components/tools/tune.ts
```diff
@@ -1,21 +1,36 @@
...
-    return undefined;
+  protected readonly constructable: BlockTuneConstructable;
+
+  /**
+   * Constructs new BlockTune instance from constructable
+   *
+   * @param data - Tune data
+   * @param block - Block API object
+   */
+  public instance(data: BlockTuneData, block: BlockAPI): IBlockTune {
```
khaydarov: i think `create` or `build` is more suitable here (for function name)

docs/block-tunes.md
```diff
@@ -0,0 +1,166 @@
+# Tools for the Inline Toolbar
```
neSpecc:
```suggestion
# Block Tunes
```

docs/block-tunes.md
```diff
@@ -0,0 +1,166 @@
...
+
+Similar with [Tools](tools.md) represented Blocks, you can create Block Tunes and connect it to particular Tool or for all Tools.
+
```
neSpecc, line 4:
```suggestion

Block Tunes allows you to set any additional options to Blocks. For example, with corresponded Block Tunes you can mark Block as «spoiler», give it an anchor, set a background, and so on.

```

docs/block-tunes.md
```diff
@@ -0,0 +1,166 @@
...
+## Base structure
+
+Tune's class should have a `isTune` property (static getter) set tu `true`.
```
neSpecc:
```suggestion
Tune's class should have the `isTune` property (static getter) set to `true`.
```

docs/block-tunes.md
```diff
@@ -0,0 +1,166 @@
...
+
+Block Tune must implement render method which returns an HTML Element inserted to block settings.
```
neSpecc:
```suggestion
Block Tune must implement the `render()` method which returns an HTML Element that will be appended to the Block Settings panel.
```

docs/block-tunes.md
```diff
@@ -0,0 +1,166 @@
...
+
+type | description |
+-- | -- |
+`any` | any data you want to save |
+
+---
+
+### static prepare()
+
+If you need to prepare some data for Tune (eg. load external script, create HTML nodes in the document, etc) you can use static prepare method.
```
neSpecc:
```suggestion
If you need to prepare some data for Tune (eg. load external script, create HTML nodes in the document, etc) you can use the static `prepare()` method.
```

src/components/block/index.ts
```diff
@@ -568,50 +605,23 @@ export default class Block {
...
-        api: this.api.getMethodsForTool(name, ToolType.Tune),
-        settings: this.config,
-      });
-    });
-  }
-
   /**
    * Enumerates initialized tunes and returns fragment that can be appended to the toolbars area
    *
    * @returns {DocumentFragment}
```
khaydarov:
```suggestion
   * @returns {DocumentFragment[]}
```

src/components/tools/base.ts
```diff
@@ -226,10 +252,31 @@ export default abstract class BaseTool<Type extends Tool> {
...
+   * Returns true if Tools is inline
+   */
+  public isInline(): this is InlineTool {
+    return this.type === ToolType.Inline;
+  }
+
+  /**
+   * Returns true if Tools is block
+   */
+  public isBlock(): this is BlockTool {
```
khaydarov, line 265: maybe it is better to rename this method to `isBlockTool`? because Block is a product of a tool

khaydarov (approved):

src/components/tools/base.ts
gohabereg: Usage would be `tool.isBlock()`, so I think it's ok

neSpecc (approved):

PR #1601, opened 2021-03-16
refactoring(modules): Tooltip module is util now

robonetphy:


src/components/utils/tooltip.ts
```diff
@@ -48,3 +46,5 @@ export default class Tooltip extends Module {
     this.lib.onHover(element, content, options);
   }
 }
+
+export default new Tooltip();
```
neSpecc: its better to create an instance in a class constructor or in the method where it used, rather than use a singletone

src/components/modules/api/tooltip.ts
```diff
@@ -1,6 +1,7 @@
 import { Tooltip } from '../../../../types/api';
 import { TooltipContent, TooltipOptions } from 'codex-tooltip';
 import Module from '../../__module';
+import ITooltip from '../../utils/tooltip';
```
neSpecc: `I` means `Interface`, so it should be used near type, not util itself

neSpecc (changes requested):

neSpecc (approved):

gohabereg (approved):

src/components/modules/toolbar/inline.ts
```diff
@@ -590,7 +612,7 @@ export default class InlineToolbar extends Module<InlineToolbarNodes> {
    */
   private addTool(tool: InlineTool): void {
     const {
-      Tooltip,
+      Tools,
```
neSpecc: why did it change?

PR #1613, opened 2021-03-27
Make submodules installation hint more explicit

a-kon:
Hi there.

If you open an example HTML file now you will see 2 errors:
`No core bundle file found. Run yarn build`
`No submodules found. Run yarn pull_tools`

But you're not able to build a core bundle w/o submodules and if you tried you would get a confusing error about path resolving.
I've swapped those messages to make it a bit simpler.

**UPD.**
I've just added a script to check missing submodules to run it before a build. Output is shown below.
pic

Also, I noticed that `.gitmodules` contains `example/tools/underline` is missing in `git submodule status` => `git submodule update --init --recursive`. Is it an error or I'm doing something wrong?

### DIFF
--- a/check_submodules.sh
+++ b/check_submodules.sh
@@ -1,0 +1,13 @@
+submodules=$(grep path .gitmodules | sed 's/.*= //')
+missing=false
+
+while read -r line ; do
+    if ! test -f "$line/.git"; then
+        echo "    - $line submodule is not initialized."
+        missing=true
+    fi
+done <<< "$submodules"
+
+if [ "$missing" = true ] ; then
+    printf "\n(!) Some modules are missing, consider running yarn pull_tools\n\n"
+fi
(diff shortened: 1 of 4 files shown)


hata6502:
I've found an error in `yarn checkout_tools`.

`nested-list` is not checked out.
```
$ yarn checkout_tools
Entering 'example/tools/nested-list'
fatal: couldn't find remote ref master
fatal: run_command returned non-zero status for example/tools/nested-list
.
error Command failed with exit code 128.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

`nested-list`'s default branch is `main` not `master`, So this command fails.
https://github.com/editor-js/nested-list

hata6502:
And... the bundle file is missing in GitHub repository.
https://github.com/editor-js/nested-list
Screenshot from 2021-04-08 21-54-45

gohabereg:
Seems we need to add for tools:
`git submodule foreach yarn`
`git submodule foreach yarn build`

PR #1620, opened 2021-03-31
feat: support merging multiple blocks

### LINKED ISSUE #1617: Select and Convert multiple blocks to one block💡
**Feature:** I have a bunch paragraphs I'd like to arrange into a list. 

**Behaviour**: Select multiple blocks, click, and the conversion tool appears and I can make a conversion into one block. Conversion config would need to have a new field on how to merge multiple blocks.

Are there any alternatives?
It's very tedious to create and modify lists from existing content. The only alternative is to backspace all paragraphs into one paragraph and then convert to list - which isn't intuitive.

evanjmg:
Closes https://github.com/codex-team/editor.js/issues/1617
By default uses import and export methods and will override import method if a mergeImport method is provided.
The mergeImport method allows a tool to indicate how to merge the blocks rather than simply string.join(' ').


evanjmg:
@neSpecc  is next branch broken? The netlify app doesn't seem to be building the Nested list git submodule.

evanjmg:
@neSpecc any update on this?

neSpecc:
Please provide more detailed description so we can better understand your solution

PR #1626, opened 2021-04-05
Added handling for delete key

### LINKED ISSUE #821: Backspace and Delete keystroke
One of my online content team members pointed out that she want to used the delete key on her keyboard to merge to paragraph blocks together and that is not working. The backspace is working fine but the delete key not. I included a gif to demonstrate

Image of Yaktocat

### LINKED ISSUE #1315: Deleting multiple selected blocks using Delete key does not work
Deleting multiple selected blocks using Delete key does not work but works using Backspace.

Steps to reproduce:
1. Create multiple blocks
2. Select the blocks
3. Press delete

Expected behavior:

The selected blocks get deleted from the document. 

Screenshots:

Device, Browser, OS:

Windows 10, Chromium Edge

Editor.js version:

Tested with 2.18 and master current.

Editor without custom plugins.

jeremyVignelles:
Pressing delete at the end of a block merges blocks (fixes #821)
Pressing delete after a rectangular block selection removes the blocks (fixes #1315)

### DIFF
--- a/src/components/modules/blockEvents.ts
+++ b/src/components/modules/blockEvents.ts
@@ -436,11 +436,20 @@ export default class BlockEvents extends Module {
      * other case will handle as usual ARROW LEFT behaviour
      */
     if (blockToMerge.name !== targetBlock.name || !targetBlock.mergeable) {
-      /** If target Block doesn't contain inputs or empty, remove it */
+      /** If target Block doesn't contain inputs or empty, remove it. This means the user pressed backspace on the blockToMerge */
       if (targetBlock.inputs.length === 0 || targetBlock.isEmpty) {
         BlockManager.removeBlock(BlockManager.currentBlockIndex - 1);
 
         Caret.setToBlock(BlockManager.currentBlock);
+        Toolbar.close();
+
+        return;
+      }
+
+      /** If blockToMerge doesn't contain input or empty, remove it. This means the user pressed delete on the targetBlock */
+      if (blockToMerge.inputs.length === 0 || targetBlock.isEmpty) {
+        BlockManager.removeBlock(BlockManager.currentBlockIndex + 1);
+
         Toolbar.close();
 
         return;
--- a/src/components/modules/blockEvents.ts
+++ b/src/components/modules/blockEvents.ts
@@ -332,7 +332,6 @@ export default class BlockEvents extends Module {
       SelectionUtils.isCollapsed &&
       currentBlock.currentInput === currentBlock.firstInput &&
       !isFirstBlock;
-
     if (canMergeBlocks) {
       /**
        * preventing browser default behaviour
@@ -372,10 +371,18 @@ export default class BlockEvents extends Module {
         BlockManager.removeBlock();
       }
 
-      Caret.setToBlock(
-        BlockManager.currentBlock,
-        Caret.positions.END
-      );
+      // The removal placed the caret to the previous block, try to put it to the next block (if not the last)
+      if (BlockManager.nextBlock) {
+        Caret.setToBlock(
+          BlockManager.nextBlock,
+          Caret.positions.START
+        );
+      } else {
+        Caret.setToBlock(
+          BlockManager.currentBlock,
+          Caret.positions.END
+        );
+      }
 
       /** Close Toolbar */
       this.Editor.Toolbar.close();
--- a/src/components/modules/blockEvents.ts
+++ b/src/components/modules/blockEvents.ts
@@ -359,13 +359,13 @@ export default class BlockEvents extends Module {
     /**
      * Check if Block should be removed by current Delete keydown
      */
-    if (currentBlock.selected || (currentBlock.isEmpty && currentBlock.currentInput === currentBlock.firstInput)) {
+    if (currentBlock.selected || (currentBlock.isEmpty && currentBlock.currentInput === currentBlock.lastInput)) {
       event.preventDefault();
 
       const index = BlockManager.currentBlockIndex;
 
       if (BlockManager.nextBlock && BlockManager.nextBlock.inputs.length === 0) {
-        /** If previous block doesn't contain inputs, remove it */
+        /** If next block doesn't contain inputs, remove it */
         BlockManager.removeBlock(index + 1);
       } else {
         /** If block is empty, just remove it */
--- a/src/components/modules/blockEvents.ts
+++ b/src/components/modules/blockEvents.ts
@@ -29,6 +29,7 @@ export default class BlockEvents extends Module {
       case _.keyCodes.BACKSPACE:
         this.backspace(event);
         break;
+
       case _.keyCodes.DELETE:
         this.deletePressed(event);
         break;
--- a/src/components/modules/blockEvents.ts
+++ b/src/components/modules/blockEvents.ts
@@ -332,6 +332,7 @@ export default class BlockEvents extends Module {
       SelectionUtils.isCollapsed &&
       currentBlock.currentInput === currentBlock.firstInput &&
       !isFirstBlock;
+
     if (canMergeBlocks) {
       /**
        * preventing browser default behaviour
--- a/src/components/modules/blockEvents.ts
+++ b/src/components/modules/blockEvents.ts
@@ -5,6 +5,7 @@ import Module from '../__module';
 import * as _ from '../utils';
 import SelectionUtils from '../selection';
 import Flipper from '../flipper';
+import Block from '../block';
 
 /**
  *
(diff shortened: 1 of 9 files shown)


src/components/modules/blockEvents.ts
```diff
@@ -28,6 +29,9 @@ export default class BlockEvents extends Module {
       case _.keyCodes.BACKSPACE:
         this.backspace(event);
         break;
+      case _.keyCodes.DELETE:
```
gohabereg:
```suggestion
        break;

      case _.keyCodes.DELETE:
```

src/components/modules/blockEvents.ts
```diff
@@ -332,17 +336,84 @@ export default class BlockEvents extends Module {
...
+   */
+  private deletePressed(event: KeyboardEvent): void {
+    const { BlockManager, BlockSelection, Caret } = this.Editor;
+    const currentBlock = BlockManager.currentBlock;
+    const tool = currentBlock.tool;
+
+    /**
+     * Check if Block should be removed by current Delete keydown
+     */
+    if (currentBlock.selected || (currentBlock.isEmpty && currentBlock.currentInput === currentBlock.firstInput)) {
```
gohabereg: The last condition is not clear for me. I suppose there should be `currentBlock.lastInput` instead
But I believe such behaviour is not really intuitive, let me explain why.
For example you have list tool, you create couple of lines (but don't type any text there) and press delete key on the last line. You wouldn't expect this will delete the list block

src/components/modules/blockEvents.ts
```diff
@@ -332,17 +336,84 @@ export default class BlockEvents extends Module {
...
+      event.preventDefault();
+
+      const index = BlockManager.currentBlockIndex;
+
+      if (BlockManager.nextBlock && BlockManager.nextBlock.inputs.length === 0) {
+        /** If previous block doesn't contain inputs, remove it */
+        BlockManager.removeBlock(index + 1);
+      } else {
+        /** If block is empty, just remove it */
+        BlockManager.removeBlock();
```
gohabereg, line 372: See comment above, I don't think delete key should remove current block in any case

src/components/modules/blockEvents.ts
```diff
@@ -332,17 +336,84 @@ export default class BlockEvents extends Module {
...
+     * Check if Block should be removed by current Delete keydown
+     */
+    if (currentBlock.selected || (currentBlock.isEmpty && currentBlock.currentInput === currentBlock.firstInput)) {
+      event.preventDefault();
+
+      const index = BlockManager.currentBlockIndex;
+
+      if (BlockManager.nextBlock && BlockManager.nextBlock.inputs.length === 0) {
+        /** If previous block doesn't contain inputs, remove it */
+        BlockManager.removeBlock(index + 1);
```
gohabereg, line 369: Indeed there should be `lastInput` above

gohabereg (changes requested):

src/components/modules/blockEvents.ts
jeremyVignelles: Indeed, looks like a bad copy/paste of that line: https://github.com/codex-team/editor.js/blob/3b20c4a139faf46a91db84f7f1998ea1e2a9b511/src/components/modules/blockEvents.ts#L288

> For example you have list tool, you create couple of lines (but don't type any text there) and press delete key on the last line. You wouldn't expect this will delete the list block

I just copied the behavior of the backspace key. Why would it make sense when pressing backspace and not when pressing delete ? Both have the same semantics for me, only the direction changes.

gohabereg: Well, it makes sense for paragraph tool, but for more complex tools it looks unintuitive for me. When I press delete I expect next entity to be deleted (or merged with the current one), but not current entity deletion. I might be wrong, so could you please apply changes and I'll take a look in action

src/components/modules/blockEvents.ts
jeremyVignelles: As a windows user, I'm using the delete key more often for such tasks than the backspace key.
If you are selecting a block, pressing delete should delete it ( isn't that the point of the delete key, after all ? 😄 )

I have to admit though, that I copied that comment without understanding what it means. I don't see how it could be empty.

jeremyVignelles:
Thanks for the review. Should I rebase on the `next` branch to integrate the new commits ?

gohabereg:
@jeremyVignelles yes please

src/components/modules/blockEvents.ts
jeremyVignelles: Changes are already made and rebased.

Being able to delete what has been selected with a rectangular selection is exactly the point of #1315 , and it is working as I would expect.

During my testing, I found a bug on firefox with the current version (on editorjs.io), that somehow disappeared with my version. When pressing delete on an empty list item, the item is removed and the carret is placed at the end of the previous item. Pressing enter does not create a new item anymore.

gohabereg:
For some reason I can't merge anything except two paragraphs using delete key and also have noticed few more uncovered cases. Will investigate it on the weekend and get back to you

jeremyVignelles:
Do the other "merges" work with the backspace key?

gohabereg:
You are right, they shouldn't, but there are some of my discovers:
| Pre-requirements | Steps | Actual | Expected |
| -- | -- | -- | -- |
| Two not-empty different blocks one by one | Set cursor to the end of the first block and press delete | Nothing is happen | If the same happens for backspace, caret is set to the previous block. So for delete I'd expect caret to set to the next block |
| Not-empty block and then an empty block | - "" - |  - "" - | I'd expect empty Block to be deleted |
| - "" - | Set cursor to the empty block and press delete | Block is deleted, caret is set to previous block | I'd expect caret to be set to the next block |
| Not-empty block and then block without inputs (eg Delimiter) | Set cursor to the end of the first block and press delete | Nothing is happen | I'd expect Delimiter to be deleted |

jeremyVignelles:
> Two not-empty different blocks one by one

I'm not sure, but I'd consider it a bug of the backspace key. I'm not sure that's something I would expect to be normal : If I pressed backspace instead of the left arrow key, that's because I wanted to destroy something, which obviously didn't work but had a "side effect" of moving the caret.

EDIT : This seems to be explicitely wanted : https://github.com/codex-team/editor.js/blob/32ac52a62fb1e36dddaad1f9fe224e46a931f056/src/components/modules/blockEvents.ts#L357 . I'm still not sure that's a desired behavior, but I can match that the other way around

## Other points
Seems like the mergeBlocks methods assume that targetBlock is just before blockToMerge, but it doesn't handle well the cases you mentionned. I'm digging further and I'll get back to you

jeremyVignelles:
So I fixed the 3 last points (in 3 different commits to make it easier to review). Thanks for testing and pointing that out !

As for the first point, I'm not sure how this should be handled:

https://github.com/codex-team/editor.js/blob/32ac52a62fb1e36dddaad1f9fe224e46a931f056/src/components/modules/blockEvents.ts#L370-L372

In this block, I dont know in which direction the merge occurred. Sure, I could detect it by the position of the caret, or by passing an additional parameter. Is that really something useful though ?

For example, I'm on the image caption control :
image
I made an error, and I press backspace to the beginning of the input, as I am not patient, I'm holding down the backspace key, until everything is removed. If I don't release the key on time, the caret is moved to the previous block, and starts removing code above.

gohabereg:
@jeremyVignelles agree the first point is arguable, I'm ok with the current behaviour, but it conflicts with backspace behaviour. Let's listen to @neSpecc and @khaydarov opinion.

I've added some test cases, hope you don't mind

gohabereg (approved):

jeremyVignelles:
Great, thanks for the test case, I never used cypress, I'll see how it works then 🙂

jeremyVignelles:
Anything else is needed here ? Can it be reviewed now ?

gohabereg:
@neSpecc @khaydarov take a look please

hyun-park:
Any updates on this?

jeremyVignelles:
Just merged the latest "next" and tried to resolve the conflicts that I had.
Ready again for review

mbagsik00:
Hi, may I know what's the update on this? Thanks

NeoKms:
Hello! When?

julianpeterson1:
Hi there, we're also waiting for this solution, thanks!

kimchi-hent:
Any progress on this?

cicelle:
Any news on that front ?

neSpecc:
Resolved by #2402
Available starting 2.28.0-rc.1
2.28 will be released this week.

Issue #1645, opened 2021-04-11
[Bug] The copied HTML data from Editor.js have wrong <p>


hata6502:
Describe a bug.
The copied HTML data from Editor.js have wrong `<p>`.
`<p>` can't contain `<h1>`~`<h6>` tags by HTML syntax rule.
https://stackoverflow.com/questions/15656785/should-a-heading-be-inside-or-outside-a-p

Steps to reproduce:
1. Instert the following debug code.

`src/components/modules/rectangleSelection.ts` line 307
```typescript
    const textHTML = fakeClipboard.innerHTML;

    console.log(textHTML);  // debug code

    e.clipboardData.setData('text/plain', textPlain);
```

2. Copy an heading block by `ctrl + c` in `example/example-dev.html`

![Screenshot from 2021-04-11 22-27-54](https://user-images.githubusercontent.com/7702653/114306000-5eadd080-9b15-11eb-94f1-0458cf7eaa53.png)

3. The wrong HTML is output to console.

```html
<p><h2>Editor.js</h2></p>
```

Expected behavior: `<div>` should be used instead of `<p>`.

Editor.js version: 2.20.1

PR #1646, opened 2021-04-11
Use <div> tag instead of <p> in the copied HTML data

hata6502:
https://github.com/codex-team/editor.js/issues/1645

### DIFF
--- a/src/components/modules/blockSelection.ts
+++ b/src/components/modules/blockSelection.ts
@@ -296,10 +296,10 @@ export default class BlockSelection extends Module {
 
     this.selectedBlocks.forEach((block) => {
       /**
-       * Make <p> tag that holds clean HTML
+       * Make <div> tag that holds clean HTML
        */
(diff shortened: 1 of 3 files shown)


src/components/modules/blockSelection.ts
```diff
@@ -296,10 +296,10 @@ export default class BlockSelection extends Module {
 
     this.selectedBlocks.forEach((block) => {
       /**
-       * Make <p> tag that holds clean HTML
+       * Make <div> tag that holds clean HTML
        */
       const cleanHTML = clean(block.holder.innerHTML, this.sanitizerConfig);
-      const fragment = $.make('p');
+      const fragment = $.make('div');
```
gohabereg, line 302: I think we better to check if cleanHTML is already a block tag (eg. h1-6 or ul/ol) and do not wrap it in this case.
But usage of `div` for paragraph would be semantically wrong I guess (and also seems would break pasting back)

PR #1667, opened 2021-04-20
Unique ids in blocks with nanoid

jggc:
- I based my work upon #1493 and #1585 
- Cherry picked commits on top of next branch to make merging easier
- Use nanoid library to generate block id

I tested manually a bit and it works nicely with some limitations (Yes the text is gibberish)
Screenshot from 2021-04-20 15-32-15

### What works
- Generates an id when a block is created
- ID stays the same when block is updated or moved
### What does not
- ID changes when block type changes (paragraph -> header, header -> list for example)


gohabereg (approved):
Change log should be added

example/example-dev.html
```diff
@@ -197,6 +197,7 @@
       data: {
         blocks: [
           {
+            id: "1bfa3a88-a6e3-4a13-8553-d8c85bf2497b",
```
neSpecc: it is not a nanoid(10)

src/components/modules/blockManager.ts
```diff
@@ -223,11 +223,13 @@ export default class BlockManager extends Module {
   public composeBlock({
     tool: name,
     data = {},
+    id = _.generateBlockId(),
```
neSpecc: not descirbed

src/components/block/index.ts
```diff
@@ -206,20 +216,23 @@ export default class Block {
 
   /**
    * @param {object} options - block constructor options
+   * @param {string} options.id - Tool's unique id
```
neSpecc: wrong description

src/components/core.ts
```diff
@@ -182,6 +182,7 @@ export default class Core {
      * @type {{type: (*), data: {text: null}}}
      */
     const defaultBlockData = {
+      id: _.generateBlockId(),
```
neSpecc: Id is not required param, so I think it should be removed from here. It will be generated in the `composeBlock` method

src/components/modules/blockManager.ts
```diff
@@ -255,13 +257,15 @@ export default class BlockManager extends Module {
    * @returns {Block}
    */
   public insert({
+    id = _.generateBlockId(),
```
neSpecc: I think, `id` should be generated in the single place — in the `composeBlock`

src/components/modules/renderer.ts
```diff
@@ -23,12 +23,14 @@ export default class Renderer extends Module {
    *
    * blocks: [
    *   {
+   *     id   : 'eab2b1cd-a3d3-48d1-aadb-bd821d8ec871',
```
neSpecc: it is not a nanoid(10)

src/components/utils.ts
```diff
@@ -607,6 +608,15 @@ export function getValidUrl(url: string): string {
   }
 }
 
+/**
+ * Create a block id
+ *
+ * @returns {string}
+ */
+export function generateBlockId(): string {
+  return nanoid();
```
neSpecc: let's use nanoid(10)

test/cypress/tests/copy-paste.spec.ts
```diff
@@ -126,7 +126,7 @@ describe('Copy pasting from Editor', () => {
         });
 
       cy.get('[data-cy=editorjs]')
-        .get('img')
+        .get('img', { timeout: 10000 })
```
neSpecc, line 130: magic number

neSpecc (changes requested):

test/cypress/tests/copy-paste.spec.ts
jggc: I'll let @gohabereg manage this one

src/components/modules/blockManager.ts
jggc: Yes I completely agree, did this change (and the other similar one) and it even fixed the id being re generated when moving the block up/down

jggc:
@gohabereg @neSpecc  I'll be mostly unavailable to work on this PR for the next couple days, don't hesitate to push more improvements until it's mergeable.

If I see more review comments I'll try my best to make the fixes quickly still but I'll try to take the weekend off.

test/cypress/tests/block-ids.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+        data.blocks.forEach(block => {
+          expect(typeof block.id).to.eq('string');
+        });
+      });
+  });
+
+  it('should preserve passed ids', () => {
+    const blocks = [
+      {
+        id: nanoid(),
```
neSpecc, line 55: We use `nanoid (10)`. What should happen when user pass `id` not in an our format? We should show error or preserve them? I think, we need to add a test for that case.

src/components/block/index.ts
```diff
@@ -23,6 +23,11 @@ import ToolsCollection from '../tools/collection';
  * Interface describes Block class constructor argument
  */
 interface BlockConstructorOptions {
+  /**
+   * Block's id
+   */
+  id: string;
```
neSpecc: I think its better to make it optional and generate block ids here instead of BlockManager.

test/cypress/tests/block-ids.spec.ts
gohabereg: I don't think so. User can use any format for ids on the backend side, even just numbers. We should just preserve everything user passes

neSpecc (approved):

test/cypress/tests/block-ids.spec.ts
neSpecc: That was a question.

neSpecc (approved):

sherqo:
i am using editorjs and for some reason i want `uuid` instead of `nanoid`. is there a solution to do this?  because i'm forking the repo and changing it internally.

`src/components/utils.ts`:
``` diff
-  export function generateBlockId(): string {
-   const idLen = 10;
- 
-   return nanoid(idLen);
- }

+  export function generateBlockId(): string {
+   return crypto.randomUUID();
+ }
```

keremnymn:
Any chance we can intervene in the ID creating process? I need to create a unique ID which is not present in an external array. I see that the previous question has a similar request.

PR #1678, opened 2021-05-08
onChange improvements

gohabereg:


src/components/modules/blockManager.ts
```diff
@@ -290,6 +291,8 @@ export default class BlockManager extends Module {
 
     this._blocks.insert(newIndex, block, replace);
 
+    this.blockDidMutated(block);
```
neSpecc, line 297: add a comment please

example/example-dev.html
```diff
@@ -319,8 +319,8 @@
       onReady: function(){
         saveButton.click();
       },
-      onChange: function() {
-        console.log('something changed');
+      onChange: function(api, block) {
+        console.log('something changed', block);
```
neSpecc, line 323: add this change to the `example.html` too

jggc:
We tested this branch in our project and it works nicely, getting the changed block or null if deleted. :+1:

neSpecc (approved):

khaydarov (approved):

PR #1681, opened 2021-05-12
Add tunes to block insert API

williamhakim10:
This PR allows users to insert blocks via the API, and tune them at the same time. This is a feature we've struggled without as there is currently no way to programmatically tune a block either.


williamhakim10:
Gentle bump?

PR #1703, opened 2021-06-02
Config option `ignoreDeleteFullBlockByKey`

burnes:
to disable block deletion via backspace.

Background:

In our project we have some blocks which are not inline editable, but have a modal editor. All those plugins do not include any input fields. When users hit the backspace key, the whole block will be deleted without being able to undo. With this customizable setting we want to disable deleting those blocks, so the users will have to explicitly click the X button in the tune settings to remove the block.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### Unreleased
+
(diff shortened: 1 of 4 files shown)


LarsWiegers:
Hey Guys, would love to use this in my project, can we get an update on this pull request?

PR #1722, opened 2021-07-13
Tunes improvements for inline actions

gohabereg:


example/example-dev.html
```diff
@@ -81,6 +81,7 @@
 
   <script src="./tools/marker/dist/bundle.js"></script><!-- Marker -->
   <script src="./tools/inline-code/dist/bundle.js"></script><!-- Inline Code -->
+  <script src="../../footnotes-tune/dist/bundle.js"></script><!-- Inline Code -->
```
neSpecc: wrong path

example/example-dev.html
```diff
@@ -205,10 +212,15 @@
...
-            type : 'paragraph',
-            id: "b6ji-DvaKb",
-            data : {
-              text : 'Hey. Meet the new Editor. On this page you can see it in action — try to edit this text. Source code of the page contains the example of connection and configuration.'
+            "id": "b6ji-DvaKb",
+            "type": "paragraph",
+            "data": {
+              "text": "Hey. Meet the new Editor. On this page you can see it in action<sup>1</sup> — try to edit this text. Source code of the page contains the example of connection and configuration."
+            },
+            "tunes": {
```
neSpecc: I would suggest not complicating the existed example. Let's move it to the separate example page, `example-tunes`, for example

src/components/block/index.ts
```diff
@@ -201,7 +203,18 @@ export default class Block extends EventsDispatcher<BlockEvents> {
   /**
    * Is fired when DOM mutation has been happened
    */
-  private didMutated = _.debounce((): void => {
+  private didMutated = _.debounce((mutations: MutationRecord[]): void => {
+
+    /**
+     * In case fake cursor is added or removed, do not trigger didMutated event
+     */
+    if (mutations.some(({ addedNodes = [], removedNodes }) => {
```
neSpecc: pls move it to the variable

src/components/block/index.ts
```diff
@@ -448,8 +461,20 @@ export default class Block extends EventsDispatcher<BlockEvents> {
   public set selected(state: boolean) {
     if (state) {
       this.holder.classList.add(Block.CSS.selected);
+
+      const range = SelectionUtils.range;
```
neSpecc: move it to a separate method. Should it be inside the Block class? Maybe it is related to the Caret util

src/components/block/index.ts
```diff
@@ -630,7 +655,7 @@ export default class Block extends EventsDispatcher<BlockEvents> {
     const defaultTunesElement = document.createDocumentFragment();
 
     this.tunesInstances.forEach((tune) => {
-      $.append(tunesElement, tune.render());
+      $.append(tunesElement, tune.render(SelectionObserver.getSelectedRange()));
```
neSpecc: do not pass range to the Block Tunes constructor, because the Range is not related to the Block Tunes

src/components/modules/ui.ts
```diff
@@ -340,7 +340,7 @@ export default class UI extends Module<UINodes> {
       this.documentKeydown(event);
     }, true);
 
-    this.readOnlyMutableListeners.on(document, 'click', (event: MouseEvent) => {
+    this.readOnlyMutableListeners.on(document, 'mousedown', (event: MouseEvent) => {
```
neSpecc, line 343: is it break something?

src/components/modules/ui.ts
```diff
@@ -591,9 +591,7 @@ export default class UI extends Module<UINodes> {
     /**
      * Clear Selection if user clicked somewhere
      */
-    if (!this.Editor.CrossBlockSelection.isCrossBlockSelectionStarted) {
-      this.Editor.BlockSelection.clearSelection(event);
-    }
+    this.Editor.BlockSelection.clearSelection(event);
```
neSpecc, line 594: why the statement was removed?

src/components/modules/ui.ts
```diff
@@ -754,18 +752,46 @@ export default class UI extends Module<UINodes> {
...
      */
-    if (!focusedElement || !focusedElement.closest(`.${Block.CSS.content}`)) {
+    if (!focusedElement) {
+
+      /**
+       * If there is no selected range, close inline toolbar
+       *
+       * @todo Make this method more straightforward
+       */
+      if (!Selection.range) {
```
neSpecc, line 763: as i can see, it already works on the release branch — when you collapse the sleection. the Inline Toolbar is closing. It is covered by `tryToShow` method. So please explain this change.

src/components/selection.ts
```diff
@@ -34,6 +34,34 @@ interface Document {
  * @typedef {SelectionUtils} SelectionUtils
  */
 export default class SelectionUtils {
+  /**
+   * Selection instances
+   *
+   * @todo Check if this is still relevant
+   */
+  public instance: Selection = null;
```
neSpecc, line 42: why the instance went public?

src/components/tools/block.ts
```diff
@@ -160,6 +160,10 @@ export default class BlockTool extends BaseTool<IBlockTool> {
       .from(this.inlineTools.values())
       .forEach(tool => Object.assign(baseConfig, tool.sanitizeConfig));
 
+    Array
```
neSpecc, line 163: pls update the jsdoc

src/components/utils/selectionObserver.ts
```diff
@@ -0,0 +1,152 @@
...
+  Start = 'fake-pointer-start',
+  End = 'fake-pointer-end'
+}
+
+/**
+ * SelectionObserver class to work with selection events
+ *
+ * @todo Investigate fake pointer concept
+ */
+class SelectionObserver extends EventsDispatcher<SelectionEvents> {
```
neSpecc: remove the class if it is not necessary

types/block-tunes/block-tune.d.ts
```diff
@@ -41,6 +41,8 @@ export interface BlockTuneConstructable {
    */
   isTune: boolean;
 
+  sanitize?: SanitizerConfig;
```
neSpecc, line 47: add jsdoc and update the Block Tunes API docs

neSpecc (changes requested):

example/example-dev.html
gohabereg: Should be removed before merge

example/example-dev.html
gohabereg: I would remove it completely and maybe add example to the tunes repository

src/components/block/index.ts
gohabereg: Would be easier to manage this inside a block because we should remove it when selection is gone

src/components/modules/ui.ts
gohabereg: No, I went thru documentClicked method and there is no dependencies on click phase

src/components/modules/ui.ts
gohabereg: Because I changed `click` to `mousedown` above. On mouse down we are sure that selection should be remove. On click we are not because it might be added on the same click

src/components/selection.ts
gohabereg: It was public... Just moved it to the top cause linter was angry

src/components/modules/ui.ts
gohabereg: `tryToShow` is called in this method on line 809. You've added this if statement with return in it, so if there is no `focusedElement` `tryToShow` is not called that's why we need additional check here

src/components/block/index.ts
gohabereg: Agree would be better to place it to Caret util but it is not util yet and Block class doesn't have access to it

docs/block-tunes.md
```diff
@@ -145,7 +145,24 @@ No return value
...
+### static get sanitize()
+
+If your Tune inserts any HTML markup into Block's content you need to provide sanitize configuration, so your HTML is not trimmed on save.
+
+Please see more information at [sanitizer page](sanitizer.md).
+
+
+```javascript
+class Tune {
+  static get reset() {
```
neSpecc:
```suggestion
  static get sanitize() {
```

neSpecc (approved):

talyguryn (approved):

Issue #1740, opened 2021-08-05
[Bug] api.blocks.render empty blocks cause "Cannot read property 'holder' of undefined"


cdllqos:
Describe a bug.

Steps to reproduce:
1. When editor is prepared
2. Call `api.blocks.render({blocks:[]})`
3. Click editor，throw `Cannot read property 'holder' of undefined`

Expected behavior:  editor works no error

Screenshots:
![image](https://user-images.githubusercontent.com/25027791/128332412-c5864547-3903-4f7a-9956-94025ab37644.png)

Device, Browser, OS:

Editor.js version: 2.22.2

Plugins you use with their versions: no plugins

christopher-caldwell:
YES! I thought I was insane, the same thing is happening to me with v2.22.2

anstapol:
Same here...

eyalcohen4:
Same here, does someone found a workaround? Need to access the holder of a tool

PR #1741, opened 2021-08-05
fix: #1740 fix holder undefined error

cdllqos:
Related issue：
https://github.com/codex-team/editor.js/issues/1740

There is no currentBlock after calling `api.blocks.render({blocks:[]})` with empty blocks, so we should protect `currentBlock` in this situation like this:
```typescript
  // file: src/components/modules/api/blocks.ts:163
  /**
   * Clear Editor's area
   */
  public clear(): void {
    this.Editor.BlockManager.clear(true);
    this.Editor.InlineToolbar.close();
  }
```
This situation could happen in  package `editorjs-undo`, reference is [here](https://github.com/kommitters/editorjs-undo/blob/master/src/index.js#L142)

### DIFF
--- a/src/components/modules/api/blocks.ts
+++ b/src/components/modules/api/blocks.ts
@@ -170,7 +170,12 @@ export default class BlocksAPI extends Module {
    * @param {OutputData} data — Saved Editor data
    */
   public render(data: OutputData): Promise<void> {
-    this.Editor.BlockManager.clear();
+    const needToAddDefaultBlock = !data.blocks.length
+    this.Editor.BlockManager.clear(needToAddDefaultBlock);
+
+    if(needToAddDefaultBlock) {
+      return Promise.resolve()
+    }
 
     return this.Editor.Renderer.render(data.blocks);
   }
--- a/src/components/modules/toolbar/index.ts
+++ b/src/components/modules/toolbar/index.ts
@@ -190,7 +190,7 @@ export default class Toolbar extends Module<ToolbarNodes> {
       this.Editor.BlockSettings.close();
     }
 
-    const currentBlock = this.Editor.BlockManager.currentBlock.holder;
+    const currentBlock = this.Editor.BlockManager.currentBlock?.holder;
 
     /**
      * If no one Block selected as a Current


petrsiegl:
I had the same thing happen to me with `editorjs-undo`, but this affects any render with empty blocks as specified. Seems like a good fix to me.

seancheung:
This happened to me when I tried to load a previously saved empty doc(`{blocks:[]}`)

petrsiegl:
For a possible quick fix (as this PR has been here for a while, and if you don't want to fork and fix) simply set default block yourself for `blocks == []` situation:
```js
      blocks = [{
        "data": {"text": ""},
        "id": "WhateverXX",
        "type": "paragraph",
      }]

```

JointVentureJones:
> For a possible quick fix (as this PR has been here for a while, and if you don't want to fork and fix) simply set default block yourself for `blocks == []` situation:
> 
> ```js
>       blocks = [{
>         "data": {"text": ""},
>         "id": "WhateverXX",
>         "type": "paragraph",
>       }]
> ```

How to do this?

petrsiegl:
> > For a possible quick fix (as this PR has been here for a while, and if you don't want to fork and fix) simply set default block yourself for `blocks == []` situation:
> > ```js
> >       blocks = [{
> >         "data": {"text": ""},
> >         "id": "WhateverXX",
> >         "type": "paragraph",
> >       }]
> > ```
> 
> How to do this?

Everytime you need to render blocks check if they are not empty. If they are empty use the mention code instead. Something like:

```js
// const editor = new EditorJS(...)
function renderBlocks(blocks) {
    if (blocks.length === 0) {
        blocks = [{
            "data": {"text": ""},
            "id": "WhateverXX",
            "type": "paragraph",
        }]
    }
    editor.render({
        blocks
    });
}
```

or4:
after a lot repeats I found solution, need two empty blocks!
```
export const emptyBlock = {
  time: new Date().getTime(),
  blocks: [
    {
      data: {text: ''},
      id: 'WhateverXX',
      type: 'paragraph'
    },
    {
      data: {text: ''},
      id: 'WhateverXX',
      type: 'paragraph'
    }
  ],
  version: '2.24.3'
}
instance.current.render({blocks: emptyBlock.blocks})
```

JorrandeWit (approved):

PR #1804, opened 2021-11-02
repair table node data when paste

kooky126:
when I add the ability of paste from word table or html table for https://github.com/editor-js/table，I found event.tag of table has lost tr/td data

in src/components/modules/paste.ts

the innerHTML of table node is tbody，and it's innerHTML not include tr/td, 

my method is keep the innerHTML of table（no clean）

new table plugin is in

https://github.com/kooky126/table

it's working


JoeWoodward:
This is the wrong way to fix this issue. I have opened another PR that solves the issue but also retains the sanitization logic which also solves another issue that the root node isn't sanitized when using innerHTML. #1807

Issue #1813, opened 2021-11-16
💡Get data from the BlockAPI


RobinDev:
1. Describe a problem.
From a tune plugin, I would like to be able to get data from the current blockI (so in the block API).

2. Describe the solution you'd like. Mockups are welcome.

Just add the method (see PR).

PR #1814, opened 2021-11-16
Add blockAPI.data

RobinDev:
See issue #1813


neSpecc:
Why not just use `.save()` ?

RobinDev:
You are right

PR #1815, opened 2021-11-20
feat(toolbar): toolbar refactored and ui improved

neSpecc:
This PR introduces several updates of the Toolbar code and UI.

This is the first part of updates due to upcoming feature called "Unified menu". 

# List of changes

1. Block Tunes toggler moved to the left

image

2. Block Actions (BT toggler + Plus Button) will appear on block hovering instead of click

- The UI module triggers `block-hovered` event
- Toolbar uses `block-hovered` for appearing
- `currentBlock` setter added to the BlockManager. We use it to set the Current Block on clicks on the BT Toggler or the Plus Button (and some other cases as well)

3. Block Tunes toggler icon and Plus button icon updated
4. Dev Example Page / menu with helpful buttons added to the bottom of the screen

image

5. Dev Example Page / 'dark' theme added. Now we can code at night more comfortably.

image

6. Rectangle Selection (optimization):  the throttling added to the `mousemove` and `scroll` handlers
7. Rectangle Selection (fix): the first click after RS was not clear selection state. Now does.
8. Blocks API: the `getBlockIndex()` method added
9. Blocks API: toolbar moving logic removed from `blocks.move()` and `blocks.swap()` methods. Instead, you should use Toolbar API (it was used by MoveUp and MoveDown tunes, they were updated).
10. Blocks API: the `insert()` method now has the `replace: boolean` parameter
11. Blocks API:  the `insert()` method now returns the inserted `Block API`
12. Listeners API: the `on()` method now returns the listener id. 
13. Listeners API: the new `offById()` method added
14. Toolbox became a standalone class instead of a Module. It can be accessed only through the Toolbar module. _`Block Settings` will be refactored the same way in the next updates._
15. The new `UiApi` section was added. It allows accessing some editor UI nodes and methods. 
16. Git submodules removed from the CI flow.

---

Related https://github.com/editor-js/header/pull/72

### DIFF
--- a/src/components/block-tunes/block-tune-move-down.ts
+++ b/src/components/block-tunes/block-tune-move-down.ts
@@ -109,7 +109,6 @@ export default class MoveDownTune implements BlockTune {
     /** Change blocks positions */
     this.api.blocks.move(currentBlockIndex + 1);
 
-    this.api.toolbar.open();
     this.api.toolbar.toggleBlockSettings(true);
 
     /** Hide the Tooltip */
--- a/src/components/block-tunes/block-tune-move-down.ts
+++ b/src/components/block-tunes/block-tune-move-down.ts
@@ -109,6 +109,9 @@ export default class MoveDownTune implements BlockTune {
     /** Change blocks positions */
     this.api.blocks.move(currentBlockIndex + 1);
 
+    this.api.toolbar.open();
+    this.api.toolbar.toggleBlockSettings(true);
+
     /** Hide the Tooltip */
     this.api.tooltip.hide();
   }
--- a/src/components/modules/toolbar/index.ts
+++ b/src/components/modules/toolbar/index.ts
@@ -482,6 +482,9 @@ export default class Toolbar extends Module<ToolbarNodes> {
       this.tooltip.hide(true);
     }, true);
 
+    /**
+     * Subscribe to the 'block-hovered' event
+     */
     this.eventsDispatcher.on(this.Editor.UI.events.blockHovered, (data: {block: Block}) => {
       /**
        * Do not move toolbar if Block Settings or Toolbox opened
--- a/src/components/modules/toolbar/index.ts
+++ b/src/components/modules/toolbar/index.ts
@@ -9,7 +9,6 @@ import { EditorConfig } from '../../../../types';
 import Block from '../../block';
 
 /**
- * @todo clear cross block selection on click on block
  * @todo Tab on empty block should insert block in place of hoveredBlock (not where caret is set)
  * @todo Debug Thin mode
  * @todo Debug Mobile mode
--- a/src/components/modules/toolbar/index.ts
+++ b/src/components/modules/toolbar/index.ts
@@ -12,6 +12,8 @@ import Block from '../../block';
  * @todo remove block margins
  * @todo clear cross block selection on click on block
  * @todo Tab on empty block should insert block in place of hoveredBlock (not where caret is set)
+ * @todo Debug Thin mode
+ * @todo Debug Mobile mode
  *
  * @todo TESTCASE - show toggler after opening and closing the Inline Toolbar
  * @todo TESTCASE - Click outside Editor holder should close Toolbar and Clear Focused blocks
--- a/src/components/modules/toolbar/index.ts
+++ b/src/components/modules/toolbar/index.ts
@@ -5,8 +5,9 @@ import I18n from '../../i18n';
 import { I18nInternalNS } from '../../i18n/namespace-internal';
 import Tooltip from '../../utils/tooltip';
 import { ModuleConfig } from '../../../types-internal/module-config';
-import { EditorConfig } from '../../../../types';
+import { BlockAPI, EditorConfig } from '../../../../types';
 import Block from '../../block';
+import Toolbox from '../../ui/toolbox';
 
 /**
  * @todo Tab on empty block should insert block in place of hoveredBlock (not where caret is set)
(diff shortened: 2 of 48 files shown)


src/components/block-tunes/block-tune-move-down.ts
```diff
@@ -109,6 +109,9 @@ export default class MoveDownTune implements BlockTune {
     /** Change blocks positions */
```
neSpecc, line 109: Toolbar moving logic removed from the `api.blocks.move()` method, so we do it manually.

src/components/modules/ui.ts
```diff
@@ -691,12 +761,21 @@ export default class UI extends Module<UINodes> {
...
       /**
        * If there is cross block selection started, target will be equal to redactor so we need additional check
        */
-      !BlockSelection.anyBlockSelected;
+      !BlockSelection.anyBlockSelected &&
+
+      /**
+       * Prevent caret jumping (to last block) when clicking between blocks
+       */
+      lastBlockBottomCoord < clickedCoord;
```
neSpecc, line 778: have improved the `isClickedBottom` statement: 

some blocks can have margins, so when a user clicks on such margins between `.ce-block`, caret was jumping to the bottom of the document. For now, it checks the exact coordinate of the last block bottom.

src/components/block-tunes/block-tune-move-down.ts
```diff
@@ -109,6 +109,9 @@ export default class MoveDownTune implements BlockTune {
...
     this.api.blocks.move(currentBlockIndex + 1);
 
+    this.api.toolbar.open();
+    this.api.toolbar.toggleBlockSettings(true);
```
gohabereg: Looks not really convenient, maybe we can at least use just one method?

src/components/modules/toolbar/index.ts
```diff
@@ -172,54 +227,80 @@ export default class Toolbar extends Module<ToolbarNodes> {
...
      */
     this.nodes.wrapper.style.transform = `translate3D(0, ${Math.floor(toolbarY)}px, 0)`;
+
+    /**
+     * Plus Button should be shown only for __empty__ __default__ block
+     */
+    if (block.tool.isDefault && block.isEmpty) {
+      this.plusButton.show();
+    } else {
+      this.plusButton.hide();
```
gohabereg, line 288: Might be a good idea to move plus button into toolbox as well

src/components/modules/ui.ts
```diff
@@ -356,6 +373,48 @@ export default class UI extends Module<UINodes> {
...
+  /**
+   * Listen redactor mousemove to emit 'block-hovered' event
+   */
+  private watchBlockHoveredEvents(): void {
+    /**
+     * Used to not to emit the same block multiple times to the 'block-hovered' event on every mousemove
+     */
+    let blockHoveredEmitted;
+
+    this.readOnlyMutableListeners.on(this.nodes.redactor, 'mousemove', _.throttle((event: MouseEvent | TouchEvent) => {
```
gohabereg, line 392: Why not `mouseenter` and `mousemove`. Also would be great to check on pointer devices and mobile devices with connected touchpad/mouse

src/components/ui/toolbox.ts
```diff
@@ -90,28 +108,51 @@ export default class Toolbox extends Module<ToolboxNodes> {
...
+   * @param options.onOpen - opening callback
+   * @param options.onClose - closing callback
+   * @param options.onBlockAdded - block adding callback
+   */
+  constructor({ api, tools, shortcutsScopeElement, onOpen, onClose, onBlockAdded }) {
+    this.api = api;
+    this.tools = tools;
+    this.onOpen = onOpen;
+    this.onClose = onClose;
+    this.onBlockAdded = onBlockAdded;
```
gohabereg: In case API would grow, might be better to just extend EventEmitter interface and provide enum of possible events

```
toolbox.on(ToolboxEvents.Close, () => {});
toolbox.on(ToolboxEvents.Open, () => {});
```

Using typescript we can strictly type that

src/components/ui/toolbox.ts
```diff
@@ -337,34 +387,42 @@ export default class Toolbox extends Module<ToolboxNodes> {
...
+     */
+    const index = currentBlock.isEmpty ? currentBlockIndex : currentBlockIndex + 1;
+
+    const newBlock = this.api.blocks.insert(
+      toolName,
+      undefined,
+      undefined,
+      index,
+      undefined,
+      currentBlock.isEmpty
```
gohabereg, line 404: We should go with object here

src/components/modules/toolbar/index.ts
```diff
@@ -335,11 +406,55 @@ export default class Toolbar extends Module<ToolbarNodes> {
...
+   * Creates the Toolbox instance and return it's rendered element
+   */
+  private makeToolbox(): Element {
+    /**
+     * Make the Toolbox
+     */
+    this.toolboxInstance = new Toolbox({
+      api: this.Editor.API.methods,
+      tools: this.Editor.Tools.blockTools,
+      shortcutsScopeElement: this.Editor.UI.nodes.redactor,
```
gohabereg: Might be better to provide some nodes from Editor API
```
editorAPI.nodes.redactor

// or

editorAPI.ui.redactor
```

gohabereg (changes requested):

src/components/modules/toolbar/index.ts
neSpecc: I don't think that it would be better. Toolbox is the list of buttons. It is about to become vertical. And it can be appeared in several ways, not only by Plus Button. And the Plus Button is just a part of Toolbar UI.

src/components/modules/ui.ts
neSpecc: The `mouseenter` would have to be bound to every Block, and there can be a large amount of them. I think it's better to use the single event on the `redactor` node.

> Also would be great to check on pointer devices and mobile devices with connected touchpad/mouse

I will test the RC version in [CodeX Docs](https://github.com/codex-team/codex.docs), so I will try to use various hardware.

src/components/modules/toolbar/index.ts
neSpecc: done

src/components/ui/toolbox.ts
neSpecc: I've decided not to change the interface of the existed API method. Should I create an overload for that?

src/components/ui/toolbox.ts
neSpecc: done

src/components/block-tunes/block-tune-move-down.ts
neSpecc: done

gohabereg (approved):

Issue #1503, opened 2021-01-16
[Bug] Uncaught TypeError: Cannot read property 'updateCurrentInput' of undefined


m-alzam:
When I click on a block of any EditorJS instance when there are two editors in the same page, I get the following error:

```js
Uncaught TypeError: Cannot read property 'updateCurrentInput' of undefined
```

This is where the error is produced:

https://github.com/codex-team/editor.js/blob/43032ebaac6a8b9a67c9bf2ada5adc0f8592d546/src/components/modules/blockManager.ts#L566-L593

Steps to reproduce:
1. Create two editors in the same page
2. Click on a block

Expected behavior: No errors reported to the console.

Device, Browser, OS: Chromium 87

Editor.js version: 2.19.1

samarjaffal:
Hi, 
I have the same problem. 
Currently using the editor on a Modal developed on React. When you open the modal again the error appears on the console.

fatheyabdelslam:
i got the same thing

davidyim:
I have the same problem, using Editor.js on React

umutbozdag:
it's same for me too when i am using more than one editor instances in the same page

aungmyohlaing:
I got the same.  Using editor on modal with Vuejs

pasloc:
I got the same error using the editor in nuxtjs. I have a toggle button which hides the editor. After toggling once the error occurs.

stevenarrol:
Hey guys,
Had the same issue using React and was digging around for days trying to fix it. Figured out a way that seems to work.

Make sure you are initializing the editor using Class-based component, and then call editor.destroy() on unmount

`componentWillUnmount() {
        this.editor.destroy()
    }`

pasloc:
`this.editor.destroy()` works for me too in Vue.js

jeffbocala:
Seems editor.js does not support multiple editors in one page?

hata6502:
This bug occurs in the following demo page. 
https://hata6502.github.io/editorjs-inline/

<img width="1440" alt="スクリーンショット 2021-01-25 19 57 37" src="https://user-images.githubusercontent.com/7702653/105697229-a7f37880-5f47-11eb-9bae-d7d0614b3ef3.png">

Editor.js supported multiple instances in v.2.18.0. 
But it seems broken in recently version.

hata6502:
[...]

PR #1850, opened 2021-12-22
fix: Multiple editor.js instances throws an error

JoseLAmador:
## Issue

When I try to use a multiple instances of editor.js in the same page I get the next error:

Captura de Pantalla 2021-12-21 a la(s) 14 30 07

## Solution

The method updateCurrentInput fails because the index not exists in the array.
I changed the method with a prev validation for only update the current block if the index exists

Maybe will be related #1503

### DIFF
--- a/src/components/modules/blockManager.ts
+++ b/src/components/modules/blockManager.ts
@@ -681,6 +681,18 @@ export default class BlockManager extends Module {
     /**
      * Update current block active input
      */
+
+    /**
+     * Fix: when we have multiple instances and
+     * currentBlockIndex is -1.
+     * 
+     * The method updateCurrentInput fails because the index not exists
+     * in the array
+     */
+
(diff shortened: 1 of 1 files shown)


src/components/modules/blockManager.ts
```diff
@@ -681,6 +681,18 @@ export default class BlockManager extends Module {
     /**
      * Update current block active input
      */
+
+    /**
+     * Fix: when we have multiple instances and
+     * currentBlockIndex is -1.
```
neSpecc, line 687: Maybe `setCurrentBlockByChildNode` should not be triggered in an inactive editor at all?

JoseLAmador: sure, i think it's the best approach. If the user clicks outside the editor, shouldn't trigger `setCurrentBlockByChildNode ` method

Issue #1853, opened 2021-12-31
[Bug] Input type email errors on selectionEnd


PetroSilenius:
Adding a custom email input block results in an error because email input doesn't support `selectionEnd` which is used to clear off selection. The error comes from `src/components/modules/caret.ts:57`.

Steps to reproduce:
1. Create a custom input block with type `email`
2. Add that block to the editor instance from the toolbar
3. Get the error
`InvalidStateError: Failed to set the 'selectionEnd' property on 'HTMLInputElement': The input element's type ('email') does not support selection.`

Expected behavior:
The selectionEnd doesn't get called and thus no error is shown

Screenshots:
![Screenshot of error](https://i.imgur.com/BcPruHE.png)

Device, Browser, OS:
Macbook Pro 16" 2019 2,6GHz i7
MacOS Big Sur version 11.6
Chrome version 96.0

Editor.js version:
2.22.2

Plugins you use with their versions:
@editorjs/paragraph
@editorjs/list
@editorjs/header

PR #1854, opened 2021-12-31
Fix: Check node supports selectionEnd before calling it

PetroSilenius:
## Issue
Email input type doesn't support `selectionEnd` which is used to clear off selection and errors on that call. More details can be found in #1853. 

Screenshot of error

## Solution
Check if node supports `selectionEnd` before calling it


robonetphy:
Hi @PetroSilenius, thanks for pr but the issue is not fully resolved as I can see.
we are using `selectionEnd` as well as  `selectionStart` at multiple locations which will eventually cause the same issue.

can you add a function to check if the input is supported and do nothing if not?

it's due to the following reason
image

PetroSilenius:
Hey and thanks for taking a look at this!
You're right that this PR doesn't fully solve the issue. I'll go ahead and a function for checking if a input supports selection and use that in places where it's needed

PetroSilenius:
@robonetphy updated the PR to add a function for checking whether given node supports selection or not. Also added a check with that function in places where I saw either selectionStart or selectionEnd being used

What do you think of these changes?

src/components/dom.ts
```diff
@@ -634,4 +634,28 @@ export default class Dom {
...
+    const selectionSupportedTypes = [
+      'text',
+      'search',
+      'tel',
+      'url',
+      'password',
+    ];
+
+    if (Dom.isNativeInput(target)) {
+      return selectionSupportedTypes.includes(target.type) || target.tagName === 'textarea';
```
robonetphy:
```suggestion
      return selectionSupportedTypes.includes(target.type) || target.tagName === 'TEXTAREA';
```

src/components/dom.ts
```diff
@@ -634,4 +634,28 @@ export default class Dom {
...
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  public static supportsSelection(target: any): boolean {
+    const selectionSupportedTypes = [
+      'text',
+      'search',
+      'tel',
+      'url',
+      'password',
+    ];
```
robonetphy, line 653:
```suggestion
      'password',
      'textarea',
    ];
```

robonetphy (changes requested):
@PetroSilenius Pls do these changes update the `changelog.md`  and we are ready

PetroSilenius:
> @PetroSilenius Pls do these changes update the `changelog.md` and we are ready

Sure! Just double-checking that has it been decided not to support textarea in editor.js? They reason I included a check for it is that it also supports `selectionStart` and `selectionEnd` according to [mdn docs](https://developer.mozilla.org/en-US/docs/Web/API/HTMLTextAreaElement#:~:text=direction%20is%20unknown.-,selectionEnd,-unsigned%20long%3A%20Returns)

robonetphy:
> > @PetroSilenius Pls do these changes update the `changelog.md` and we are ready
> 
> Sure! Just double-checking that has it been decided not to support textarea in editor.js? They reason I included a check for it is that it also supports `selectionStart` and `selectionEnd` according to [mdn docs](https://developer.mozilla.org/en-US/docs/Web/API/HTMLTextAreaElement#:~:text=direction%20is%20unknown.-,selectionEnd,-unsigned%20long%3A%20Returns)

@PetroSilenius you are correct `textarea` also support both `selectionStart` and `selectionEnd` but comparison of `tagName` must be with the capital letter `TEXTAREA`  as [mdn docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/tagName) or you can also add inside `supported type` as `textarea` both the ways are acceptable

PetroSilenius:
> @PetroSilenius you are correct `textarea` also support both `selectionStart` and `selectionEnd` but comparison of `tagName` must be with the capital letter `TEXTAREA` as [mdn docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/tagName) or you can also add inside `supported type` as `textarea` both the ways are acceptable

Right, thanks for catching the lowercase mistake! That's now fixed and updated the changelog as well as package version

robonetphy (approved):

robonetphy (approved):

src/components/dom.ts
```diff
@@ -634,4 +634,28 @@ export default class Dom {
...
+
+  /**
+   * Returns true if element supports selection
+   *
+   * @param {*} target - HTML element or string
+   *
+   * @returns {boolean}
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  public static supportsSelection(target: any): boolean {
```
neSpecc: we already have the `canSetCaret` method. What is the difference between them?

neSpecc: the logic looks similar except the case with contenteditable

robonetphy: @neSpecc So what should we any suggestions?

PetroSilenius: HTML inputs have 22 different types, according to [mdn docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#input_types). `canSetCaret` checks for 8 of them which leaves 14. Out of those 14 only 5 support selection so the methods check for quite a different set of input types.

I totally get the idea of reusing code but I'm not quite sure if that's feasible here. Also separating them creates a bit more easily understandable code

neSpecc: I think the reason of the `canSetCaret` existance is the same as that method. So maybe we could improve `canSetCart` and update its logic to cover more cases?

PetroSilenius: Sure that might work, issue is that with current implementation `canSetCaret` return true for `contenteditable` but `contenteditable` doesn't support selection. I now basically changed `canSetCaret` implementation to the one I made for `supportsSelection` and thus dropped support for `contenteditable`.

It seems to work quite well as far as I checked since previously the `isNativeInput` was used for checking selection stuff and it doesn't support `contenteditable`. Would be glad to hear thoughts on this!

src/components/dom.ts
```diff
@@ -349,27 +349,20 @@ export default class Dom {
...
-  public static canSetCaret(target: HTMLElement): boolean {
-    let result = true;
+  public static canSetCaret(target: HTMLElement | Node): boolean {
+    const selectionSupportedTypes = [
+      'text',
+      'search',
+      'tel',
+      'url',
+      'password',
+    ];
```
neSpecc: Looks like some types are missed:

- email
- number
- month
- week

And maybe these, not sure:

- date
- datetime-local
- time

PetroSilenius: Only the five defined types support selection. Here's [HTML specs](https://html.spec.whatwg.org/multipage/input.html#concept-input-apply:~:text=%C2%B7-,selectionStart,-%C2%B7) and [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/setSelectionRange#:~:text=text%2C%20search%2C%20URL%2C%20tel%20and%20password) for it

src/components/dom.ts
```diff
@@ -349,27 +349,20 @@ export default class Dom {
    *
    * @returns {boolean}
    */
-  public static canSetCaret(target: HTMLElement): boolean {
-    let result = true;
+  public static canSetCaret(target: HTMLElement | Node): boolean {
+    const selectionSupportedTypes = [
```
neSpecc:
```suggestion
    /**
     * @see https://html.spec.whatwg.org/multipage/input.html#concept-input-apply
     * @see https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/setSelectionRange
     */
    const selectionSupportedTypes = [
```

src/components/dom.ts
neSpecc: Got it, thanks. Let's add this information to the jsdoc, please

PetroSilenius: Good idea, done now!

PetroSilenius:
If there something I could do still related to this PR, or are these change ready to be merged?😄
Last [tests/edge](https://github.com/codex-team/editor.js/actions/runs/2973528843/jobs/4763247621) run failed because of connetion issues when fetching packages. I think a simple rerun would fix it but don't have access to do that

TatianaFomina:
Hi @PetroSilenius! 

We are going to include this PR in one of the upcoming releases. At the moment we are trying to test it properly, so that there won't be any unexpected bugs. 

As for [deb240e](https://github.com/codex-team/editor.js/pull/1854/commits/deb240ee04511925367fcf01444dde19a6fb3430), there was an error after attempt to set caret to input of unsupported type: `Uncaught TypeError: Cannot read properties of undefined (reading 'top')`, and that is the fix

image

PetroSilenius:
Cool to hear! And thanks for taking care of the bug! I might have been a bit too easy on the testing after the newer changes so I didn't catch that😄

neSpecc:
Please, cover the change with tests

PetroSilenius:
> Please, cover the change with tests

Added some test cases now, let me know if there's anything to improve in them!

PR #1857, opened 2022-01-05
[Fix] isFunction function for async functions

### LINKED ISSUE #1803: [Bug] Async onPaste handlers not recognized
When using an async function as an `onPaste` handler in a tool, the tool's `pasteConfig` is not recognized by editor.js. The problem is [this check here](https://github.com/codex-team/editor.js/blob/4e7b33c2b89d184726fab7c7ca9c3f784183773f/src/components/modules/paste.ts#L293). The function [`isFunction`](https://github.com/codex-team/editor.js/blob/4e7b33c2b89d184726fab7c7ca9c3f784183773f/src/components/utils.ts#L197) returns `false` for async functions. And this is because the function [`typeof`](https://github.com/codex-team/editor.js/blob/4e7b33c2b89d184726fab7c7ca9c3f784183773f/src/components/utils.ts#L185) returns the string `asyncfunction` instead of `function`. In my opinion, the implementation of `isFunction` is wrong. It should return `true` for async functions, because async functions *are* functions, too. 

Steps to reproduce:
1. Import a tool whose `onPaste` handler is an async function
2. Paste something that matches the tool's `pasteConfig`

Behavior: The pasted content is pasted inside a new default block.

Expected behavior: The pasted content should be pasted inside a new block of the imported tool.

ilyamore88:
I've added a second condition for AsyncFunction and changed the return type because Function is a banned type.
This PR will fix #1803.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -20,7 +20,7 @@
 - `New` — `API` — The new `UiApi` section was added. It allows accessing some editor UI nodes and methods.
 - `Refactoring` — Toolbox became a standalone class instead of a Module. It can be accessed only through the Toolbar module.
 - `Refactoring` — CI flow optimized.
-- `Fix` - Recognize async onPaste handlers in tools [#1803](https://github.com/codex-team/editor.js/issues/1803)
+- `Fix` - Recognize async `onPaste` handlers in tools [#1803](https://github.com/codex-team/editor.js/issues/1803).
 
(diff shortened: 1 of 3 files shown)


gohabereg (approved):

neSpecc (changes requested):
Pls add a changelog and increment a version

neSpecc (approved):

PR #1928, opened 2022-01-18
Fix: check for Files substitutions before handling files paste

swapster:
I've faced with an unexpected behaviour when pasted some text from MS Word (v. 16.57, build 22011101, macOS). 
Paste didn't work. 
I've made some debugging and found that `DataTransferItemList` object in this case looks like

```js
[
   {kind: 'string', type: 'text/plain'},
   {kind: 'string', type: 'text/html'},
   {kind: 'string', type: 'text/rtf'},
   {kind: 'file', type: 'image/png'}
]
```

and `types` list is

```js
[
   'text/plain',
   'text/html',
   'text/rtf',
   'Files'
]
```

So, MS Word adds a screenshot to clipboard when you copy a text :) 
And now, as I understand, there is no ways to handle text pasting in this case. 

My suggestion is to add a check for Files substitutions before `processFiles` call (and return from `processDataTransfer` method). If there is at least one substitution for files, behaviour stay the same. But if you don't need to handle files pasting, you will be able to handle a text.


swapster:
@gohabereg @khaydarov @neSpecc Hi! 
Is there any chance of merging this PR in coming weeks? :) 
Or maybe there will be another solution for MS Word paste?

neSpecc:
I've tested pasting from the Microsoft Word for Mac 16.61.

At the `next` branch as well as at this branch.

In Safari the pasting works fine in both cases:

1) When the Image or SimpleImage tool connected — the text is pasting well
2) When there is no Image or SimpleImage tool — the text is pasting well

In Chrome at the `next` branch:

3) When the Image or SimpleImage tool connected — the image (screenshot) is pasting
4) When there is no Image or SimpleImage tool — **nothing is happenning, pasting is not working**

In Chrome at this branch:

5) When the Image or SimpleImage tool connected — the image (screenshot) is pasting
6) When there is no Image or SimpleImage tool — **the text is pasting well**


So it looks like this PR does not fixes the screenshot pasting when we have some tool supporting File pasting connected.

But it fixes the case (4), so maybe we can merge it. Or think about more universal solution that will always treat pasting from the Microsoft Work as a text.

kamil-ciesielski:
UP: @gohabereg @khaydarov @neSpecc

kamil-ciesielski (approved):

swapster:
> So it looks like this PR does not fixes the screenshot pasting when we have some tool supporting File pasting connected.

Right, this issue stay unfixed. 

> think about more universal solution that will always treat pasting from the Microsoft Work as a text.

I think this is the task for another PR :) 
Current small fix can help in some cases and it is literally **fix** (checking for handlers before handle something). 
Correct pasting from MS Word is a more complex task and will require more testing and carefull refactoring. 
I think it can be done separately)

kamil-ciesielski:
> > So it looks like this PR does not fixes the screenshot pasting when we have some tool supporting File pasting connected.
> 
> Right, this issue stay unfixed.
> 
> > think about more universal solution that will always treat pasting from the Microsoft Work as a text.
> 
> I think this is the task for another PR :) Current small fix can help in some cases and it is literally **fix** (checking for handlers before handle something). Correct pasting from MS Word is a more complex task and will require more testing and carefull refactoring. I think it can be done separately)

And when could it be for release? I need this feature badly.

ilyamore88 (approved):

neSpecc (approved):

PR #1959, opened 2022-02-05
target block move event

Dellord:
Both blocks should know when they moved

### DIFF
--- a/src/components/blocks.ts
+++ b/src/components/blocks.ts
@@ -181,6 +181,15 @@ export default class Blocks {
     });
 
     block.call(BlockToolAPI.MOVED, event);
+
+    // invoke hook target
+    const targetBlock = this.blocks[fromIndex];
+    const eventTarget: MoveEvent = this.composeBlockEvent('move', {
+      fromIndex: toIndex,
+      toIndex: fromIndex,
+    });
+
+    targetBlock.call(BlockToolAPI.MOVED, eventTarget);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


neSpecc:
Please, attach an Issue describing your use case.

What do you mean `both` blocks? If I'm moving the single block via MoveUp Tune, there are two blocks that change their indexes. 

But what if I'm moving a block to a more distant location? For example, using `api.blocks.move()` or via drag-n-drop (soon). All following block will change their indexes. So they all should receive `moved` event?

Dellord:
I worked on a feature when block changes self settings depends on position.

For example: Header in position 0 is always level 1, in position 1 - level 2, other headers - level 3 - 6.

But when you moving blocks via MoveUp Tune, only one block receives move event, on which you clicked.

neSpecc:
> I worked on a feature when block changes self settings depends on position.
> 
> For example: Header in position 0 is always level 1, in position 1 - level 2, other headers - level 3 - 6.

I see. This behavior could also be achieved using the `onChange` callback, then just saving the data and looking at the Header position. Blocks API allows changing Header block level then.

Dellord:
> > I worked on a feature when block changes self settings depends on position.
> > For example: Header in position 0 is always level 1, in position 1 - level 2, other headers - level 3 - 6.
> 
> I see. This behavior could also be achieved using the `onChange` callback, then just saving the data and looking at the Header position. Blocks API allows changing Header block level then.

this method requires some code outside of block

PR #1998, opened 2022-03-14
Fix/focus active item 1383

### LINKED ISSUE #1383: Focus active item of Block Settings by opening it from keyboard
1. Describe the problem.

When I'm opening a Block Settings by Tab keydown, they are opening with the first item focused. 

image

It would be nice if the active item will be focused by default

gklake:
Fix #1383 
Automatically selects the correct header for keyboard focus.

### DIFF
--- a/src/components/domIterator.ts
+++ b/src/components/domIterator.ts
@@ -109,7 +109,7 @@ export default class DomIterator {
       return this.cursor;
     }
 
-    let focusedButtonIndex = this.cursor;
+    let focusedButtonIndex = getActiveToolByCssClass();
 
     /**
      * If activeButtonIndex === -1 then we have no chosen Tool in Toolbox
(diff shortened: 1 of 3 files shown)


src/components/domIterator.ts
```diff
@@ -174,5 +174,16 @@ export default class DomIterator {
      * Return focused button's index
      */
     return focusedButtonIndex;
+
+    function getActiveToolByCssClass(){
+      const activeToolWindow = document.querySelector('.ce-settings--opened');
```
neSpecc, line 179: The dom iterator is the abstract utility. It shouldn't know about Toolbar CSS classes.

neSpecc (changes requested):

neSpecc:
I'm closing it now, but you can create another solution and open another PR

PR #2004, opened 2022-03-20
feat(toolbox): popover adapted for mobile devices

TatianaFomina:
This pull request includes changes related to making toolbox vertical instead of horizontal (mostly done by @neSpecc).
From my side there is an attempt to fix mobile popover positioning by changing the way top position is set to popover (from `transform` to `top`). Also added overlay under popover on mobile and disabled toolbox buttons hover for touch screens.

### DIFF
--- a/src/components/utils/popover.ts
+++ b/src/components/utils/popover.ts
@@ -2,6 +2,7 @@ import Dom from '../dom';
 import Listeners from './listeners';
 import Flipper from '../flipper';
 import SearchInput from './search-input';
+import { isMobile } from '../utils';
 
 /**
  * Describe parameters for rendering the single item of Popover
@@ -96,7 +97,7 @@ export default class Popover {
     noFoundMessageShown: string;
     popoverOverlay: string;
     popoverOverlayHidden: string;
-    } {
+  } {
     return {
       popover: 'ce-popover',
       popoverOpened: 'ce-popover--opened',
@@ -160,6 +161,11 @@ export default class Popover {
     if (this.searchable) {
       window.requestAnimationFrame(() => {
         this.search.focus();
+      });
+    }
+    if (isMobile) {
+      this.listeners.on(document, 'scroll', () => {
+        this.hide();
       });
     }
   }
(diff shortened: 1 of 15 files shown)


test/cypress/tests/onchange.spec.ts
```diff
@@ -171,6 +171,11 @@ describe('onChange callback', () => {
   it('should fire onChange callback when block is removed', () => {
     createEditor();
 
+    cy.get('[data-cy=editorjs]')
```
neSpecc, line 177: pls leave a comment for these lines

src/styles/variables.css
```diff
@@ -1,5 +1,6 @@
 @custom-media --mobile (width <= 650px);
```
neSpecc, line 4: pls add a comment noticing that changing of this value should be done with changing it at the `utils.ts@isMobile`

src/styles/toolbar.css
```diff
@@ -4,7 +4,8 @@
   right: 0;
   top: 0;
   transition: opacity 100ms ease;
-  will-change: opacity, transform;
+  will-change: opacity;
```
neSpecc:
```suggestion
  will-change: opacity, top;
```

src/components/utils/popover.ts
```diff
@@ -17,6 +18,11 @@ export interface PopoverItem {
    */
   label: string;
 
+  /**
+   * Item name
```
neSpecc, line 23: pls describe the reason of this property at the doc

src/components/utils/popover.ts
```diff
@@ -255,6 +302,8 @@ export default class Popover {
    */
   private createItem(item: PopoverItem): HTMLElement {
     const el = Dom.make('div', Popover.CSS.item);
+
+    el.setAttribute('data-tool', item.name);
```
neSpecc:
```suggestion
    el.setAttribute('data-item-name', item.name);
```
A popover shouldn't know what does the `tool` mean

src/styles/variables.css
```diff
@@ -119,9 +123,11 @@
       height: var(--toolbox-buttons-size--mobile);
     }
 
-    &:hover,
-    &--active {
-      background-color: var(--bg-light);
+    @media (--can-hover) {
+      &:hover,
+      &--active {
```
neSpecc: but the `active` state does not depend on hovering ability?

neSpecc:
The scroll lock doesn't work (tested on Safari, iOS 12). Other is fine.

neSpecc (approved):

PR #2007, opened 2022-03-29
Fix OnChange Event Invocation

### LINKED ISSUE #1997: onChange event not called when adding new block
When I create a new element, e.g. a delimiter, and instantly store my content, the new element is not stored because the `onChange` event is not triggered. But when creating a new element and changing content (active keyup), the element gets stored because the `onChange` is triggered.

Expected behavior: `onChange` should be called after adding a new element.

Screenshots: https://www.loom.com/share/5bba14a6eb844bfdae677b47ee5e2553

Device, Browser, OS: macOS Safari

Editor.js version: 2.23.2

robonetphy:
resolve #1997

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -14,21 +14,21 @@
 
 ### 2.23.0
 
-- `Improvement` — _EditorConfig_ — The `onChange` callback now accepts two arguments: EditorJS API and the CustomEvent with `type` and `detail` allowing to determine what happened with a Block
-- `New` — _Block API_ — The new `dispatchChange()` method allows to manually trigger the 'onChange' callback. Useful when Tool made a state mutation that is invisible for editor core.
-- `Improvement` — _UI_ — Block Tunes toggler moved to the left
-- `Improvement` — _UI_ — Block Actions (BT toggler + Plus Button) will appear on block hovering instead of click
-- `Improvement` — _UI_ — Block Tunes toggler icon and Plus button icon updated
-- `Improvement` — _Dev Example Page_ — The menu with helpful buttons added to the bottom of the screen
-- `Improvement` — _Dev Example Page_ — The 'dark' theme added. Now we can code at night more comfortably.
-- `Improvement` — _Rectangle Selection_ — paint optimized
-- `Fix` — _Rectangle Selection_ — the first click after RS was not clear selection state. Now does.
-- `Improvement` — _Blocks API_ — toolbar moving logic removed from `blocks.move()` and `blocks.swap()` methods. Instead, you should use Toolbar API (it was used by MoveUp and MoveDown tunes, they were updated).
-- `New` — _Blocks API_ — The `getBlockIndex()` method added
-- `New` — _Blocks API_ — the `insert()` method now has the `replace: boolean` parameter
-- `New` — _Blocks API_ — the `insert()` method now returns the inserted `Block API`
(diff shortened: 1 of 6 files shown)


ilyamore88 (approved):

src/components/modules/saver.ts
```diff
@@ -28,15 +28,11 @@ export default class Saver extends Module {
...
    */
   public async save(): Promise<OutputData> {
-    const { BlockManager, Tools, ModificationsObserver } = this.Editor;
+    const { BlockManager, Tools } = this.Editor;
     const blocks = BlockManager.blocks,
         chainData = [];
 
     try {
-      /**
-       * Disable onChange callback on save to not to spam those events
```
neSpecc, line 37: Do you mean that this problem is not relevant for now? Why?

neSpecc: And how is it related to the Delimiter tool?

robonetphy: Its not related to the delimiter because when we call `save` function inside OnChange event it disables upcomming onchange events by calling `modificationObserver.disable()`.

robonetphy: @neSpecc do you have any test cases such that I can check because I don't know if it occurs or not

> Do you mean that this problem is not relevant for now? Why?

neSpecc: I've dug in git history a little bit, but could not find the reason why that disable/enable logic was added. So I think, it could be removed.

docs/CHANGELOG.md
```diff
@@ -10,21 +14,21 @@
...
-- `Improvement` — *Dev Example Page* — The 'dark' theme added. Now we can code at night more comfortably.
-- `Improvement` — *Rectangle Selection* — paint optimized
-- `Fix` — *Rectangle Selection* — the first click after RS was not clear selection state. Now does.
-- `Improvement` — *Blocks API* — toolbar moving logic removed from `blocks.move()` and `blocks.swap()` methods. Instead, you should use Toolbar API (it was used by MoveUp and MoveDown tunes, they were updated).
-- `New` — *Blocks API* — The `getBlockIndex()` method added
-- `New` — *Blocks API* — the `insert()` method now has the `replace: boolean` parameter
-- `New` — *Blocks API* —  the `insert()` method now returns the inserted `Block API`
-- `New` — *Listeners API* — the `on()` method now returns the listener id.
-- `New` — *Listeners API* — the new `offById()` method added
+- `Improvement` — _EditorConfig_ — The `onChange` callback now accepts two arguments: EditorJS API and the CustomEvent with `type` and `detail` allowing to determine what happened with a Block
```
neSpecc: why the whole changelog has changed?

robonetphy: Ohh sorry it's due to my editor configuration

neSpecc (approved):

gohabereg (approved):

robonetphy:
@gohabereg I checked `tests` with `next` branch is failing.
onChange callback -- should fire onChange callback on block insertion with save inside onChange (failed)

ilyamore88 (approved):

gohabereg (approved):

drakst:
Hi. Instead of onChange event you can try to connect MutationObserver to editor ad watch for updates.

Config of observe can be something like: {
	attributes: false,
	childList: true,
	subtree: true,
	characterData: true
}

Im using this config on vue project for change v-model value

PR #2009, opened 2022-03-31
feat(dx): add a dev-server for debugging on mobile devices from a local network

TatianaFomina:
Added simple static server allowing to open files from `/example` from local network. Could be useful for testing on mobile devices.

### DIFF
--- a/devserver.js
+++ b/devserver.js
@@ -1,11 +1,69 @@
-const express = require('express');
+/**
+ * Server for static files.
+ * Usage:
+ *  1. run `yarn devserver:start`
+ *  2. get access to files from `/example` or `/dist` on any device in local network.
+ *     example: `http://{ip_address}:3000/example/example-dev.html`
+ *     where {ip_address} is IP of your machine in local network.
+ */
 const path = require('path');
-const app = express();
+const fs = require('fs');
+const http = require('http');
+
 const port = 3000;
+const host = '0.0.0.0';
+const server = http.createServer(serveStatic(['/example', '/dist']));
 
-app.use('/dist', express.static(path.resolve(__dirname, 'dist')));
-app.use(express.static(path.resolve(__dirname, 'example')));
-
-app.listen(port, () => {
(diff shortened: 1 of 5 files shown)


devserver.js
```diff
@@ -0,0 +1,11 @@
+const express = require('express');
```
neSpecc: Maybe we could use internal `http` module instead of `express` to get rid of extra deps?

We have a similar [dev server](https://github.com/editor-js/image/blob/master/dev/server.js) at the Image tool

neSpecc: and also, provide some docs of how to use it, please

TatianaFomina: ok, i'll take a look

neSpecc (changes requested):
Dependencies were not changed, so there shouldn't be a lock file change.

devserver.js
```diff
@@ -0,0 +1,69 @@
...
+const path = require('path');
+const fs = require('fs');
+const http = require('http');
+
+const port = 3000;
+const host = '0.0.0.0';
+const server = http.createServer(serveStatic(['/example', '/dist']));
+
+server.listen(port, host, () => {
+  console.log(`Server is running on http://${host}:${port}`);
```
neSpecc: I think we can make this message more developer-friendly. Add some instructions: how to know the network IP, maybe other useful info. And also, the should be a link to the example page.

devserver.js
```diff
@@ -0,0 +1,69 @@
...
+      });
+      readStream.pipe(response);
+    } catch (e) {
+      response.writeHead(500);
+      response.end(e.toString());
+    }
+  };
+}
+
+// const express = require('express');
```
neSpecc: remove unused code, please

neSpecc (approved):

ilyamore88 (approved):

gohabereg (approved):

neSpecc:
Add a changelog please

PR #2017, opened 2022-04-09
Vertical toolbox fixes

TatianaFomina:
Includes fixes on
- arrow keys navigation within toolbox popover search bar
- hidden popover causing rendering issues

### DIFF
--- a/src/components/flipper.ts
+++ b/src/components/flipper.ts
@@ -18,20 +18,22 @@ export interface FlipperOptions {
   items?: HTMLElement[];
 
   /**
-   * Defines arrows usage. By default Flipper leafs items also via RIGHT/LEFT.
-   *
-   * true by default
-   *
-   * Pass 'false' if you don't need this behaviour
-   * (for example, Inline Toolbar should be closed by arrows,
-   * because it means caret moving with selection clearing)
-   */
-  allowArrows?: boolean;
-
-  /**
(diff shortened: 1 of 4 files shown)


src/components/flipper.ts
```diff
@@ -206,23 +206,19 @@ export default class Flipper {
    * @returns {boolean}
    */
   private isEventReadyForHandling(event: KeyboardEvent): boolean {
-    const handlingKeyCodeList = [
-      _.keyCodes.TAB,
-      _.keyCodes.ENTER,
-    ];
-
     const isCurrentItemIsFocusedInput = this.iterator.currentItem == document.activeElement;
```
neSpecc: do we need this statement in the new logic? Looks like it was used to handle similar problem with arrows in inputs, but it does not cover the case when you allows up/down but disallow left/right.

src/styles/popover.css
```diff
@@ -1,9 +1,8 @@
 .ce-popover {
   position: absolute;
   opacity: 0;
-  visibility: hidden;
```
neSpecc, line 4: unfortunatelly, this breaks the animation. Try to open/close the popover several times.

neSpecc (changes requested):

neSpecc (approved):

gohabereg (approved):

PR #2018, opened 2022-04-10
[Feature] Add state manager

ilyamore88:
In this request I've added state manager to Editor.js and written a small demo (this file will be removed after adding tests):

```shell
yarn global add ts-node
ts-node demo.ts
```

In the next two requests, I will add tests for the state manager and will connect them to the core.

## State manager

State manager is used for managing the internal state of an application. In our situation, we need to contain the data of each block in the editor. This data will be used for collaborative editing when we need to compute diffs for new block data.

We can subscribe to state changes to get new data from it and compute diffs.


src/components/store/reducer.ts
```diff
@@ -0,0 +1,22 @@
+import { ActionType } from './types/ActionType';
+import { EditorState } from './types/EditorState';
+import { Action } from './types/Action';
+
+const reducer = (state: EditorState, action: Action): EditorState => {
+  const stateCopy = { ...state };
```
gohabereg: To make store fully immutable, it's better to use deep copy here

src/components/store/types/Store.ts
```diff
@@ -0,0 +1,28 @@
+import { Action } from './Action';
+import { EditorState } from './EditorState';
+
+/**
+ * Store type contains functions for use it
+ */
+export type Store = {
```
gohabereg: Looks like interface would be better here

src/components/store/types/Action.ts
```diff
@@ -0,0 +1,22 @@
...
+
+type ChangeBlockDataAction = {
+  type: ActionType.CHANGE_BLOCK_DATA;
+  data: OutputBlockData;
+}
+
+type RemoveBlockAction = {
+  type: ActionType.REMOVE_BLOCK;
+  blockId: string;
+};
```
gohabereg: Better to use interfaces here

neSpecc:
Please, add more detailed description. What is the State Manager and for what it will be used

gohabereg (approved):

src/components/store/createStore.ts
```diff
@@ -0,0 +1,56 @@
+import { EditorState } from './types/EditorState';
+import { Reducer } from './types/Reducer';
+import { Action } from './types/Action';
+import { Store } from './types/Store';
+
+/**
+ * Function creates store and returns functions for use it
+ *
+ * @param reducer - current Editor reducer function
```
neSpecc: the description is not clear.

src/components/store/createStore.ts
```diff
@@ -0,0 +1,56 @@
...
+ * @param initialState - initial state of the store
+ */
+const createStore = (reducer: Reducer, initialState: EditorState = {}): Store => {
```
neSpecc: why anonymous function? It would better to use named function insted, i think.

```suggestion
function createStore(reducer: Reducer, initialState: EditorState = {}): Store => {
```
```

File: src/components/store/reducer.ts [hunk 6]
```diff
@@ -0,0 +1,23 @@
+import { ActionType } from './types/ActionType';
+import { EditorState } from './types/EditorState';
+import { Action } from './types/Action';
+import * as _ from '../utils';
+
+const reducer = (state: EditorState, action: Action): EditorState => {
```
neSpecc: Description missed.

src/components/store/types/Action.ts
```diff
@@ -0,0 +1,22 @@
+import { ActionType } from './ActionType';
+import { OutputBlockData } from '../../../../types';
+
+interface CreateBlockAction {
```
neSpecc: jsdocs missed

src/components/store/types/ActionType.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Available action types
+ */
+export enum ActionType {
+  CREATE_BLOCK = 'CREATE_BLOCK',
```
neSpecc:
```suggestion
  CreateBlock = 'CREATE_BLOCK',
```

See [Enums](https://www.typescriptlang.org/docs/handbook/enums.html)

neSpecc: do we need string representations?

neSpecc: and also, add jsdocs please for every case

src/components/store/types/EditorState.ts
```diff
@@ -0,0 +1,7 @@
+import { OutputBlockData } from '../../../../types';
+
+/**
+ * Type of the state object
+ * It contains a blocks' data by block IDs as a key
+ */
+export type EditorState = Record<string, OutputBlockData>;
```
neSpecc: Looks like it contains the state only of Editor Blocks, not the whole Editor state.

Maybe something like this would more useful in future?


```ts
type EditorState = {
  blocks: Record<string, OutputBlockData>,
  // other state properties
}
```

src/components/store/types/Reducer.ts
```diff
@@ -0,0 +1,7 @@
+import { EditorState } from './EditorState';
```
neSpecc: we store all types at the `types` directory. Maybe it would better to place the Store types same way?

```
/types/
  /store/
    ...
```

src/components/store/types/Store.ts
```diff
@@ -0,0 +1,28 @@
+import { Action } from './Action';
```
neSpecc: We name files in lowercase at other places, so I would suggest to rename `Store.ts` to `store.ts` as well as other files.

src/components/store/types/Store.ts
```diff
@@ -0,0 +1,28 @@
...
+ */
+export interface Store {
+  /**
+   * Function for subscribing on state changes
+   *
+   * @param listener - function, that will execute every state change
+   *
+   * @returns {() => void} unsubscribe function
+   */
+  subscribe: (listener: () => void) => (() => void);
```
neSpecc: Why there is no ability to subscribe to the specific action?

src/components/store/types/ActionType.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Available action types
+ */
+export enum ActionType {
```
neSpecc: It looks similar with the [BlockMutationType](https://github.com/codex-team/editor.js/blob/next/types/events/block/mutation-type.ts)

Could we use the same type in both cases?

neSpecc (changes requested):

src/components/store/types/Store.ts
ilyamore88: Cause I used redux as a reference. Subscribe function is used for subscribing to each state change.

src/components/store/types/ActionType.ts
ilyamore88: I use BlockMutationType instead

src/components/store/createStore.ts
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * It returns functions to use the store:
+ *  subscribe - function for subscribing to each state changes
+ *  dispatch - function for applying actions to the store
+ *  getState - function returns a current state of the store
+ *
+ * @param reducer - current Editor reducer function
+ * @param initialState - initial state of the store
+ */
+function createStore(reducer: Reducer, initialState: EditorState = { blocks: {} }): Store {
```
neSpecc, line 23: why the Store is not a class?

src/components/store/reducer.ts
```diff
@@ -0,0 +1,30 @@
+import { EditorState } from '../../../types/store/editorState';
+import { Action } from '../../../types/store/action';
+import * as _ from '../utils';
+import { BlockMutationType } from '../../../types/events/block/mutation-type';
+
+/**
+ * Reducer function for Editor.js state
+ * This function applies actions in the current state
```
neSpecc:
```suggestion
 * This function applies the passed action to the current state and returns the new state
```

types/store/reducer.ts
```diff
@@ -0,0 +1,7 @@
+import { EditorState } from './editorState';
+import { Action } from './action';
+
+/**
+ * Type of reducer function
```
neSpecc:
```suggestion
 * Type of the function that applies the passed action to the current state and returns the new state
```

src/components/store/createStore.ts
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * This function is an entry point to use the store in the editor
+ * It creates the store with an initial state
+ *
+ * It returns functions to use the store:
+ *  subscribe - function for subscribing to each state changes
+ *  dispatch - function for applying actions to the store
+ *  getState - function returns a current state of the store
+ *
+ * @param reducer - current Editor reducer function
```
neSpecc:
```suggestion
 * @param reducer - the function that applies the passed action to the current state and returns the new state
```

src/components/store/createStore.ts
neSpecc, line 23: please, describe in jsdoc why do we pass the reducer to the Store. Is there are some usecases when there will be a different reducer? Or it is just a DI for the tests

ilyamore88: I think, we don't need the class for the Store. Cause we will probably use the Store as a singleton. So, I don't see any benefit in using the class here.

ilyamore88: It's not a part of the Store. I can pass different reducers to the Store for different tasks. Also, I can implement function to combine reducers and pass it to the Store (we don't need it right now, but it will help us to store not only blocks data).

neSpecc: Please, describe it in a jsdoc, not in a PR.

> for different tasks

Some examples will be helpful. I still can't get the idea.

src/components/store/reducer.ts
```diff
@@ -0,0 +1,30 @@
...
+import { BlockMutationType } from '../../../types/events/block/mutation-type';
+
+/**
+ * Reducer function for Editor.js state
+ * This function applies the passed action to the current state and returns the new state
+ *
+ * @param state - previous state to apply action
+ * @param action - information about the action in the previous state
+ */
+const reducer = (state: EditorState, action: Action): EditorState => {
```
neSpecc: why it is the anonymous function? 

```suggestion
export default function reducer(state: EditorState, action: Action): EditorState {
```

neSpecc: also, naming is not clear. If you're talking that we can have several different reducers, so we need to call this reducer somehow more specific.

Along with that, looks like this reducer works not with the whole Editor State, but only with the `blocks` property. 

Maybe it would better to call it `BlocksReducer`?

src/components/store/createStore.ts
ilyamore88: We have the reducer for blocks (I can rename `reducer` function to `blocksReducer`):
- This reducer works with `state.blocks` field;
- Has logic only for blocks actions.

If I want to work with Editor parameters, for example with Editor version:
- I will create `parametersReducer`;
- This reducer will work with `state.parameters.version` field;
- It has logic only for Editor parameters.

So, I have two different reducers and I need to use this two reducers together. I can create `combineReducers` function and combine all reducers to one big reducer and pass it to `createStore` function.

I need to store new data -> I create a new reducer -> I add the reducer to other reducers -> I don't need to change the logic of `createStore` function

I will add information about it to JSdoc. The main idea is that I can add new data to the storage without changing the storage creation interface.

src/components/store/reducer.ts
ilyamore88: Yep, it's a good idea, I will change it

neSpecc (approved):

src/components/store/createStore.ts
```diff
@@ -0,0 +1,66 @@
...
+  const currentListeners = [];
+
+  /**
+   * Function for subscribing on state changes
+   *
+   * @param listener - function, that will execute every state change
+   *
+   * @returns {() => void} unsubscribe function
+   */
+  const subscribe = (listener): (() => void) => {
```
TatianaFomina: param type missing

src/components/store/createStore.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  /**
+   * Dispatch action on the current state
+   *
+   * @param action - action that will be dispatched
+   */
+  const dispatch = (action: Action): void => {
+    state = currentReducer(state, action);
+    currentListeners.forEach((listener) => {
+      listener();
```
TatianaFomina: probably would make sense to pass new state to listener as a param

TatianaFomina (approved):

PR #2022, opened 2022-04-16
fix: popover opening direction

TatianaFomina:
Now toolbox popover opening direction is calculated based on amount of available space below and above it.

### DIFF
--- a/src/components/ui/toolbox.ts
+++ b/src/components/ui/toolbox.ts
@@ -110,12 +110,6 @@ export default class Toolbox extends EventsDispatcher<ToolboxEvent> {
   private clickListenerId: string = null;
 
   /**
-   * Stores popover height.
-   * Allows not to recalculate height on each popover opening
-   */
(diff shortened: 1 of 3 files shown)


src/components/ui/toolbox.ts
```diff
@@ -108,6 +109,12 @@ export default class Toolbox extends EventsDispatcher<ToolboxEvent> {
    */
   private clickListenerId: string = null;
 
+  /**
+   * Stores popover height.
+   * Allows not to recalculate height on each popover opening
+   */
+  private popoverHeightCached: number | null = null;
```
neSpecc: you don't need this. We have a special `cacheable` decorator for that reason.

Try something like this:

image

neSpecc (changes requested):

src/components/ui/toolbox.ts
TatianaFomina: fixed

neSpecc (approved):

PR #2023, opened 2022-04-17
[Feature] Add tests of state manager

ilyamore88:
In this request I've added tests for our state manager.


test/cypress/tests/store/store.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+          text: 'Editor.js',
+          level: 2,
+        },
+      };
+      const expectedResult = {
+        '3JPEqh8_Wc': block,
+      };
+
+      store.dispatch({
+        type: ActionType.CREATE_BLOCK,
```
gohabereg: Would be great to have action creators

```
const createBlock = block => ({
  type: ActionType.CREATE_BLOCK,
  data: block
});
```

gohabereg (approved):

test/cypress/tests/store/store.spec.ts
ilyamore88: For tests or inside the app?

gohabereg: Both

gohabereg (approved):

test/cypress/tests/store/store.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+        },
+      };
+      const expectedResult = {
+        blocks: {
+          '3JPEqh8_Wc': block,
+        },
+      };
+
+      store.dispatch({
+        type: BlockMutationType.Added,
```
TatianaFomina: What do you think about adding a test case for each of the `BlockMutationType`'s ?

src/components/store/actions/blockReducerActions.ts
```diff
@@ -0,0 +1,33 @@
+import { OutputBlockData } from '../../../../types';
+import { BlockMutationType } from '../../../../types/events/block/mutation-type';
+import { ChangeBlockDataAction, CreateBlockAction, RemoveBlockAction } from '../../../../types/store/action';
+
+/**
+ * Action creator for creating a new block in the store
+ *
+ * @param block - new block data
+ */
+export const createBlock = (block: OutputBlockData): CreateBlockAction => ({
```
neSpecc: why anonymous function? Its better to create regular named function:

```ts
function createBlock(block: OutputBlockData): CreateBlockAction {
  return {
    // ...
  }
}

```

neSpecc: and probably it would better to name it `makeCreateBlockAction` because that is what this function does.

TatianaFomina (approved):

neSpecc (approved):

PR #2050, opened 2022-05-04
[Feature] Multiple toolbox items for single tool

TatianaFomina:
Adds support of multiple toolbox entries for single tool. 
A tool's toolbox property now can be represented as an array of items. Which item can include tool's default config overrides (such as level for header tool). These config overrides are merged with default tool settings when inserting relevant block. 

Example of tool with multiple toolbox items can be found at https://github.com/editor-js/header/pull/79


src/components/block/index.ts
```diff
@@ -734,6 +746,25 @@ export default class Block extends EventsDispatcher<BlockEvents> {
...
 
+  /**
+   * Returns current active toolbox entry
+   */
+  public get activeToolboxEntry(): ToolboxConfig {
+    return Array.isArray(this.tool.toolbox) ? this.toolInstance.activeToolboxEntry : this.tool.toolbox;
+  }
+
+  /**
+   *
```
ilyamore88: Missing doc

src/components/tools/base.ts
```diff
@@ -78,6 +78,10 @@ export enum InternalBlockToolSettings {
    * Tool Toolbox config
    */
   Toolbox = 'toolbox',
+  /**
+   * Current active toolbox item
+   */
+  ToolboxItem = 'toolboxItem',
```
neSpecc: There is no such an option at the Block Tool API for now. As I can see, there in only the `toolbox` and the `activeToolboxEntry`

src/components/tools/block.ts
```diff
@@ -72,19 +72,17 @@ export default class BlockTool extends BaseTool<IBlockTool> {
...
-    if (_.isEmpty(toolToolboxSettings)) {
-      return;
-    }
-
-    if ((userToolboxSettings ?? toolToolboxSettings) === false) {
-      return;
+    if (Array.isArray(toolToolboxSettings)) {
+      return toolToolboxSettings.map(item => this.getActualToolboxSettings(item)).map(item => ({
+        ...item,
+        id: _.md5(item.icon + item.title),
```
neSpecc: It would be better to move this logic to a separate method

src/components/modules/blockManager.ts
```diff
@@ -277,6 +282,7 @@ export default class BlockManager extends Module {
     needToFocus = true,
     replace = false,
     tunes = {},
+    config,
```
neSpecc: Two things looks strange:

1) `config` means the Tool's config, but we're already passing the `tool`. It could be look like a redundant property (that could be got via the `tool.config`) from some perspective. Maybe we need to clarify naming.
2) `config` is passing again and again through many methods. Maybe we could get this information at the lowest level once?

src/components/modules/renderer.ts
```diff
@@ -100,8 +100,9 @@ export default class Renderer extends Module {
 
       if (Tools.unavailable.has(tool)) {
         const toolboxSettings = (Tools.unavailable.get(tool) as BlockTool).toolbox;
+        const toolboxTitle = (Array.isArray(toolboxSettings) ? toolboxSettings[0] : toolboxSettings)?.title;
```
neSpecc: maybe it should be done on the `BlockTool` class's side? I'm not sure.

src/components/modules/toolbar/conversion.ts
```diff
@@ -47,9 +48,9 @@ export default class ConversionToolbar extends Module<ConversionToolbarNodes> {
   public opened = false;
 
   /**
-   * Available tools
+   * Available tools data
    */
-  private tools: { [key: string]: HTMLElement } = {};
+  private tools: {name: string; toolboxItem: ToolboxConfig; button: HTMLElement}[] = []
```
neSpecc: `toolboxItem = toolboxConfig` — this looks not so clear because the Toolbox's config may consist of several items.

src/components/modules/toolbar/conversion.ts
```diff
@@ -167,18 +167,17 @@ export default class ConversionToolbar extends Module<ConversionToolbarNodes> {
    * Returns true if it has more than one tool available for convert in
    */
   public hasTools(): boolean {
-    const tools = Object.keys(this.tools); // available tools in array representation
-
-    return !(tools.length === 1 && tools.shift() === this.config.defaultBlock);
+    return !(this.tools.length === 1 && this.tools[0].name === this.config.defaultBlock);
```
neSpecc: the code could be simplified i guess

src/components/modules/toolbar/conversion.ts
```diff
@@ -276,49 +277,73 @@ export default class ConversionToolbar extends Module<ConversionToolbarNodes> {
...
         /**
          * Skip tools without «import» rule specified
          */
         if (!conversionConfig || !conversionConfig.import) {
           return;
         }
 
-        this.addTool(name, toolboxSettings.icon, toolboxSettings.title);
+        if (Array.isArray(tool.toolbox)) {
+          tool.toolbox.forEach((configItem, i) => this.addToolIfValid(name, configItem, (tool.toolbox as ToolboxConfig[]).slice(0, i)));
```
neSpecc: its better to move it to mulitple lines here

src/components/ui/toolbox.ts
```diff
@@ -262,24 +253,93 @@ export default class Toolbox extends EventsDispatcher<ToolboxEvent> {
...
+        if (Array.isArray(toolToolboxSettings)) {
+          const validToolboxSettings = toolToolboxSettings
+            .filter(item => this.areToolboxSetttingsValid(item, tool.name))
+            .filter((item, i) => {
+              const notUnique = toolToolboxSettings.slice(0, i).find(otherItem => otherItem.id === item.id);
+
+              if (notUnique) {
+                _.log('Toolbox entry has not unique combination of icon and title. Toolbox entry %o is skipped', 'warn', item.title);
+
+                return false;
```
neSpecc: maybe we should not skip non-unique items? I think, nothing horrible will happen, but the code will become a bit simpler

src/components/utils.ts
```diff
@@ -4,6 +4,7 @@
 
 import { nanoid } from 'nanoid';
 import Dom from './dom';
+import { createHash } from 'crypto';
```
neSpecc: It would be better to use native Crypto API, or even more simple method

src/components/modules/blockManager.ts
TatianaFomina: I guess `config` either could be renamed to `configOverrides` or replaced with id of selected toolbox entry. 
Anyway I don't see how passing a param could be avoided -- we need to know which toolbox entry was clicked when inserting a new block.

docs/tools.md
```diff
@@ -56,7 +56,7 @@ Options that Tool can specify. All settings should be passed as static propertie
 
 | Name | Type | Default Value | Description |
 | -- | -- | -- | -- |
-| `toolbox` | _Object_ | `undefined` | Pass here `icon` and `title` to display this `Tool` in the Editor's `Toolbox` <br /> `icon` - HTML string with icon for Toolbox <br /> `title` - optional title to display in Toolbox |
+| `toolbox` | _Object_ | `undefined` | Pass the `icon` and the `title` there to display this `Tool` in the Editor's `Toolbox` <br /> `icon` - HTML string with icon for the Toolbox <br /> `title` - title to be displayed at the Toolbox. <br /><br />May contain an array of `{icon, title, data}` to display the several variants of the tool, for example "Ordered list", "Unordered list". See details at the edi |
```
ilyamore88: Is it a mistake at the end?

ilyamore88 (approved):
👍

neSpecc (approved):

src/components/modules/toolbar/conversion.ts
```diff
@@ -252,6 +246,14 @@ export default class ConversionToolbar extends Module<ConversionToolbarNodes> {
       return;
     }
 
+    /**
+     * If this conversion fired by the one of multiple Toolbox items,
+     * extend converted data with this item's "data" overrides
+     */
+    if (blockDataOverrides) {
+      newBlockData = Object.assign(newBlockData, blockDataOverrides);
```
gohabereg, line 254: There might be a problem if override is not on the first level.
Eg.
```
// data
{
  settings: {
    level: 1, 
    anotherProp: 'value',
  }
}

// override
{
  config: {
    level: 2
  }
}
```

Maybe just left a todo to use deep merge in the future

src/components/modules/toolbar/conversion.ts
```diff
@@ -276,64 +278,98 @@ export default class ConversionToolbar extends Module<ConversionToolbarNodes> {
...
-
         /**
          * Skip tools without «import» rule specified
          */
         if (!conversionConfig || !conversionConfig.import) {
           return;
         }
 
-        this.addTool(name, toolboxSettings.icon, toolboxSettings.title);
+        if (Array.isArray(tool.toolbox)) {
```
gohabereg: Maybe we can unify that to array somewhere at the start, so we don't need to add checks everywhere

gohabereg (approved):

src/components/modules/api/blocks.ts
```diff
@@ -247,6 +249,27 @@ export default class BlocksAPI extends Module {
...
+    const block = new Block({
+      tool,
+      api: this.Editor.API,
+      readOnly: true,
+      data: {},
+      tunesData: {},
+    });
+    const blockData = await block.data;
+
+    return Object.assign(blockData, dataOverrides);
```
neSpecc: I think, merging is too specific thing. So it should be done outside ot this method, at the `toolbox.ts@insertNewBlock`. 

You've added the new public API method, so it should be useful for developers. I can't imagine any usecases for now, but let it be there. 

Also, please add a changelog about this new method.

TatianaFomina: done

src/components/modules/toolbar/conversion.ts
neSpecc: We will add a note at the docs describing that data overrides should have the same structure that block data

src/components/ui/toolbox.ts
```diff
@@ -346,9 +393,15 @@ export default class Toolbox extends EventsDispatcher<ToolboxEvent> {
      */
     const index = currentBlock.isEmpty ? currentBlockIndex : currentBlockIndex + 1;
 
+    /**
+     * Merge real tool's data with data overrides
+     */
+    const defaultBlockData = await this.api.blocks.composeBlockData(toolName);
```
neSpecc: I think we should do that only if we have the `blockDataOverrides`. In other cases (insertion of a regular tool) we will skip extra block construction/saving/merging.

types/api/blocks.d.ts
```diff
@@ -113,6 +113,13 @@ export interface Blocks {
   ): BlockAPI;
 
 
+  /**
+   * Retrieves default block data by creating fake block.
```
neSpecc:
```suggestion
   * Creates data of an empty block with a passed type.
```

src/components/modules/api/blocks.ts
```diff
@@ -247,6 +249,24 @@ export default class BlocksAPI extends Module {
     return new BlockAPI(insertedBlock);
   }
 
+  /**
+   * Retrieves default block data by creating fake block.
```
neSpecc:
```suggestion
   * Creates data of an empty block with a passed type.
```

neSpecc (approved):

PR #2055, opened 2022-05-09
[Fix] Stop jumping to block start on toolbox close

### LINKED ISSUE #2045: [Bug] Keyboard selection by word doesn't work when selecting from right to left
Describe a bug.

Steps to reproduce:
1. Go to https://editorjs.io
2. Click in the middle of any paragraph 
3. Try to select several words via keyboard from right to left (on MacOS Opt + Shift + Arrow left)

Expected behaviour:
Words should be selected

Actual behaviour:
After selecting the second word selection is lost and cursor jumps to the beginning of the paragraph

Device, Browser, OS:
Chrome, MacOS

Editor.js version:
2.24.2

TatianaFomina:
Fixes problem with text selection from right to left. Moving caret to block start on toolbox close was causing the problem. 

Resolves https://github.com/codex-team/editor.js/issues/2045

### DIFF
--- a/src/components/modules/toolbar/index.ts
+++ b/src/components/modules/toolbar/index.ts
@@ -167,7 +167,6 @@ export default class Toolbar extends Module<ToolbarNodes> {
       opened: this.toolboxInstance.opened,
       close: (): void => {
         this.toolboxInstance.close();
-        this.Editor.Caret.setToBlock(this.Editor.BlockManager.currentBlock);
       },
       open: (): void => {
         /**
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### 2.24.4
+
+- `Fix` — Keyboard selection by word [2045](https://github.com/codex-team/editor.js/issues/2045)
+
 ### 2.24.3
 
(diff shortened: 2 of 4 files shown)


src/components/modules/toolbar/index.ts
```diff
@@ -167,7 +167,6 @@ export default class Toolbar extends Module<ToolbarNodes> {
       opened: this.toolboxInstance.opened,
       close: (): void => {
         this.toolboxInstance.close();
-        this.Editor.Caret.setToBlock(this.Editor.BlockManager.currentBlock);
```
TatianaFomina, line 170: @neSpecc 
didn't get why this line is needed, so removed it. If it is needed, it can be executed on condition or moved to some other place

neSpecc: Try this:

1) Place the caret to the empty block
2) Press Tab to open the Toolbox
3) Press Esc to close the Toolbox

The Caret should return to the empty block. In case of removing this line that wouldn't happenn.

TatianaFomina: Ok, moved this line to function called on esc press

neSpecc (approved):

gohabereg (approved):

ilyamore88 (approved):

PR #2061, opened 2022-05-13
pref: remove range just editorjs'range

aquarius-wing:
I want to pref because this situation:
1. I get a list of editorjs element, and a search input
2. when I search, the editorjs will trigger readonly change,and the range of input is lost!

### DIFF
--- a/src/components/modules/blockSelection.ts
+++ b/src/components/modules/blockSelection.ts
@@ -14,6 +14,7 @@ import Shortcuts from '../utils/shortcuts';
 import SelectionUtils from '../selection';
 import { SanitizerConfig } from '../../../types/configs';
 import { clean } from '../utils/sanitizer';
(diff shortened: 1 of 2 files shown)


src/components/modules/blockSelection.ts
```diff
@@ -194,8 +195,13 @@ export default class BlockSelection extends Module {
    * @param {boolean} readOnlyEnabled - "read only" state
    */
   public toggleReadOnly(readOnlyEnabled: boolean): void {
-    SelectionUtils.get()
-      .removeAllRanges();
+    const sel = SelectionUtils.get();
```
neSpecc, line 198: We already have similar method `SelectionUtils.isAtEditor`, check it out

neSpecc:
The original problem is not clear, and a solution as well. Please describe a problem with an Issue first.

PR #2064, opened 2022-05-17
[Feature] Multiple toolbox items: using of data overrides instead of config overrides

neSpecc:
This PR is a proposal change to the #2050 

See also changes on the Block API side: https://github.com/editor-js/header/pull/80

I'm suggesting to use the `data` overrides instead of the `config` overrides
It allows:

- to get rid of the `activeToolboxEntry` getter (it could be matched by current block data)
- to get rid of passing `config` through many methods
- to not to force developers for using configs
- to allow developers overriding any Block properties via multiple toolbox items. For example, set predefined texts and so on.
- the data is used in blocks rendering logic already

### DIFF
--- a/src/components/block/index.ts
+++ b/src/components/block/index.ts
@@ -772,7 +772,7 @@ export default class Block extends EventsDispatcher<BlockEvents> {
     return toolboxItems.find((item) => {
       return Object.entries(item.data)
         .some(([propName, propValue]) => {
-          return blockData[propName] && blockData[propName] === propValue;
+          return blockData[propName] && _.equals(blockData[propName], propValue);
         });
     });
   }
--- a/src/components/block/index.ts
+++ b/src/components/block/index.ts
@@ -54,11 +54,6 @@ interface BlockConstructorOptions {
    * Tunes data for current Block
    */
   tunesData: { [name: string]: BlockTuneData };
-
-  /**
-   * May contain overrides for tool default config
-   */
-  configOverrides: ToolConfig;
 }
 
 /**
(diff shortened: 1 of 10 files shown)


src/components/block/index.ts
```diff
@@ -747,15 +736,45 @@ export default class Block extends EventsDispatcher<BlockEvents> {
...
+     *
+     *  that means that for the current block, the second toolbox item (matched by "{level: 2}") is active
+     */
+    const blockData = await this.data;
+    const toolboxItems = toolboxSettings as ToolboxConfig[];
+
+    return toolboxItems.find((item) => {
+      return Object.entries(item.data)
+        .some(([propName, propValue]) => {
+          return blockData[propName] && blockData[propName] === propValue;
```
TatianaFomina: can data have nested objects or arrays? or it's always flat?

src/components/modules/toolbar/inline.ts
```diff
@@ -480,7 +480,7 @@ export default class InlineToolbar extends Module<InlineToolbarNodes> {
     /**
      * Get icon or title for dropdown
      */
-    const toolboxSettings = currentBlock.activeToolboxEntry || {};
+    const toolboxSettings = await currentBlock.getActiveToolboxEntry() || {};
```
TatianaFomina, line 483: there's one problem with setting toggler's content this way: if none of the configured toolbox entries is active, toggler will contain only tool name

https://user-images.githubusercontent.com/31101125/168940401-759ee229-ac7c-4513-bbc0-5d37012bee1c.mov

TatianaFomina (commented):
Got several comments but other than that looks like a big improvement

src/components/modules/toolbar/inline.ts
neSpecc: I think, it is ok because other titles and icons does not correctly fit here.

TatianaFomina (approved):

PR #2091, opened 2022-07-05
Block tunes as a popover

TatianaFomina:
- Updates block tunes to use popover component. As a temporary solution, non-default block tunes are rendered as before and inserted to special area inside the popover. 
- New getter `blockSettings` (naming could be discussed) is added to tunes interface. It allows to configure the way tune is displayed in the popover. 
image

### DIFF
--- a/src/components/block-tunes/block-tune-delete.ts
+++ b/src/components/block-tunes/block-tune-delete.ts
@@ -53,6 +53,7 @@ export default class DeleteTune implements BlockTune {
       confirmation: {
         icon: $.svg('cross', 14, 14).outerHTML,
         label: 'Click to delete',
+        name: 'delete',
         onActivate: (item, e): void => this.handleClick(e),
       },
     };
--- a/src/components/block-tunes/block-tune-delete.ts
+++ b/src/components/block-tunes/block-tune-delete.ts
@@ -84,31 +84,6 @@ export default class DeleteTune implements BlockTune {
    * @param {MouseEvent} event - click event
    */
   public handleClick(event: MouseEvent): void {
-    /**
-     * if block is not waiting the confirmation, subscribe on block-settings-closing event to reset
-     * otherwise delete block
-     */
-    // if (!this.needConfirmation) {
-    //   this.setConfirmation(true);
-    //   const button = (event.target as HTMLElement)
-    //     .closest('.' + Popover.CSS.item)
-    //     .querySelector('.' + Popover.CSS.itemIcon);
-
-    //   button.classList.add(this.CSS.buttonDelete);
-    //   button.classList.add(this.CSS.buttonConfirm);
-
-    //   /**
-    //    * Subscribe on event.
-    //    * When toolbar block settings is closed but block deletion is not confirmed,
-    //    * then reset confirmation state
-    //    */
-    //   this.api.events.on('block-settings-closed', this.resetConfirmation);
-    // } else {
-    /**
-     * Unsubscribe from block-settings closing event
-     */
-    // this.api.events.off('block-settings-closed', this.resetConfirmation);
-
     this.api.blocks.delete();
     this.api.toolbar.close();
     this.api.tooltip.hide();
@@ -117,7 +92,6 @@ export default class DeleteTune implements BlockTune {
      * Prevent firing ui~documentClicked that can drop currentBlock pointer
      */
     event.stopPropagation();
-    // }
   }
 
   /**
--- a/src/components/block-tunes/block-tune-delete.ts
+++ b/src/components/block-tunes/block-tune-delete.ts
@@ -51,9 +51,7 @@ export default class DeleteTune implements BlockTune {
       label: this.api.i18n.t('Delete'),
       name: 'delete',
       confirmation: {
-        icon: $.svg('cross', 14, 14).outerHTML,
-        label: 'Click to delete',
-        name: 'delete',
+        label: this.api.i18n.t('Click to delete'),
         onActivate: (item, e): void => this.handleClick(e),
       },
     };
--- a/src/components/block-tunes/block-tune-delete.ts
+++ b/src/components/block-tunes/block-tune-delete.ts
@@ -6,7 +6,7 @@
  */
 import { API, BlockTune } from '../../../types';
 import $ from '../dom';
-import { PopoverItem } from '../utils/popover';
+import Popover, { PopoverItem } from '../utils/popover';
 
 /**
  *
@@ -64,27 +64,7 @@ export default class DeleteTune implements BlockTune {
   }
 
   /**
-   * Create "Delete" button and add click event listener
-   *
-   * @returns {HTMLElement}
-   */
-  public render(): HTMLElement {
-    this.nodes.button = $.make('div', [this.CSS.button, this.CSS.buttonDelete], {});
-    this.nodes.button.appendChild($.svg('cross', 12, 12));
-    this.api.listeners.on(this.nodes.button, 'click', (event: MouseEvent) => this.handleClick(event), false);
-
-    /**
-     * Enable tooltip module
-     */
-    this.api.tooltip.onHover(this.nodes.button, this.api.i18n.t('Delete'), {
-      hidingDelay: 300,
-    });
-
-    return this.nodes.button;
-  }
-
-  /**
-   *
+   * Tune's appearance in block settings menu
    */
   public get blockSettings(): PopoverItem {
     return {
@@ -107,7 +87,9 @@ export default class DeleteTune implements BlockTune {
      */
     if (!this.needConfirmation) {
       this.setConfirmation(true);
-      const button = (event.target as HTMLElement).closest('.ce-popover__item').querySelector('.ce-popover__item-icon');
+      const button = (event.target as HTMLElement)
+        .closest('.' + Popover.CSS.item)
+        .querySelector('.' + Popover.CSS.itemIcon);
 
       button.classList.add(this.CSS.buttonDelete);
       button.classList.add(this.CSS.buttonConfirm);
@@ -142,6 +124,5 @@ export default class DeleteTune implements BlockTune {
    */
   private setConfirmation(state: boolean): void {
     this.needConfirmation = state;
-    // this.nodes.button.classList.add(this.CSS.buttonConfirm);
   }
 }
--- a/src/components/block-tunes/block-tune-delete.ts
+++ b/src/components/block-tunes/block-tune-delete.ts
@@ -6,7 +6,7 @@
  */
 import { API, BlockTune } from '../../../types';
 import $ from '../dom';
-import Popover, { PopoverItem } from '../utils/popover';
+import { PopoverItem } from '../utils/popover';
 
 /**
  *
--- a/src/components/block-tunes/block-tune-delete.ts
+++ b/src/components/block-tunes/block-tune-delete.ts
@@ -91,6 +91,7 @@ export default class DeleteTune implements BlockTune {
       icon: $.svg('cross', 14, 14).outerHTML,
       label: this.api.i18n.t('Delete'),
       onClick: (item, e): void => this.handleClick(e),
+      name: 'delete',
     };
   }
 
--- a/src/components/block-tunes/block-tune-delete.ts
+++ b/src/components/block-tunes/block-tune-delete.ts
@@ -4,9 +4,8 @@
  *
  * @copyright <CodeX Team> 2018
  */
-import { API, BlockTune } from '../../../types';
+import { API, BlockTune, PopoverItem } from '../../../types';
 import $ from '../dom';
-import { PopoverItem } from '../utils/popover';
 
 /**
  *
--- a/src/components/block-tunes/block-tune-delete.ts
+++ b/src/components/block-tunes/block-tune-delete.ts
@@ -24,15 +24,6 @@ export default class DeleteTune implements BlockTune {
   private readonly api: API;
 
   /**
-   * Styles
-   */
-  private CSS = {
-    button: 'ce-settings__button',
-    buttonDelete: 'ce-settings__button--delete',
-    buttonConfirm: 'ce-settings__button--confirm',
-  };
-
-  /**
    * DeleteTune constructor
    *
    * @param {API} api - Editor's API
@@ -63,12 +54,5 @@ export default class DeleteTune implements BlockTune {
    */
   public handleClick(event: MouseEvent): void {
     this.api.blocks.delete();
-    this.api.toolbar.close();
-    this.api.tooltip.hide();
-
-    /**
-     * Prevent firing ui~documentClicked that can drop currentBlock pointer
-     */
-    event.stopPropagation();
   }
 }
--- a/src/components/block-tunes/block-tune-delete.ts
+++ b/src/components/block-tunes/block-tune-delete.ts
@@ -53,7 +53,7 @@ export default class DeleteTune implements BlockTune {
       confirmation: {
         icon: $.svg('cross', 14, 14).outerHTML,
         label: 'Click to delete',
-        onClick: (item, e): void => this.handleClick(e),
+        onActivate: (item, e): void => this.handleClick(e),
       },
     };
   }
--- a/src/components/block-tunes/block-tune-delete.ts
+++ b/src/components/block-tunes/block-tune-delete.ts
@@ -66,7 +66,7 @@ export default class DeleteTune implements BlockTune {
   /**
    * Tune's appearance in block settings menu
    */
-  public get blockSettings(): PopoverItem {
+  public render(): PopoverItem {
     return {
       icon: $.svg('cross', 14, 14).outerHTML,
       label: this.api.i18n.t('Delete'),
--- a/src/components/block-tunes/block-tune-delete.ts
+++ b/src/components/block-tunes/block-tune-delete.ts
@@ -72,6 +72,9 @@ export default class DeleteTune implements BlockTune {
       label: this.api.i18n.t('Delete'),
       onClick: (item, e): void => this.handleClick(e),
       name: 'delete',
+      confirmation: {
+        label: 'Click to delete',
+      },
     };
   }
 
@@ -85,36 +88,36 @@ export default class DeleteTune implements BlockTune {
      * if block is not waiting the confirmation, subscribe on block-settings-closing event to reset
      * otherwise delete block
      */
-    if (!this.needConfirmation) {
-      this.setConfirmation(true);
-      const button = (event.target as HTMLElement)
-        .closest('.' + Popover.CSS.item)
-        .querySelector('.' + Popover.CSS.itemIcon);
+    // if (!this.needConfirmation) {
+    //   this.setConfirmation(true);
+    //   const button = (event.target as HTMLElement)
+    //     .closest('.' + Popover.CSS.item)
+    //     .querySelector('.' + Popover.CSS.itemIcon);
 
-      button.classList.add(this.CSS.buttonDelete);
-      button.classList.add(this.CSS.buttonConfirm);
+    //   button.classList.add(this.CSS.buttonDelete);
+    //   button.classList.add(this.CSS.buttonConfirm);
 
-      /**
-       * Subscribe on event.
-       * When toolbar block settings is closed but block deletion is not confirmed,
-       * then reset confirmation state
-       */
-      this.api.events.on('block-settings-closed', this.resetConfirmation);
-    } else {
-      /**
-       * Unsubscribe from block-settings closing event
-       */
-      this.api.events.off('block-settings-closed', this.resetConfirmation);
+    //   /**
+    //    * Subscribe on event.
+    //    * When toolbar block settings is closed but block deletion is not confirmed,
+    //    * then reset confirmation state
+    //    */
+    //   this.api.events.on('block-settings-closed', this.resetConfirmation);
+    // } else {
+    /**
+     * Unsubscribe from block-settings closing event
+     */
+    // this.api.events.off('block-settings-closed', this.resetConfirmation);
 
-      this.api.blocks.delete();
-      this.api.toolbar.close();
-      this.api.tooltip.hide();
+    this.api.blocks.delete();
+    this.api.toolbar.close();
+    this.api.tooltip.hide();
 
-      /**
-       * Prevent firing ui~documentClicked that can drop currentBlock pointer
-       */
-      event.stopPropagation();
-    }
+    /**
+     * Prevent firing ui~documentClicked that can drop currentBlock pointer
+     */
+    event.stopPropagation();
+    // }
   }
 
   /**
--- a/src/components/block-tunes/block-tune-delete.ts
+++ b/src/components/block-tunes/block-tune-delete.ts
@@ -6,6 +6,7 @@
  */
 import { API, BlockTune } from '../../../types';
 import $ from '../dom';
+import { PopoverItem } from '../utils/popover';
 
 /**
  *
@@ -83,6 +84,17 @@ export default class DeleteTune implements BlockTune {
   }
 
   /**
+   *
+   */
+  public get blockSettings(): PopoverItem {
+    return {
+      icon: $.svg('cross', 14, 14).outerHTML,
+      label: this.api.i18n.t('Delete'),
+      onClick: (item, e): void => this.handleClick(e),
+    };
+  }
+
+  /**
    * Delete block conditions passed
    *
    * @param {MouseEvent} event - click event
@@ -94,6 +106,10 @@ export default class DeleteTune implements BlockTune {
      */
     if (!this.needConfirmation) {
       this.setConfirmation(true);
+      const button = (event.target as HTMLElement).closest('.ce-popover__item').querySelector('.ce-popover__item-icon');
+
+      button.classList.add(this.CSS.buttonDelete);
+      button.classList.add(this.CSS.buttonConfirm);
 
       /**
        * Subscribe on event.
--- a/src/components/block-tunes/block-tune-delete.ts
+++ b/src/components/block-tunes/block-tune-delete.ts
@@ -141,6 +141,6 @@ export default class DeleteTune implements BlockTune {
    */
   private setConfirmation(state: boolean): void {
     this.needConfirmation = state;
-    this.nodes.button.classList.add(this.CSS.buttonConfirm);
+    // this.nodes.button.classList.add(this.CSS.buttonConfirm);
   }
 }
--- a/src/components/block-tunes/block-tune-delete.ts
+++ b/src/components/block-tunes/block-tune-delete.ts
@@ -34,33 +34,12 @@ export default class DeleteTune implements BlockTune {
   };
 
   /**
-   * Delete confirmation
-   */
-  private needConfirmation: boolean;
-
-  /**
-   * set false confirmation state
-   */
-  private readonly resetConfirmation: () => void;
-
-  /**
-   * Tune nodes
-   */
-  private nodes: {button: HTMLElement} = {
-    button: null,
-  };
-
-  /**
    * DeleteTune constructor
    *
    * @param {API} api - Editor's API
    */
   constructor({ api }) {
     this.api = api;
-
-    this.resetConfirmation = (): void => {
-      this.setConfirmation(false);
-    };
   }
 
   /**
@@ -70,10 +49,11 @@ export default class DeleteTune implements BlockTune {
     return {
       icon: $.svg('cross', 14, 14).outerHTML,
       label: this.api.i18n.t('Delete'),
-      onClick: (item, e): void => this.handleClick(e),
       name: 'delete',
       confirmation: {
+        icon: $.svg('cross', 14, 14).outerHTML,
         label: 'Click to delete',
+        onClick: (item, e): void => this.handleClick(e),
       },
     };
   }
@@ -92,14 +72,5 @@ export default class DeleteTune implements BlockTune {
      * Prevent firing ui~documentClicked that can drop currentBlock pointer
      */
     event.stopPropagation();
-  }
-
-  /**
-   * change tune state
-   *
-   * @param {boolean} state - delete confirmation state
-   */
-  private setConfirmation(state: boolean): void {
-    this.needConfirmation = state;
   }
 }
--- a/src/components/block-tunes/block-tune-move-up.ts
+++ b/src/components/block-tunes/block-tune-move-up.ts
@@ -5,8 +5,8 @@
  * @copyright <CodeX Team> 2018
  */
 import $ from '../dom';
-import { API, BlockTune, BlockAPI } from '../../../types';
-import Popover, { PopoverItem } from '../../components/utils/popover';
+import { API, BlockTune, BlockAPI, PopoverItem } from '../../../types';
+import Popover from '../../components/utils/popover';
 
 /**
  *
--- a/src/components/block-tunes/block-tune-move-up.ts
+++ b/src/components/block-tunes/block-tune-move-up.ts
@@ -51,7 +51,7 @@ export default class MoveUpTune implements BlockTune {
     return {
       icon: $.svg('arrow-up', 14, 14).outerHTML,
       label: this.api.i18n.t('Move up'),
-      onClick: (item, e): void => this.handleClick(e),
+      onActivate: (item, e): void => this.handleClick(e),
       name: 'move-up',
     };
   }
--- a/src/components/block-tunes/block-tune-move-up.ts
+++ b/src/components/block-tunes/block-tune-move-up.ts
@@ -47,7 +47,7 @@ export default class MoveUpTune implements BlockTune {
   /**
    * Tune's appearance in block settings menu
    */
-  public get blockSettings(): PopoverItem {
+  public render(): PopoverItem {
     return {
       icon: $.svg('arrow-up', 14, 14).outerHTML,
       label: this.api.i18n.t('Move up'),
--- a/src/components/block-tunes/block-tune-move-up.ts
+++ b/src/components/block-tunes/block-tune-move-up.ts
@@ -26,12 +26,8 @@ export default class MoveUpTune implements BlockTune {
 
   /**
    * Styles
-   *
-   * @type {{wrapper: string}}
    */
   private CSS = {
-    button: 'ce-settings__button',
-    wrapper: 'ce-tune-move-up',
     animation: 'wobble',
   };
 
@@ -108,8 +104,5 @@ export default class MoveUpTune implements BlockTune {
     this.api.blocks.move(currentBlockIndex - 1);
 
     this.api.toolbar.toggleBlockSettings(true);
-
-    /** Hide the Tooltip */
-    this.api.tooltip.hide();
   }
 }
--- a/src/components/block-tunes/block-tune-move-up.ts
+++ b/src/components/block-tunes/block-tune-move-up.ts
@@ -77,7 +77,7 @@ export default class MoveUpTune implements BlockTune {
     return {
       icon: $.svg('arrow-up', 14, 14).outerHTML,
       label: this.api.i18n.t('Move up'),
-      onClick: (item): void => this.handleClick(),
+      onClick: (item, e): void => this.handleClick(e),
     };
   }
 
@@ -99,7 +99,7 @@ export default class MoveUpTune implements BlockTune {
       //   button.classList.remove(this.CSS.animation);
       // }, 500);
 
-      // return;
+      return;
     }
 
     const currentBlockElement = (currentBlock as BlockAPI).holder;
--- a/src/components/block-tunes/block-tune-move-up.ts
+++ b/src/components/block-tunes/block-tune-move-up.ts
@@ -76,8 +76,8 @@ export default class MoveUpTune implements BlockTune {
       return;
     }
 
-    const currentBlockElement = (currentBlock as BlockAPI).holder;
-    const previousBlockElement = (previousBlock as BlockAPI).holder;
+    const currentBlockElement = currentBlock.holder;
+    const previousBlockElement = previousBlock.holder;
 
     /**
      * Here is two cases:
--- a/src/components/block-tunes/block-tune-move-up.ts
+++ b/src/components/block-tunes/block-tune-move-up.ts
@@ -78,6 +78,7 @@ export default class MoveUpTune implements BlockTune {
       icon: $.svg('arrow-up', 14, 14).outerHTML,
       label: this.api.i18n.t('Move up'),
       onClick: (item, e): void => this.handleClick(e),
+      name: 'move-up',
     };
   }
 
--- a/src/components/block-tunes/block-tune-move-up.ts
+++ b/src/components/block-tunes/block-tune-move-up.ts
@@ -5,7 +5,8 @@
  * @copyright <CodeX Team> 2018
  */
 import $ from '../dom';
-import { API, BlockTune } from '../../../types';
+import { API, BlockTune, BlockAPI } from '../../../types';
+import { PopoverItem } from '../../components/utils/popover';
 
 /**
  *
@@ -70,28 +71,39 @@ export default class MoveUpTune implements BlockTune {
   }
 
   /**
+   *
+   */
+  public get blockSettings(): PopoverItem {
+    return {
+      icon: $.svg('arrow-up', 14, 14).outerHTML,
+      label: this.api.i18n.t('Move up'),
+      onClick: (item): void => this.handleClick(),
+    };
+  }
+
+  /**
    * Move current block up
    *
    * @param {MouseEvent} event - click event
    * @param {HTMLElement} button - clicked button
    */
-  public handleClick(event: MouseEvent, button: HTMLElement): void {
+  public handleClick(event?: MouseEvent, button?: HTMLElement): void {
     const currentBlockIndex = this.api.blocks.getCurrentBlockIndex();
     const currentBlock = this.api.blocks.getBlockByIndex(currentBlockIndex);
     const previousBlock = this.api.blocks.getBlockByIndex(currentBlockIndex - 1);
 
     if (currentBlockIndex === 0 || !currentBlock || !previousBlock) {
-      button.classList.add(this.CSS.animation);
+      // button.classList.add(this.CSS.animation);
 
-      window.setTimeout(() => {
-        button.classList.remove(this.CSS.animation);
-      }, 500);
+      // window.setTimeout(() => {
+      //   button.classList.remove(this.CSS.animation);
+      // }, 500);
 
-      return;
+      // return;
     }
 
-    const currentBlockElement = currentBlock.holder;
-    const previousBlockElement = previousBlock.holder;
+    const currentBlockElement = (currentBlock as BlockAPI).holder;
+    const previousBlockElement = (previousBlock as BlockAPI).holder;
 
     /**
      * Here is two cases:
--- a/src/components/block-tunes/block-tune-move-up.ts
+++ b/src/components/block-tunes/block-tune-move-up.ts
@@ -6,7 +6,7 @@
  */
 import $ from '../dom';
 import { API, BlockTune, BlockAPI } from '../../../types';
-import { PopoverItem } from '../../components/utils/popover';
+import Popover, { PopoverItem } from '../../components/utils/popover';
 
 /**
  *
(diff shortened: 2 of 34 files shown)


src/components/block-tunes/block-tune-delete.ts
```diff
@@ -63,23 +64,15 @@ export default class DeleteTune implements BlockTune {
...
-
-    /**
-     * Enable tooltip module
-     */
-    this.api.tooltip.onHover(this.nodes.button, this.api.i18n.t('Delete'), {
-      hidingDelay: 300,
-    });
-
-    return this.nodes.button;
+  public get blockSettings(): PopoverItem {
```
neSpecc: Dont you think, that we could use the existed `render` method? Just extend possible return types

src/components/block/index.ts
```diff
@@ -642,22 +643,15 @@ export default class Block extends EventsDispatcher<BlockEvents> {
...
-      $.append(defaultTunesElement, tune.render());
-    });
+  public getTunesList(): PopoverItem[] {
+    const tunes = Array.from(this.tunesInstances.values());
+    const defaultTunes = Array.from(this.defaultTunesInstances.values());
 
-    return [tunesElement, defaultTunesElement];
+    return tunes.concat(defaultTunes)
+      .map(tune => tune.blockSettings)
+      .filter(item => !!item);
```
neSpecc: pls add comments describing this case

src/components/modules/toolbar/blockSettings.ts
```diff
@@ -144,24 +97,41 @@ export default class BlockSettings extends Module<BlockSettingsNodes> {
...
     /** Tell to subscribers that block settings is opened */
     this.eventsDispatcher.emit(this.events.opened);
 
-    this.flipper.activate(this.blockTunesButtons);
+    this.makeToolTunesButtonsNavigatable();
+
+    this.popover = new Popover({
+      className: this.CSS.settings,
+      searchable: true,
+      filterLabel: I18n.ui(I18nInternalNS.ui.toolbar.toolbox, 'Filter'),
```
neSpecc: maybe these translations should be moved to the `I18nInternalNS.ui.popover`?

src/components/modules/toolbar/blockSettings.ts
```diff
@@ -184,105 +154,44 @@ export default class BlockSettings extends Module<BlockSettingsNodes> {
...
-        /**
-         * Restoring focus on current Block after settings clicked.
-         * For example, when H3 changed to H2 — DOM Elements replaced, so we need to focus a new one
-         */
-        _.delay(() => {
-          this.Editor.Caret.setToBlock(this.Editor.BlockManager.currentBlock);
-        }, 50)();
-      },
-    } as FlipperOptions);
+  private onOverlayClicked = (): void => {
```
neSpecc, line 193:
```suggestion
  private onOverlayClicked(): void {
```

src/components/utils/popover.ts
```diff
@@ -340,11 +380,18 @@ export default class Popover extends EventsDispatcher<PopoverEvent> {
 
         this.nodes.nothingFound.classList.toggle(Popover.CSS.noFoundMessageShown, itemsVisible.length === 0);
 
+        const allItemsDisdplayed = filteredItems.length === this.items.length;
```
neSpecc:
```suggestion
        const allItemsDisplayed = filteredItems.length === this.items.length;
```

src/components/block/index.ts
```diff
@@ -642,22 +643,15 @@ export default class Block extends EventsDispatcher<BlockEvents> {
...
-    this.defaultTunesInstances.forEach((tune) => {
-      $.append(defaultTunesElement, tune.render());
-    });
+  public getTunesList(): PopoverItem[] {
+    const tunes = Array.from(this.tunesInstances.values());
+    const defaultTunes = Array.from(this.defaultTunesInstances.values());
 
-    return [tunesElement, defaultTunesElement];
+    return tunes.concat(defaultTunes)
+      .map(tune => tune.blockSettings)
```
neSpecc: so you have removed the `render()` support at all?

src/components/block-tunes/block-tune-delete.ts
TatianaFomina: i think method `render` is not suitable for such role. There's no rendering, we just need to return a popover entry configuration

neSpecc: Agree. So we had to add the new API method and explain it to the world. Ok.

But I don't like the `blockTune.blockSettins` property — it makes no sense. Maybe `toolbox`? Just like at the Tools API

src/components/modules/toolbar/blockSettings.ts
TatianaFomina: arrow function is used on purpose to prevent loss of `this`

src/components/block-tunes/block-tune-delete.ts
```diff
@@ -63,23 +64,15 @@ export default class DeleteTune implements BlockTune {
...
-
-    /**
-     * Enable tooltip module
-     */
-    this.api.tooltip.onHover(this.nodes.button, this.api.i18n.t('Delete'), {
-      hidingDelay: 300,
-    });
-
-    return this.nodes.button;
+  public render(): PopoverItem {
```
neSpecc, line 38: why it is called `render`? I agreed with your previous sentence:

> i think method render is not suitable for such role. There's no rendering, we just need to return a popover entry configuration

I'm suggesting to name it `get toolbox()` just like at the Tools API. But there is also a difference: this getter will be non-static. Maybe it will cause misunderstanings. What do you think?

src/components/block-tunes/block-tune-delete.ts
```diff
@@ -94,6 +87,12 @@ export default class DeleteTune implements BlockTune {
      */
     if (!this.needConfirmation) {
       this.setConfirmation(true);
+      const button = (event.target as HTMLElement)
```
neSpecc: i think the action confirmation should be handled by the Popover. Maybe we should add a property to the `PopoverItem`?

```js
{
  name: 'delete',
  icon: $.svg('cross', 14, 14).outerHTML,
  label: this.api.i18n.t('Delete'),
  onClick: (item, e): void => this.handleClick(e),
  requireConfirmation: true. // <----- here is it
}
```

Of course, we can do it via a separate PR, if you want. But leave a `todo` for that.

src/components/block/index.ts
```diff
@@ -642,22 +644,36 @@ export default class Block extends EventsDispatcher<BlockEvents> {
...
+    let tunesItems: PopoverItem[] = [];
 
-    this.tunesInstances.forEach((tune) => {
-      $.append(tunesElement, tune.render());
-    });
-    this.defaultTunesInstances.forEach((tune) => {
-      $.append(defaultTunesElement, tune.render());
+    const defaultTunesInstances = Array.from(this.defaultTunesInstances.values());
+    const customTunesInstances = Array.from(this.tunesInstances.values());
+    const tunesDefinedInTool =
```
neSpecc: pls add a comment there since it is not well readable

src/components/block/index.ts
```diff
@@ -642,22 +644,36 @@ export default class Block extends EventsDispatcher<BlockEvents> {
...
+      typeof this.toolInstance.renderSettings === 'function'
+        ? [ this.toolInstance.renderSettings() ].flat()
+        : [];
+    const otherTunes = customTunesInstances.concat(defaultTunesInstances).map(item => item.render());
```
neSpecc:
```suggestion
    const otherTunes = customTunesInstances
      .concat(defaultTunesInstances)
      .map(item => item.render());
```

neSpecc: naming is not correct since there is a combination of custom tunes and default tunes as I can see.

src/components/block/index.ts
```diff
@@ -642,22 +644,36 @@ export default class Block extends EventsDispatcher<BlockEvents> {
...
+
+    tunesDefinedInTool.concat(otherTunes).forEach(rendered => {
```
neSpecc:
```suggestion
    tunesDefinedInTool
      .concat(otherTunes)
      .forEach(rendered => {
```

src/components/block/index.ts
```diff
@@ -642,22 +644,36 @@ export default class Block extends EventsDispatcher<BlockEvents> {
...
+      const isHTMLElement = rendered.nodeName !== undefined;
+
+      if (isHTMLElement) {
+        tunesElement.appendChild(rendered);
+      } else {
+        const normalized = [ rendered ].flat();
```
neSpecc: It's hard to understand this line

src/components/modules/toolbar/blockSettings.ts
```diff
@@ -184,105 +157,61 @@ export default class BlockSettings extends Module<BlockSettingsNodes> {
...
-   * @param targetBlock - Block to render settings
+   * Adds special class to rendered tool settings signalising that button should
+   * be available for keyboard navigation
    */
-  private addToolSettings(targetBlock): void {
-    const settingsElement = targetBlock.renderSettings();
+  private makeToolTunesButtonsNavigatable(): void {
+    const { StylesAPI } = this.Editor;
+    const toolSettings = this.nodes.renderedTunes?.querySelectorAll(
+      // Select buttons and inputs
```
neSpecc: please, add more detailed description

src/components/utils/popover.ts
```diff
@@ -414,4 +467,31 @@ export default class Popover extends EventsDispatcher<PopoverEvent> {
       ],
     });
   }
+
+  /**
+   * Returns list of items available for keyboard navigation
+   */
+  private get flippableItems(): HTMLElement[] {
```
neSpecc: please, improve the readability of this method

src/components/utils/popover.ts
```diff
@@ -414,4 +467,31 @@ export default class Popover extends EventsDispatcher<PopoverEvent> {
...
+    return customItems.concat(popoverItems);
+  }
+
+  /**
+   * Checks if there popover should be opened up.
+   * It happens in case there is enough space below or not enough space above
+   */
+  private get shouldOpenPopoverBottom(): boolean {
+    const toolboxRect = this.nodes.wrapper.getBoundingClientRect();
+    const editorElementRect = this.api.ui.nodes.redactor.getBoundingClientRect();
```
neSpecc: You're depends on API only for accessing this element? I thinks, it is useless coheision. It would be better to pass that element to the constructor.

src/components/utils/popover.ts
```diff
@@ -340,11 +385,18 @@ export default class Popover extends EventsDispatcher<PopoverEvent> {
 
         this.nodes.nothingFound.classList.toggle(Popover.CSS.noFoundMessageShown, itemsVisible.length === 0);
 
+        const allItemsDisplayed = filteredItems.length === this.items.length;
+        const flippableItems = allItemsDisplayed ? this.flippableItems : itemsVisible;
```
neSpecc: Please, add a description. 

Why not to always use `itemsVisible`? What is the difference between the `this.flippableItems` (that are "displayed") and the `itemsVisible`? Both parts of the statements looks like the same ("displayed" = "visible").

types/block-tunes/block-tune.d.ts
```diff
@@ -1,16 +1,16 @@
 import {API, BlockAPI, SanitizerConfig, ToolConfig} from '../index';
 import { BlockTuneData } from './block-tune-data';
+import { PopoverItem } from '../../src/components/utils/popover';
```
neSpecc: seems unused

neSpecc (changes requested):

src/components/block-tunes/block-tune-delete.ts
TatianaFomina: i don't see why popover should contain logic of confirmation. 
1) popover is just a simple component for displaying items - the least logic it has inside, the better
2) confirmation feature does not seem to be that commonly used to have global support

src/components/block-tunes/block-tune-delete.ts
TatianaFomina: I changed my mind regarding the need of separate getter for popover items. I decided that tool should not have 2 ways of configuring its appearance in tunes menu at the same time. 
I happened because i could not resolve the question how to handle situation when tool has `renderSettings()` returning html and `toolbox()` returning array of popover items at once

types/tools/tool-settings.d.ts
```diff
@@ -27,6 +28,28 @@ export interface ToolboxConfigEntry {
   data?: BlockToolData
 }
 
+/**
+ * Represents single tunes menu entry
+ */
+export type TunesMenuEntry = PopoverItem & {
```
neSpecc: move new props to the popover item

types/tools/block-tool.d.ts
```diff
@@ -97,6 +98,11 @@ export interface BlockToolConstructable extends BaseToolConstructable {
    */
   toolbox?: ToolboxConfig;
 
+  /**
+   * Configures the way tool's tunes appear in tunes menu
+   */
+  tunes?: TunesMenuConfig;
```
neSpecc: is it actual?

src/components/block/index.ts
TatianaFomina: added comments

src/components/utils/popover.ts
TatianaFomina: `flippableItems` contains not only popover items that were filtered during search, but also buttons from custom html area -- this allows to navigate via keyboard through all the popover content, not only items configured with json. Anyway, added some comments. Hope, they will make everything a little bit more clear

src/components/block-tunes/block-tune-delete.ts
```diff
@@ -32,54 +33,29 @@ export default class DeleteTune implements BlockTune {
...
-    });
-
-    return this.nodes.button;
+  public render(): PopoverItem {
+    return {
+      icon: $.svg('cross', 14, 14).outerHTML,
+      label: this.api.i18n.t('Delete'),
+      name: 'delete',
+      confirmation: {
+        icon: $.svg('cross', 14, 14).outerHTML,
```
neSpecc: Lets make icon optional to allow skipping icon if it is not changed

src/components/block-tunes/block-tune-delete.ts
```diff
@@ -32,54 +33,29 @@ export default class DeleteTune implements BlockTune {
...
+        label: 'Click to delete',
```
neSpecc: i18n missed

src/components/modules/toolbar/blockSettings.ts
```diff
@@ -42,82 +35,42 @@ export default class BlockSettings extends Module<BlockSettingsNodes> {
...
-    this.nodes.defaultSettings = $.make('div', this.CSS.defaultSettings);
-
-    $.append(this.nodes.wrapper, [this.nodes.toolSettings, this.nodes.defaultSettings]);
-
-    /**
-     * Active leafing by arrows/tab
-     * Buttons will be filled on opening
-     */
-    this.enableFlipper();
+    this.nodes.wrapper = $.make('div');
```
neSpecc, line 75: why do we need a div without any classname? Could we get rid of it?

src/components/block/index.ts
```diff
@@ -642,22 +643,45 @@ export default class Block extends EventsDispatcher<BlockEvents> {
...
-      $.append(defaultTunesElement, tune.render());
+    /** Default tunes: Move up, Move down, Delete */
+    const defaultTunesInstances = Array.from(this.defaultTunesInstances.values());
+
+    /** Third-party tunes */
+    const customTunesInstances = Array.from(this.tunesInstances.values());
+
+    /** Retrieve tunes that may be defined in tool as return value of optional renderSettings function. */
+    const tunesDefinedInTool =
+      typeof this.toolInstance.renderSettings === 'function'
```
neSpecc: Could we add an optional support of the `PopoverItem` format for the `renderSettings()` as well?

src/components/block/index.ts
```diff
@@ -642,22 +643,45 @@ export default class Block extends EventsDispatcher<BlockEvents> {
...
+        ? [ this.toolInstance.renderSettings() ].flat()
```
neSpecc: why do we use `[].flat()` here? Please, explain in comments

src/components/block/index.ts
```diff
@@ -642,22 +643,45 @@ export default class Block extends EventsDispatcher<BlockEvents> {
...
+        : [];
+
+    /** Combine default and thirid-party tunes and call render() on each to retrieve their controls */
+    const otherTunes = customTunesInstances
```
neSpecc:
```suggestion
    const commonTunes = customTunesInstances
```

src/components/block/index.ts
```diff
@@ -642,22 +643,45 @@ export default class Block extends EventsDispatcher<BlockEvents> {
...
+      .concat(defaultTunesInstances)
+      .map(item => item.render());
+
+    tunesDefinedInTool.concat(otherTunes).forEach(rendered => {
+      const isHTMLElement = rendered.nodeName !== undefined;
```
neSpecc: Could we use the existed `dom.isElement()` util here?

src/components/block/index.ts
```diff
@@ -642,22 +643,45 @@ export default class Block extends EventsDispatcher<BlockEvents> {
...
+
+      if (isHTMLElement) {
+        tunesElement.appendChild(rendered);
+      } else {
+        const items = Array.isArray(rendered) ? rendered : [ rendered ];
+
+        tunesItems.push(...items);
+      }
```
neSpecc: We need to cover these 3 cases with tests at least.

src/components/domIterator.ts
```diff
@@ -60,6 +60,18 @@ export default class DomIterator {
     return this.items[this.cursor];
   }
 
+  /**
+   * Sets cursor to specified position
+   *
+   * @param cursorPosition - new cursor position
+   */
+  public setCursor(cursorPosition: number): void {
+    if (cursorPosition < this.items.length && cursorPosition >= -1) {
```
neSpecc, line 69: do we need to `dropCursor()` first?

src/components/modules/toolbar/blockSettings.ts
```diff
@@ -144,24 +97,43 @@ export default class BlockSettings extends Module<BlockSettingsNodes> {
     /**
      * Fill Tool's settings
      */
-    this.addToolSettings(targetBlock);
+    const [tunesItems, tunesElement] = targetBlock.getTunes();
```
neSpecc:
```suggestion
    const [tunesItemsConfig, addiitionalTunesContainer] = targetBlock.getTunes();
```

src/components/modules/toolbar/blockSettings.ts
```diff
@@ -144,24 +97,43 @@ export default class BlockSettings extends Module<BlockSettingsNodes> {
...
+    this.makeToolTunesButtonsNavigatable();
+
+    this.popover = new Popover({
+      className: this.CSS.settings,
+      searchable: true,
+      filterLabel: I18n.ui(I18nInternalNS.ui.popover, 'Filter'),
+      nothingFoundLabel: I18n.ui(I18nInternalNS.ui.popover, 'Nothing found'),
+      items: tunesItems,
+      customContent: this.nodes.renderedTunes,
+      editorElement: this.Editor.API.methods.ui.nodes.redactor,
```
neSpecc:
```suggestion
      scopeElement: this.Editor.API.methods.ui.nodes.redactor,
```

src/components/modules/toolbar/blockSettings.ts
```diff
@@ -184,105 +156,44 @@ export default class BlockSettings extends Module<BlockSettingsNodes> {
     }
 
     /** Clear settings */
-    this.nodes.toolSettings.innerHTML = '';
-    this.nodes.defaultSettings.innerHTML = '';
+    if (this.nodes.renderedTunes) {
+      this.nodes.renderedTunes.innerHTML = '';
```
neSpecc: maybe the Popover should do that?

src/components/utils/popover.ts
```diff
@@ -7,9 +7,9 @@ import { isMobileScreen, keyCodes, cacheable } from '../utils';
 import ScrollLocker from './scroll-locker';
 
 /**
- * Describe parameters for rendering the single item of Popover
+ * Describe parameters for rendering a single popover item
  */
-export interface PopoverItem {
+export interface PopoverItemBase {
```
neSpecc: Do we really need to export this interface?

src/components/utils/popover.ts
```diff
@@ -32,13 +32,45 @@ export interface PopoverItem {
...
+   */
+  closeOnActivate?: boolean;
+}
+
+/**
+ * Represents popover item with confirmation state configuration
+ */
+interface PopoverItemWithConfirmation extends PopoverItemBase {
+  /**
+   * Item parameters that should be applied on item activation.
```
neSpecc:
```suggestion
   * Popover Item parameters that should be applied on item activation.
```

src/components/utils/popover.ts
```diff
@@ -32,13 +32,45 @@ export interface PopoverItem {
...
+  confirmation?: never;
+
+  /**
+   * Item activation handler
    *
-   * @param item - clicked item
+   * @param item - activated item
    */
-  onClick: (item: PopoverItem) => void;
+  onActivate: (item: PopoverItem, event?: MouseEvent) => void;
```
neSpecc: pls add jsdoc for the new argument

src/components/utils/popover.ts
```diff
@@ -32,13 +32,45 @@ export interface PopoverItem {
...
 }
 
+export type PopoverItem = PopoverItemWithConfirmation | PopoverItemWithoutConfirmation
```
neSpecc: description missed

src/components/block-tunes/block-tune-delete.ts
```diff
@@ -6,6 +6,7 @@
  */
 import { API, BlockTune } from '../../../types';
 import $ from '../dom';
+import Popover, { PopoverItem } from '../utils/popover';
```
neSpecc:
```suggestion
import { PopoverItem } from '../utils/popover';
```

src/components/utils/popover.ts
neSpecc: lets move all the types to the `/types/configs/popover.d.ts` (and export them from types/index.d.ts)

src/components/utils/popover.ts
```diff
@@ -155,6 +202,16 @@ export default class Popover extends EventsDispatcher<PopoverEvent> {
...
 
+  /**
+   * Editor container element
+   */
+  private editorElement: HTMLElement;
+
+  /**
+   * Reference to popover item that was clicked but requires second click to confirm action
+   */
+  private itemRequiringConfirmation = null;
```
neSpecc: Does it mean that we support only the singe confirmation button at the popover? Could we support several buttons in that state at once? Or how to handle that case?

src/components/utils/popover.ts
```diff
@@ -163,19 +220,25 @@ export default class Popover extends EventsDispatcher<PopoverEvent> {
    * @param options.className - additional class name to be added to the popover wrapper
    * @param options.filterLabel - label for the search Field
    * @param options.nothingFoundLabel - label of the 'nothing found' message
+   * @param options.customContent - arbitrary html element to be inserted before items list
+   * @param options.editorElement - editor container element
```
neSpecc:
```suggestion
   * @param options.scopeElement - parent element that scopes the Popover possible positions
```

src/components/utils/popover.ts
```diff
@@ -381,30 +480,92 @@ export default class Popover extends EventsDispatcher<PopoverEvent> {
...
   }
 
   /**
    * Item click handler
    *
    * @param itemEl - clicked item
+   * @param event - click event
    */
-  private itemClicked(itemEl: HTMLElement): void {
+  private itemClicked(itemEl: HTMLElement, event: MouseEvent): void {
```
neSpecc: > MouseEvent

and what about Enter keydowns?

src/components/utils/popover.ts
```diff
@@ -381,30 +480,92 @@ export default class Popover extends EventsDispatcher<PopoverEvent> {
...
     const allItems = this.nodes.wrapper.querySelectorAll(`.${Popover.CSS.item}`);
     const itemIndex = Array.from(allItems).indexOf(itemEl);
     const clickedItem = this.items[itemIndex];
 
-    clickedItem.onClick(clickedItem);
+    this.cleanUpConfirmationState();
```
neSpecc: looks like we need to pass the item from which we need to clean that state

```ts
this.cleanUpConfirmationStateFromItem(clickedItem);
```

src/components/utils/popover.ts
```diff
@@ -381,30 +480,92 @@ export default class Popover extends EventsDispatcher<PopoverEvent> {
...
+      confirmationStateItemEl.classList.add(Popover.CSS.itemConfirmation);
+      itemEl.parentElement.replaceChild(confirmationStateItemEl, itemEl);
+
+      this.reactivateFlipper(
+        this.flippableElements,
+        this.flippableElements.indexOf(confirmationStateItemEl)
+      );
+
+      return;
+    }
```
neSpecc: maybe it would be better to move this code to the `enableConfirmationStateForItem(item)` ?

src/components/utils/popover.ts
```diff
@@ -381,30 +480,92 @@ export default class Popover extends EventsDispatcher<PopoverEvent> {
...
 
-    clickedItem.onClick(clickedItem);
+    this.cleanUpConfirmationState();
+
+    if (clickedItem.confirmation) {
+      /** Save root item requiring confirmation to restore original state on popover hide */
+      if (this.itemRequiringConfirmation === null) {
+        this.itemRequiringConfirmation = clickedItem;
+      }
+      this.items[itemIndex] = clickedItem.confirmation;
```
neSpecc: what if some data was not passed? For example, we override only the `label`, but icon should be preserved. 

We need to merge the confirmatuib config with the root?

src/components/modules/toolbar/blockSettings.ts
```diff
@@ -184,105 +156,44 @@ export default class BlockSettings extends Module<BlockSettingsNodes> {
...
-   *
-   * @param targetBlock - Block to render its Tunes set
-   */
-  private addTunes(targetBlock): void {
-    const [toolTunes, defaultTunes] = targetBlock.renderTunes();
-
-    $.append(this.nodes.toolSettings, toolTunes);
-    $.append(this.nodes.defaultSettings, defaultTunes);
+    toolSettings.forEach((item) => {
+      item.classList.add(Popover.CSS.itemFlippable);
```
neSpecc: That looks strange. Why external dude knows about some internal Popover classname? And then the Popover relies on that external logic inside its internal `flippableElemnts` getter.

If it is a part of the Popover API, it should be presented more explicitly, for example via the public method.

src/styles/popover.css
```diff
@@ -72,6 +72,33 @@
...
+    &--active {
+      @apply --button-active;
+    }
+
+    &--confirmation {
+      background: var(--color-confirm);
+
+      @media (--can-hover) {
+        &:hover {
+          background: #ce4343;
```
neSpecc: please unhardcode the color value

src/styles/variables.css
```diff
@@ -104,6 +104,11 @@
     background: rgba(34, 186, 255, 0.08) !important;
   };
 
+  --button-active: {
+    background: rgba(56, 138, 229, 0.1);
```
neSpecc, line 112: how to see this style?

src/components/block-tunes/block-tune-delete.ts
TatianaFomina: can't remove this import. Popover.CSS is used for animation.

src/components/utils/popover.ts
TatianaFomina: seems like it should be a PointerEvent anyway

image

src/components/utils/popover.ts
TatianaFomina: added merging configs

src/components/utils/popover.ts
TatianaFomina: yep, moved the code to separate method

src/components/block-tunes/block-tune-delete.ts
TatianaFomina: done

src/components/block-tunes/block-tune-delete.ts
TatianaFomina: done

src/components/modules/toolbar/blockSettings.ts
TatianaFomina: not sure, if this line is needed at all

src/components/utils/popover.ts
TatianaFomina: only one popover item at once can be in confirmation state. Once any other element is clicked, confirmation state is disappearing.

src/components/utils/popover.ts
TatianaFomina: no need in parameter, there can only be one item in confirmation state at the same time. Corresponding html element is queried inside the method

src/components/modules/toolbar/blockSettings.ts
TatianaFomina: logic on determining which elements inside custom content of popover are flippable could be moved to popover itself. But to achieve that StylesApi should be available inside popover. It's easy to pass it as constructor param from blockSettings.ts, but popover is also used in toolbox, where StylesAPI is not available and should be passed from outside as well.

What you see was the easiest solution allowing to prevent passing StylesApi ref through chain of instances. Probably, not the best one

src/components/domIterator.ts
TatianaFomina: added

src/components/utils/popover.ts
TatianaFomina: removed export

TatianaFomina: done

src/styles/variables.css
TatianaFomina: it's applied for the popover items with `isActive: true`

(for the screenshot below i made simple-image plugin's renderSettings() return popover items)
image

src/components/block-tunes/block-tune-delete.ts
```diff
@@ -88,43 +62,13 @@ export default class DeleteTune implements BlockTune {
...
-
-  /**
-   * change tune state
-   *
-   * @param {boolean} state - delete confirmation state
-   */
-  private setConfirmation(state: boolean): void {
-    this.needConfirmation = state;
-    this.nodes.button.classList.add(this.CSS.buttonConfirm);
+    event.stopPropagation();
```
neSpecc: do we still need this?

src/components/modules/toolbar/blockSettings.ts
neSpecc: so check it and remove these lines, or move them to the popover.destroy()

src/components/utils/popover.ts
```diff
@@ -381,30 +419,127 @@ export default class Popover extends EventsDispatcher<PopoverEvent> {
...
+    /** Save root item requiring confirmation to restore original state on popover hide */
+    if (this.itemRequiringConfirmation === null) {
+      this.itemRequiringConfirmation = item;
+    }
+    const newItemData = {
+      ...item,
+      ...item.confirmation,
+    } as PopoverItem;
+
+    /** Prevent confirmation loop */
```
neSpecc: I cant get this case, please provide a more detailed comment with examples of config

neSpecc: I've tried such a config, but it throws an error:

```js
{
      icon: $.svg('cross', 14, 14).outerHTML,
      label: this.api.i18n.t('Delete'),
      name: 'delete',
      confirmation: {
        label: this.api.i18n.t('Click to delete'),
        confirmation: {
          label: this.api.i18n.t('Second click to delete'),
          onActivate(){
            console.log('22');
          }
        },
      },
    };
```

image

src/components/utils/popover.ts
```diff
@@ -381,30 +419,127 @@ export default class Popover extends EventsDispatcher<PopoverEvent> {
...
+  /**
+   * Brings popover item in confirmation state to its original state
+   */
+  private cleanUpConfirmationState(): void {
+    if (!this.itemRequiringConfirmation) {
+      return;
+    }
+    const confirmationStateItemEl = this.nodes.wrapper.querySelector(`.${Popover.CSS.itemConfirmation}`);
+    const allItems = this.nodes.wrapper.querySelectorAll(`.${Popover.CSS.item}`);
+    const itemIndex = Array.from(allItems).indexOf(confirmationStateItemEl);
```
neSpecc: Duplicated lines, could be moved to the separate method

src/styles/variables.css
neSpecc: Looks different with the Figma file:

image

1. Remove border
2. Icon border should have light-blue color

src/styles/variables.css
```diff
@@ -104,6 +105,18 @@
...
+    color:  var(--color-active-icon);
+  };
+
+  --button-disabled: {
+    color: var(--grayText);
+    cursor: default;
+    background: rgba(232,232,235,0.49);
```
neSpecc: There is no background in Figma:

image

types/configs/popover.d.ts
```diff
@@ -0,0 +1,75 @@
...
+  /**
+   * Displayed text
+   */
+  label: string;
+
+  /**
+   * Item name
+   * Used in data attributes needed for cypress tests
+   */
+  name?: string;
```
neSpecc: move this system attribute to the end of the interface. Developers shouldn't know about it.

types/tools/tool-settings.d.ts
```diff
@@ -1,5 +1,6 @@
 import { ToolConfig } from './tool-config';
 import { ToolConstructable, BlockToolData } from './index';
+import { PopoverItem } from '../../src/components/utils/popover';
```
neSpecc: looks like this path is not correct

neSpecc (changes requested):

src/styles/variables.css
TatianaFomina: border and background are applied due to focused state. Try to navigate with arrow keys and you'll see there's no border. Probably, special styles for focused active element are needed

src/components/block-tunes/block-tune-delete.ts
TatianaFomina: nope, removed

src/components/utils/popover.ts
TatianaFomina: fixed

types/tools/tool-settings.d.ts
TatianaFomina: fixed

types/configs/popover.d.ts
TatianaFomina: fixed

src/styles/variables.css
```diff
@@ -69,6 +69,10 @@
    * Confirm deletion bg
    */
   --color-confirm: #E24A4A;
+  --color-confirm-hover: #CE4343;
+
+  --color-line-active: #D8E8FA;
+  --color-line-gray: #EFF0F1;
```
neSpecc, line 75: so please derpecate the `--color-gray-border`;

src/styles/variables.css
```diff
@@ -69,6 +69,10 @@
    * Confirm deletion bg
    */
   --color-confirm: #E24A4A;
+  --color-confirm-hover: #CE4343;
+
+  --color-line-active: #D8E8FA;
```
neSpecc: is it still used?

types/configs/popover.d.ts
```diff
@@ -0,0 +1,75 @@
...
+   * Additional displayed text
+   */
+  secondaryLabel?: string;
+
+  /**
+   * True if item should be highlighted as active
+   */
+  isActive?: boolean;
+
+
```
neSpecc:
```suggestion
```

neSpecc:
Prevent the whole document selection by double clicks on the popover item

neSpecc:
1. Create empty Block by Enter
2. Press Tab to Open the Toolbox
3. Select Header by flipping using Tab, then press Enter
4. Press Tab to open Block Tunes
5. Flipper won't work

neSpecc:
Popover width should be fixed. Long labels should be wrapped with ellipsis

neSpecc:
H4 could not be activated
image

neSpecc:
1. Remove hover state right after clicking
2. Do not set the focused state to the confirmation item
image

neSpecc:
Prevent icons size & bordr jumping. Use 'outside' border.

image

TatianaFomina:
> H4 could not be activated

You are probably testing current feature with branch `feat/vertical-toolbox-multiple-items` of header-plugin. H4 icon in that branch has hardcoded black color, that's why active styles are not applied

src/styles/variables.css
TatianaFomina: removed

src/components/modules/toolbar/blockSettings.ts
neSpecc: so?

src/components/utils/popover.ts
```diff
@@ -381,37 +443,229 @@ export default class Popover extends EventsDispatcher<PopoverEvent> {
...
+    }
+    const originalStateItemEl = this.createItem(originalItem);
+
+    itemEl.parentElement.replaceChild(originalStateItemEl, itemEl);
+    this.items[index] = originalItem;
+
+    delete this.itemsRequiringConfirmation[index];
+
+    itemEl.removeEventListener('mouseleave', this.removeSpecialHoverBehavior);
+    this.flipper.removeOnFlip(this.onFlip);
```
neSpecc: it should be moved to `disableSpecialHoverAndFocusBehavior` to make it opposite the `enableSpecialHoverAndFocusBehavior`

src/components/utils/popover.ts
```diff
@@ -381,37 +443,229 @@ export default class Popover extends EventsDispatcher<PopoverEvent> {
...
+    );
   }
+
+  /**
+   * Enables special focus and hover behavior for item in confirmation state.
+   * This is needed to prevent item from being highlighted as hovered/focused just after click.
+   *
+   * @param item - html element of the item to enable special behavior for
+   */
+  private enableSpecialHoverAndFocusBehavior(item: HTMLElement): void {
```
neSpecc, line 575: could we move these method to some utility?

src/styles/variables.css
neSpecc: no answer

neSpecc:
1. Set Caret inside the Heading block using keyboard only
2. Press Tab to open Block Tunes
3. Select some Tune (for example, Delete)
4. Press Enter

Nothing happened, default Enter behaviour occurs

neSpecc:
1. Create the new Header
2. Click on the Plus Button near it.
3. The Heading 1 will be highlighted as "active" (it has no sense for the Toolbox?)
image

src/components/utils/popover.ts
TatianaFomina: does not seem possible

src/components/utils/popover.ts
TatianaFomina: added `disableSpecialHoverAndFocusBehavior`

src/styles/variables.css
TatianaFomina: added deprecated comment

neSpecc:
The code seems ok. Let's start testing.

neSpecc:
1. Create the new Block via keyboard
2. Choose the Checklist tool
3. Write several symbols in the Checklist
4. Press Tab to open Block Tunes
5. Select the Delete tune via keyboard. Press Enter to times.
6. The Block won't be deleted.   

https://user-images.githubusercontent.com/3684889/193423728-24efa058-9180-4765-888d-af8a6d5232a7.mp4

This problem does not reproduce at the `next` branch

src/components/block-tunes/block-tune-move-up.ts
```diff
@@ -90,8 +76,8 @@ export default class MoveUpTune implements BlockTune {
       return;
     }
 
-    const currentBlockElement = currentBlock.holder;
-    const previousBlockElement = previousBlock.holder;
+    const currentBlockElement = (currentBlock as BlockAPI).holder;
+    const previousBlockElement = (previousBlock as BlockAPI).holder;
```
gohabereg: Looks like you could just do this:
```suggestion
    const currentBlockElement = currentBlock!.holder;
    const previousBlockElement = previousBlock!.holder;
```

src/components/modules/toolbar/blockSettings.ts
```diff
@@ -42,82 +36,50 @@ export default class BlockSettings extends Module<BlockSettingsNodes> {
...
-  /**
-   * Instance of class that responses for leafing buttons by arrows/tab
+   * Getter for inner popover's flipper instance
    *
-   * @type {Flipper|null}
+   * @todo remove once BlockSettings becomes standalone non-module class
    */
-  private flipper: Flipper = null;
+  public get flipper(): Flipper {
+    return this.popover?.flipper;
```
gohabereg: It is either `?` is obsolete here or something is wrong with types

src/components/modules/toolbar/blockSettings.ts
```diff
@@ -144,24 +106,43 @@ export default class BlockSettings extends Module<BlockSettingsNodes> {
     /**
      * Fill Tool's settings
      */
-    this.addToolSettings(targetBlock);
+    const [tunesItemsConfig, additionalTunesContainer] = targetBlock.getTunes();
```
gohabereg: A bit confusing that those variables called differently inside `getTunes` method and here

src/components/modules/toolbar/index.ts
```diff
@@ -388,7 +388,7 @@ export default class Toolbar extends Module<ToolbarNodes> {
      * Appending Toolbar components to itself
      */
     $.append(this.nodes.actions, this.makeToolbox());
-    $.append(this.nodes.actions, this.Editor.BlockSettings.nodes.wrapper);
+    $.append(this.nodes.actions, this.Editor.BlockSettings.getElement());
```
gohabereg, line 391: Why not `render` instead of `getElement`?

src/components/utils/popover.ts
```diff
@@ -163,19 +170,31 @@ export default class Popover extends EventsDispatcher<PopoverEvent> {
    * @param options.className - additional class name to be added to the popover wrapper
    * @param options.filterLabel - label for the search Field
    * @param options.nothingFoundLabel - label of the 'nothing found' message
+   * @param options.customContent - arbitrary html element to be inserted before items list
+   * @param options.customContentFlippableItems - list of html elements inside custom content area that should be available for keyboard navigation
+   * @param options.scopeElement - editor container element
+   * @param shouldHighlightActivated - If true, activated item should be highlighted as active once clicked.
    */
-  constructor({ items, className, searchable, filterLabel, nothingFoundLabel }: {
+  constructor({ items, className, searchable, filterLabel, nothingFoundLabel, customContent, customContentFlippableItems, scopeElement, shouldHighlightActivated = false }: {
```
gohabereg: Looks like the builder pattern might be useful here

src/components/utils/popover.ts
```diff
@@ -381,30 +451,198 @@ export default class Popover extends EventsDispatcher<PopoverEvent> {
...
+      ...item,
+      ...item.confirmation,
+      confirmation: item.confirmation.confirmation,
+    } as PopoverItem;
+
+    this.items[itemIndex] = newItemData;
+
+    const confirmationStateItemEl = this.createItem(newItemData as PopoverItem);
+
+    confirmationStateItemEl.classList.add(Popover.CSS.itemConfirmation, ...Array.from(itemEl.classList));
```
gohabereg, line 528: Wouldn't it work?
```suggestion
    confirmationStateItemEl.classList.add(Popover.CSS.itemConfirmation, ...itemEl.classList);
```

src/components/utils/popover.ts
```diff
@@ -5,39 +5,7 @@ import SearchInput from './search-input';
 import EventsDispatcher from './events';
```
gohabereg, line 5: This file is way too much large, maybe we can split it somehow to follow SOLID?

src/components/modules/toolbar/index.ts
TatianaFomina: `getElement()` is not rendering anything, it only returns the wrapper

src/components/utils/popover.ts
TatianaFomina: there's a plan to refactor popover in further pull requests. probably it will even be moved to separate repository. I wouldn't touch it for now

src/components/utils/popover.ts
TatianaFomina: nope 
image

src/components/modules/toolbar/blockSettings.ts
TatianaFomina: fixed

src/components/utils/popover.ts
TatianaFomina: Is it big number of constructor params that bothers you? They are wrapped in object, so, client will only have to specify those props that are actually needed. What's the advantage of using builder pattern here?

neSpecc (approved):
image

neSpecc:
Write change logs, please

gohabereg (approved):

src/components/utils/popover.ts
gohabereg: Builder pattern would be helpful in case of common use cases of this class. Moreover, Director pattern might be useful as well. Anyway I think refactoring is required so maybe final version would be ok without Builder/Director

types/configs/index.d.ts
```diff
@@ -1,7 +1,10 @@
+import { fromCallback } from 'cypress/types/bluebird';
```
DenysVuika, line 1: this is actually a problem as it makes the cypress dependency, and causing the following errors for the apps
```
Error: node_modules/@editorjs/editorjs/types/configs/index.d.ts:1:30 - error TS2307: Cannot find module 'cypress/types/bluebird' or its corresponding type declarations.

1 import { fromCallback } from 'cypress/types/bluebird';
```

PR #2093, opened 2022-07-07
Improved block.insert() by allowing passing parameters as an object

aditink:


aditink:
Apologies, this was created accidentally.

PR #2095, opened 2022-07-12
Improved blocks.insert() by allowing passing parameters as an object …

### LINKED ISSUE #1817: API: improve block.insert() interface to allow passing parameters as an object instead of set
1. Describe a problem.

The current interface of the `api.blocks.insert` accepts parameters as a list:

```ts
this.api.blocks.insert(
      toolName,
      undefined,
      undefined,
      index,
      undefined,
      currentBlock.isEmpty
);    
```

It would better to pass parameters as an object:

```ts
this.api.blocks.insert({
  type: toolName,
  index,
  replace: currentBlock.isEmpty
});    
```

**Important**. We need to save the backward capability.

zang77:
* Improved block.insert() by allowing passing parameters as an object

* Fixed some issues commented out from last pr. Set the default parameter for insert.

* Added changelog for insert function.

Resolves #1817

### DIFF
--- a/src/components/modules/api/blocks.ts
+++ b/src/components/modules/api/blocks.ts
@@ -1,4 +1,4 @@
-import { BlockAPI as BlockAPIInterface, Blocks } from '../../../../types/api';
+import { BlockAPI as BlockAPIInterface, Blocks, NewBlock } from '../../../../types/api';
 import { BlockToolData, OutputData, ToolConfig } from '../../../../types';
 import * as _ from './../../utils';
 import BlockAPI from '../../block/api';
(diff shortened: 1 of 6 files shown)


aditink:
Fixes issue #1817

src/components/modules/api/blocks.ts
```diff
@@ -221,29 +221,23 @@ export default class BlocksAPI extends Module {
...
    *
-   * @param {string} type — Tool name
-   * @param {BlockToolData} data — Tool data to insert
-   * @param {ToolConfig} config — Tool config
-   * @param {number?} index — index where to insert new Block
-   * @param {boolean?} needToFocus - flag to focus inserted Block
-   * @param replace - pass true to replace the Block existed under passed index
+   * @param {NewBlock} newBlock — the block that will be inserted
    */
   public insert = (
```
neSpecc, line 228: it would be better to save the backward capability

types/api/block.d.ts
```diff
@@ -74,3 +74,39 @@ export interface BlockAPI {
    */
   dispatchChange(): void;
 }
+
+/**
+ * @interface NewBlock Describes new block configuration
+ */
+
+export interface NewBlock {
```
neSpecc, line 82:
```suggestion
export interface BlockInsertionParams {
```

neSpecc (changes requested):

neSpecc:
The solution should save the backward compatibility, but this PR is dropping it.

PR #2100, opened 2022-07-16
fix(tools-api): pasteConfig.tags now supports a sanitize config

robonetphy:
depends on next version of [image](https://github.com/editor-js/image/pull/189) tool and [simple-image](https://github.com/editor-js/simple-image/pull/55) tool


neSpecc:
I think, the best soultuion is to support our sanitization config at the `pasteConfig.tags` configuration:

```ts
  static get pasteConfig() {
    return {
      tags: [ 'img[src]' ],  // <-- all attributes except 'src' will be stripped
    };
  }
```

jorgectf:
Thank you for addressing this issue! FWIW, I'd suggest using tools focused on XSS sanitization such as [dompurify](https://github.com/cure53/DOMPurify).

robonetphy:
> I think, the best soultuion is to support our sanitization config at the `pasteConfig.tags` configuration:
> 
> ```ts
>   static get pasteConfig() {
>     return {
>       tags: [ 'img[src]' ],  // <-- all attributes except 'src' will be stripped
>     };
>   }
> ```
@neSpecc I tried this solution but in that case, the configuration for sanitisation is as below:
```
{
    "p": true,
    "h1": true,
    "h2": true,
    "h3": true,
    "h4": true,
    "h5": true,
    "h6": true,
    "img[src]": true,
    "pre": true,
    "b": {},
    "i": {},
    "a": {
        "href": true,
        "target": "_blank",
        "rel": "nofollow"
    },
    "mark": {
        "class": "cdx-marker"
    },
    "code": {
        "class": "inline-code"
    },
    "br": {}
}```
and in that case, if I paste this correct configuration.
`<img src='https://cdn.pixabay.com/photo/2015/04/23/22/00/tree-736885__480.jpg'/>`
it's not working. the sanitized data comes empty and it pastes as text.

**Comment by robonetphy** [comment]
> Thank you for addressing this issue! FWIW, I'd suggest using tools focused on XSS sanitization such as [dompurify](https://github.com/cure53/DOMPurify).

@jorgectf We are planning to integrate the library you recommended in our next release because accordingly, we need to update every tool with the API. So, this is just a hotfix right now.

**Comment by neSpecc** [comment]
> > I think, the best soultuion is to support our sanitization config at the `pasteConfig.tags` configuration:
> > ```ts
> >   static get pasteConfig() {
> >     return {
> >       tags: [ 'img[src]' ],  // <-- all attributes except 'src' will be stripped
> >     };
> >   }
> > ```
> 
> @neSpecc I tried this solution but in that case, the configuration for sanitisation is as below:
> 
> ```
> {
>     "p": true,
>     "h1": true,
>     "h2": true,
>     "h3": true,
>     "h4": true,
>     "h5": true,
>     "h6": true,
>     "img[src]": true,
>     "pre": true,
>     "b": {},
>     "i": {},
>     "a": {
>         "href": true,
>         "target": "_blank",
>         "rel": "nofollow"
>     },
>     "mark": {
>         "class": "cdx-marker"
>     },
>     "code": {
>         "class": "inline-code"
>     },
>     "br": {}
> }```
> and in that case, if I paste this correct configuration.
> ``
> it's not working. the sanitized data comes empty and it pastes as text.
> ```

You should not create a config manually, it should be collected from tools. For example, Image tool should change `img` with `img[src]` [here](https://github.com/editor-js/image/blob/master/src/index.js#L237)

**Comment by robonetphy** [comment]
> > > I think, the best soultuion is to support our sanitization config at the `pasteConfig.tags` configuration:
> > > ```ts
> > >   static get pasteConfig() {
> > >     return {
> > >       tags: [ 'img[src]' ],  // <-- all attributes except 'src' will be stripped
> > >     };
> > >   }
> > > ```
> > 
> > 
> > @neSpecc I tried this solution but in that case, the configuration for sanitisation is as below:
> > ```
> > {
> >     "p": true,
> >     "h1": true,
> >     "h2": true,
> >     "h3": true,
> >     "h4": true,
> >     "h5": true,
> >     "h6": true,
> >     "img[src]": true,
> >     "pre": true,
> >     "b": {},
> >     "i": {},
> >     "a": {
> >         "href": true,
> >         "target": "_blank",
> >         "rel": "nofollow"
> >     },
> >     "mark": {
> >         "class": "cdx-marker"
> >     },
> >     "code": {
> >         "class": "inline-code"
> >     },
> >     "br": {}
> > }```
> > and in that case, if I paste this correct configuration.
> > ``
> > it's not working. the sanitized data comes empty and it pastes as text.
> > ```
> 
> You should not create a config manually, it should be collected from tools. For example, Image tool should change `img` with `img[src]` [here](https://github.com/editor-js/image/blob/master/src/index.js#L237)

Yup, I have done the same things I am just doing `console.log` inside the `Paste` module as debugging. So I know it will take updated value from `simple-image` tool

File: src/components/modules/paste.ts [hunk 1]
```diff
@@ -324,13 +335,31 @@ export default class Paste extends Module {
 
         return;
       }
-
-      this.toolsTags[tag.toUpperCase()] = {
-        tool,
-      };
+      /** Sanitization configuration is string */
+      if (_.isString(tag)) {
```
gohabereg: Maybe we should rename the variable? Something like `tagOrSanitizeConfig`, because it's confusing rn — comment says about config but variable name is `tag`

src/components/modules/paste.ts
```diff
@@ -324,13 +335,31 @@ export default class Paste extends Module {
...
+        this.toolsTags[tag.toUpperCase()] = {
+          tool,
+        };
+      }
+      /** Sanitization configuration is object */
+      if (_.isObject(tag)) {
+        this.toolsTags[Object.keys(tag)[0].toUpperCase()] = {
```
gohabereg: You can extract `Object.keys(tag)[0]` to a variable

src/components/modules/paste.ts
```diff
@@ -546,7 +575,14 @@ export default class Paste extends Module {
...
 
         const toolTags = tags.reduce((result, tag) => {
-          result[tag.toLowerCase()] = {};
+          /** Sanitization configuration is string */
+          if (_.isString(tag)) {
+            result[tag.toLowerCase()] = {};
+          }
+          /** Sanitization configuration is object */
+          if (_.isObject(tag)) {
+            result[Object.keys(tag)[0].toUpperCase()] = tag[Object.keys(tag)[0]];
```
gohabereg: Repeated code, could be extracted to a separate method

gohabereg (changes requested):
Would be great to cover it with unit tests

src/components/modules/paste.ts
```diff
@@ -324,13 +335,31 @@ export default class Paste extends Module {
...
-        tool,
-      };
+      /** Sanitization configuration is string */
+      if (_.isString(tag)) {
+        this.toolsTags[tag.toUpperCase()] = {
+          tool,
+        };
+      }
+      /** Sanitization configuration is object */
+      if (_.isObject(tag)) {
```
neSpecc: pls use `else if` there

src/components/modules/paste.ts
```diff
@@ -324,13 +335,31 @@ export default class Paste extends Module {
...
     });
 
-    this.tagsByTool[tool.name] = tags.map((t) => t.toUpperCase());
+    this.tagsByTool[tool.name] = tags.map((t) => {
+      /** Sanitization configuration is string */
+      if (_.isString(t)) {
+        return t.toUpperCase();
+      }
+      /** Sanitization configuration is object */
+      if (_.isObject(t)) {
```
neSpecc: `else if`

src/components/modules/paste.ts
neSpecc: What if developer will specify several tags?

```js
tags: [
  'tbody',
  {
    table: {
      width: true,
      height: true,
    },
  },
  {
    td: {
      colspan: true,
      rowspan: true
    },
    tr: {  // <-- the second tag
      height: true
    }
  },
]
```

will we allow that? If not, we need to write about it somewhere.

neSpecc (changes requested):

src/components/modules/paste.ts
gohabereg: Good point btw, we can implement multiple tags in the future

robonetphy: @gohabereg, @neSpecc  I can do the same right now but I need one tag which I can use parent to point out corresponding `tool` in `this.toolsTags` (which is mapping of tool tag name to tools). I can consider the first element for the same. what are your suggestions?

robonetphy: this pr also include this too

gohabereg (approved):
LGTM, still lack of tests but ok to add them later

ilyamore88 (approved):

src/components/modules/paste.ts
```diff
@@ -306,31 +313,66 @@ export default class Paste extends Module {
     }
   }
 
+  /**
+   * Get tags from tags configuration.
+   *
+   * @param {string | object} tagsConfig - tag name or sanitize config object.
+   * @returns {string[]} array of tags.
+   */
+  private getTags = (tagsConfig: (string | object)): string[] => {
```
neSpecc: seems like this method could return `undefined` as well in case if `tagConfig` will be passed as something different than `string` or `object`. 

Could you handle this please? Because for now the Exception will occur in the `[...tags]` statement

src/components/modules/paste.ts
```diff
@@ -543,10 +585,16 @@ export default class Paste extends Module {
             break;
         }
 
-        const { tags } = tool.pasteConfig;
+        const { tags: tagsOrSanitizeConfigs } = tool.pasteConfig;
+
+        const toolTags = tagsOrSanitizeConfigs.reduce((result, tagOrSanitizeConfig) => {
```
neSpecc, line 640: please add some comments here for the future maintenance.

types/configs/paste-config.d.ts
```diff
@@ -6,7 +6,7 @@ export interface PasteConfig {
    * Array of tags Tool can substitute
```
neSpecc:
```suggestion
   * Array of tags Tool can substitute. 
   * 
   * Could also contain a sanitise-config if you need to save some tag's attribute. 
   * For example: 
   * [
   *   { 
   *     img: { src: true },
   *   }
   * ],
```

neSpecc (changes requested):
The solution does not work at the moment. See my comment.

src/components/modules/paste.ts
```diff
@@ -543,10 +585,16 @@ export default class Paste extends Module {
...
+          const tags = this.getTags(tagOrSanitizeConfig);
+
+          tags.forEach((tag) => {
+            const sanitizationConfig = _.isObject(tagOrSanitizeConfig) ? tagOrSanitizeConfig[tag] : null;
```
neSpecc, line 644: This statement will always return `undefined` or `null` because `tag` is uppercased (`IMG`), but keys of `tagOrSanitizeConfig` (`img`) are not

image

src/components/modules/paste.ts
```diff
@@ -306,31 +313,66 @@ export default class Paste extends Module {
...
+   *
+   * @param {string | object} tagsConfig - tag name or sanitize config object.
+   * @returns {string[]} array of tags.
+   */
+  private getTags = (tagsConfig: (string | object)): string[] => {
+    /**
+     * If tagsConfig is string, return it as an array.
+     */
+    if (_.isString(tagsConfig)) {
+      return [ tagsConfig.toUpperCase() ];
```
neSpecc: I think, the uppercasing is out of the scope of this method and should be removed from it to somewhere else

src/components/modules/paste.ts
neSpecc: Still actual

src/components/modules/paste.ts
```diff
@@ -306,31 +313,71 @@ export default class Paste extends Module {
     }
   }
 
+  /**
+   * Get tags from tags configuration.
```
neSpecc: The description is not clear.

robonetphy: changed

src/components/modules/paste.ts
robonetphy: return empty array

src/components/modules/paste.ts
```diff
@@ -306,31 +313,69 @@ export default class Paste extends Module {
...
+   */
+  private getTags = (tagOrSanitizeConfig: (string | object)): string[] => {
+    /**
+     * If string, then it is a tag name.
+     */
+    if (_.isString(tagOrSanitizeConfig)) {
+      return [ tagOrSanitizeConfig ];
+    }
+    /**
+     * If object, then it's keys are tags.
```
jorgectf:
```suggestion
     * If object, then its keys are tags.
```

robonetphy: Thanks

neSpecc:
add a changelog describing fix and API change please

neSpecc:
The solution still does not work as expected. If a Tool does not specify the sanitizer config, all attributes should be removed. For now, it doesn't work.

I'm digging into it.

neSpecc:
Seems ok for now. Will test it again tomorrow

robonetphy:
hey @neSpecc,  Pls check and let me know when to merge.

neSpecc:
Sorry, the last test became a problem and I've spent some time on it. I've found the bug with our sanitizing dependency HTMLJanitor. It has a bug with Table sanitizing (https://github.com/guardian/html-janitor/issues/3) so I've added a few lines to handle that case. Now all the tests work fine.

neSpecc (approved):

PR #2101, opened 2022-07-16
feat(collaborative): function for getting changes from string

ilyamore88:
I've created a function that will help us to get changes from a string. This algorithm compares two strings from the left side first, then from the right. The resulting values allow you to identify the operation that was performed on the string. It can be Insert or Remove operations.
Also, I've written some tests for all of these operations.


src/components/collaborative-editing/diff/getChangesFromString.ts
```diff
@@ -0,0 +1,153 @@
+/**
+ * Possible operation types
+ */
+export enum Operation {
```
neSpecc:
```suggestion
export enum OperationType {
```

src/components/collaborative-editing/diff/getChangesFromString.ts
```diff
@@ -0,0 +1,153 @@
...
+   */
+  Remove,
+
+  /**
+   * User replaces symbols
+   */
+  Replace
+}
+
+type InsertOperation = {
```
neSpecc: why not an interface?

src/components/collaborative-editing/diff/getChangesFromString.ts
```diff
@@ -0,0 +1,153 @@
...
+  type: Operation.Insert;
+
+  /**
+   * Index where user inserted symbols
+   */
+  from: number;
+
+  /**
+   * Which symbols user inserted
```
neSpecc: Are several symbols allowed there? Or just the single one?

src/components/collaborative-editing/diff/getChangesFromString.ts
```diff
@@ -0,0 +1,153 @@
...
+   */
+  data: string;
+
+  /**
+   * Length of inserted data
+   */
+  length: number;
+}
+
+type RemoveOperation = {
```
neSpecc: I think, they should be moved to separate files

src/components/collaborative-editing/diff/getChangesFromString.ts
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * Function finds difference between two different strings
+ *
+ * @param str1 - original string
+ * @param str2 - new string that comes from the first
+ *
+ * @returns {object} object of operation
+ */
+export function getChangesFromString(str1: string, str2: string): InsertOperation | RemoveOperation | ReplaceOperation {
```
neSpecc:
```suggestion
export function createOperationByStringsDiff(str1: string, str2: string): InsertOperation | RemoveOperation | ReplaceOperation {
```

src/components/collaborative-editing/diff/getChangesFromString.ts
```diff
@@ -0,0 +1,153 @@
...
+  const length1 = str1.length,
+      length2 = str2.length;
+
+  let left = 0;
+
+  while (str1[left] === str2[left]) {
```
neSpecc: Please provide comments for the main logical blocks.

src/components/collaborative-editing/diff/getChangesFromString.ts
```diff
@@ -0,0 +1,153 @@
...
+    left++;
+  }
+
+  let right = 0;
+
+  while (str1[length1 - right - 1] === str2[length2 - right - 1]) {
+    right++;
+  }
+
+  if (left + right < length1 && left + right < length2) {
```
neSpecc: Looks like we need to move such statementes to the separate methods somehow. To prevent growing of the main method.

test/cypress/tests/collaborative-editing/diff/getChangesFromString.spec.ts
```diff
@@ -0,0 +1,216 @@
+import {
+  getChangesFromString,
+  Operation
+} from '../../../../../src/components/collaborative-editing/diff/getChangesFromString';
+
+describe('getChangesFromString function', () => {
+  describe('Insert operation', () => {
+    it('should compute operation when a user inserts symbols at the start of the string', () => {
```
neSpecc:
```suggestion
    it('should create an Insert Operation when a user inserts symbols at the start of the string', () => {
```

src/components/collaborative-editing/diff/getChangesFromString.ts
```diff
@@ -0,0 +1,153 @@
...
+   */
+  from: number;
+
+  /**
+   * Which symbols user inserted
+   */
+  data: string;
+
+  /**
+   * Length of inserted data
```
neSpecc: And why do we need this length?

test/cypress/tests/collaborative-editing/diff/getChangesFromString.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+        data: 'Changed ',
+        length: 8,
+      };
+
+      const result = getChangesFromString(before, after);
+
+      expect(result).deep.equal(expected);
+    });
+
+    it('should compute operation when a user inserts symbols in the string', () => {
```
neSpecc:
```suggestion
    it('should create an Insert Operation when a user inserts symbols inside the string', () => {
```

neSpecc (changes requested):

src/components/collaborative-editing/diff/getChangesFromString.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  /**
+   * User removes symbols
+   */
+  Remove,
+
+  /**
+   * User replaces symbols
+   */
+  Replace
```
gohabereg: It's better to have two operations (Insert & Remove) instead of replace.

Number of transformations is n^2 where n is number of operations. So it's better to keep number of operations as little as possible

src/components/collaborative-editing/diff/getChangesFromString.ts
```diff
@@ -0,0 +1,153 @@
...
+ * Function finds difference between two different strings
+ *
+ * @param str1 - original string
+ * @param str2 - new string that comes from the first
+ *
+ * @returns {object} object of operation
+ */
+export function getChangesFromString(str1: string, str2: string): InsertOperation | RemoveOperation | ReplaceOperation {
+  const length1 = str1.length,
+      length2 = str2.length;
```
gohabereg:
```suggestion
  const length1 = str1.length;
  const length2 = str2.length;
```

src/components/collaborative-editing/diff/getChangesFromString.ts
gohabereg: In case of equal strings this cycle would go infinite

src/components/collaborative-editing/diff/getChangesFromString.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  let left = 0;
+
+  while (str1[left] === str2[left]) {
+    left++;
+  }
+
+  let right = 0;
+
+  while (str1[length1 - right - 1] === str2[length2 - right - 1]) {
```
gohabereg: Same here, if strings are equal, cycle would go infinite

gohabereg (changes requested):

src/components/collaborative-editing/diff/getChangesFromString.ts
ilyamore88: It might be several symbols. When a user is typing too fast, or he's replacing some symbols with others.

src/components/collaborative-editing/diff/getChangesFromString.ts
ilyamore88: I've replaced it with variables.

src/components/collaborative-editing/diff/createOperationByStringsDiff.ts
```diff
@@ -0,0 +1,100 @@
+import { InsertOperation, OperationType, RemoveOperation } from './types';
+
+/**
+ * Function finds difference between two different strings
+ *
+ * @param str1 - original string
+ * @param str2 - new string that comes from the first
+ *
+ * @returns {object[]} array of operations
```
neSpecc: if it could return several operations, so name the function `createOperationsByStringsDiff`. And the file as well.

test/cypress/tests/collaborative-editing/diff/createOperationByStringsDiff.spec.ts
```diff
@@ -0,0 +1,222 @@
...
+          index: 7,
+          data: 'is ',
+        },
+      ];
+
+      const result = createOperationByStringsDiff(before, after);
+
+      expect(result).deep.equal(expected);
+    });
+
```
neSpecc: what about a case with several differences?

```js
const before = 'String is changed'
const after = 'Number is changed many times'
```

What will return the `createOperationByStringsDiff` method? Do we need to add test cases for that?

neSpecc (changes requested):

test/cypress/tests/collaborative-editing/diff/createOperationByStringsDiff.spec.ts
ilyamore88: We decided, that a user can't do several changes. In this case, the algorithm will return remove and insert operations (like the user replaced the first string with the second one.

neSpecc: ok, but add a test case for that please

ilyamore88: Done!

src/components/collaborative-editing/diff/createOperationByStringsDiff.ts
ilyamore88: Done

neSpecc (approved):

gohabereg (approved):

Issue #1406, opened 2020-10-29
[Bug] Dragging an image block to another location copies it & re-uploads


Duncank:
When you drag an image block to another spot in the editor, the block stays at the old location and a new block is created in the new location. After creation, the 'uploader'-function is triggered, re-uploading the image to the server.

Steps to reproduce:
1. Go to https://editorjs.io/
2. Click and hold on the image block with the Codex logo.
3. Drag it to another location and let go.
4. The block is duplicated and the image is uploaded again.

Expected behavior:
That the block moved, instead of duplicated.

Screenshots:
![image](https://user-images.githubusercontent.com/719365/97578803-97d40780-19f1-11eb-807c-f6844042a88f.png)

Device, Browser, OS:
Desktop, Chrome 87, Windows 10

Editor.js version:
2.19.0

Plugins you use with their versions:
image: 2.6.0

hyun-park:
@neSpecc 
Hi! Any update on this issue?

zang77:
Hello all, I will be working on this issue with a mentor.

neSpecc:
I think we should disable the dragging of such blocks by their content. 

The dragging of Blocks by pointer (Block Tune Toggler) will be implemented separately.

PR #2107, opened 2022-07-25
Fixed issue #1406

zang77:
#1406

### DIFF
--- a/src/components/modules/dragNDrop.ts
+++ b/src/components/modules/dragNDrop.ts
@@ -118,5 +118,10 @@ export default class DragNDrop extends Module {
    */
   private processDragOver(dragEvent: DragEvent): void {
     dragEvent.preventDefault();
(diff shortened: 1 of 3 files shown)


src/components/modules/dragNDrop.ts
```diff
@@ -118,5 +118,10 @@ export default class DragNDrop extends Module {
    */
   private processDragOver(dragEvent: DragEvent): void {
     dragEvent.preventDefault();
+    const selectedBlock = this.Editor.BlockManager.currentBlock;
+    if (selectedBlock.holder.querySelector('img')) {
```
neSpecc, line 122: It is the hardcoding of the Image Tool internal thing into editor core. I think we need to fix the reason of described behavior instead of removing a block.

neSpecc:
Thanks for trying to resolve the issue. The provided solution is not suitable and requires more designing.

Issue #2070, opened 2022-05-26
Feature Request: API for opening/closing the Toolbox


nberongoy:
this.api.toolbar.open();

is not working toolbar is not opening

neSpecc:
Couldn't reproduce. The Plus Button and the Block Settings toggler (not the Toolbox) appears after calling the `toolbar.open()`. Could you provide steps to reproduce?

nberongoy:
this is my code toggleBlockSettings is working I wanted the plus button to trigger so the dropdown of blocks will appear

```

render() {
    const rootNode = document.createElement('div');
    this.nodes.holder = rootNode;

    const root = createRoot(rootNode);

    root.render(
      <NeyarTextComponent
        data={this.data.neyarText}
        onKeyUp={() => {
          this.api.toolbar.open();
        }}
      />
    );

    return this.nodes.holder || '';
  }
  
  
```

nberongoy:
is there way to open the Toolbox ?

neSpecc:
We could expose an API method for toggling the Toolbox

nberongoy:
Nice yes please and make to focus shift to the filter. I would deeply appreciate that!

PR #2114, opened 2022-08-02
Added API method for opening/closing toolbox

zang77:
Related issue: #2070 

Added an api method for opening toolbox. Close can be done by using this.api.toolbar.close().

### DIFF
--- a/src/components/modules/api/toolbar.ts
+++ b/src/components/modules/api/toolbar.ts
@@ -15,6 +15,7 @@ export default class ToolbarAPI extends Module {
     return {
       close: (): void => this.close(),
       open: (): void => this.open(),
+      openToolbox: (): void => this.openToolbox(),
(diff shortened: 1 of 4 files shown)


src/components/modules/api/toolbar.ts
```diff
@@ -15,6 +15,7 @@ export default class ToolbarAPI extends Module {
     return {
       close: (): void => this.close(),
       open: (): void => this.open(),
+      openToolbox: (): void => this.openToolbox(),
```
neSpecc, line 18: maybe `toogleToolbox(openingState?: boolean)` will be more convenient, like `toggleBlockSettins`? It will allow closing the Toolbox without closing the Toolbar

neSpecc:
Also, it would be nice to have a tests for the new Toolbox API method

PR #2117, opened 2022-08-02
fix(onChange): fix of exception on input autofill event

### LINKED ISSUE #2116: JS error when Input event is not of type InputEvent
Describe a bug.

Steps to reproduce:
1. Please download the attached zip. file - [editojs_issue_dueto_chrome_bug.html.zip](https://github.com/codex-team/editor.js/files/9243287/editojs_issue_dueto_chrome_bug.html.zip). The zip contains a single html file.
2. Put the file in a location so that it can be server by a web server.
3. Open the html in Chrome (issue also exists in Safari). The `input` fields needs to be pre-filled with username and password. If for the domain from where the html is served has no saved credentials then save one and reload the page.
4. The credentials should not come pre-filled. Click anywhere on the page. This cause the `input` event to fire but this time the object type is not instance of `InputEvent` but `Event`. Because of that the code shown below: https://github.com/codex-team/editor.js/blob/6c0555a3229a16641709cdc83ec248af87f3c77a/src/components/block/index.ts#L206 It tries to evaluate `mutationsOrInputEvent.some` but `Event` has no method `some` resulting into this error.

Expected behavior:

It seems to be a browser defect but the same issue exists in Safari as well. Today only I have filed this defect (https://bugs.chromium.org/p/chromium/issues/detail?id=1349207) so not sure if this will get accepted or it will be dismissed. Either way I think Editor JS should take care of this edge case and not run into JS error.

The fix is simple. Just check if it is `instanceof Event`  with `type` property set to `input`.

Screenshots:
Screenshot 2022-08-02 at 8 20 56 PM

Device, Browser, OS:
Chrome 103.0.5060.134 (Official Build) (x86_64)
MacOS 12.5

Editor.js version:
2.25.0

None

applegrew:
This is to fix https://github.com/codex-team/editor.js/issues/2116

### DIFF
--- a/src/components/block/index.ts
+++ b/src/components/block/index.ts
@@ -203,7 +203,7 @@ export default class Block extends EventsDispatcher<BlockEvents> {
    * Is fired when DOM mutation has been happened
    */
   private didMutated = _.debounce((mutationsOrInputEvent: MutationRecord[] | InputEvent = []): void => {
-    const shouldFireUpdate = mutationsOrInputEvent instanceof InputEvent ||
+    const shouldFireUpdate = mutationsOrInputEvent instanceof InputEvent || (mutationsOrInputEvent instanceof Event && mutationsOrInputEvent.type === 'input') ||
       !mutationsOrInputEvent.some(({
         addedNodes = [],
         removedNodes,


neSpecc:
Please, provide a test case for added usecase

Issue #2090, opened 2022-06-27
[Bug] Cannot read properties of undefined (reading 'holder')


applegrew:
Describe a bug.

Steps to reproduce:
1. Go to https://codepen.io/applegrew/pen/RwMbqdo
2. Open browser console and click within the black rectangle randomly. You will see the error mentioned here.
3. Click on "Readonly" checkbox to uncheck it and you will see that clicking inside the box still results into this error.

Expected behavior:
 There should be no JS error.

Screenshots:
<img width="1792" alt="Screenshot 2022-06-27 at 9 37 15 PM" src="https://user-images.githubusercontent.com/434574/175986878-237526d8-60cf-4861-a16e-19745da7169f.png">

Device, Browser, OS: Chrome on MacOS

Editor.js version: 2.25.0

Plugins you use with their versions: None

7polo:
> 
same question
when 'blocks' property is empty

MEBonham:
I'm running into what I think is the same issue.

AlexRMU:
Use `blocks: [{ type: "paragraph", data: {} }]`

berkaygurbuz:
I've got same error. any updates ?

dimuska139:
Any news?

644792799:
Use 
```
...
data: {
  time: (new Date()).getTime(),
  version: '2.25.0',
  blocks: [{
	type: 'paragraph',
	data: {
	  text: '<a></a>'
	}
  }]
}
...
```

antoineol:
I also reproduce with `editor.blocks.renderFromHTML('<p></p>');` without the "readonly" switch, see this codepen:

https://codepen.io/antoine_ol/pen/BaGKaKE

Open the browser console and click anywhere in the page. You will see errors.

Add any content inside the `<p>` tag (e.g. `<p>a</p>`) and it will work. But it fails with empty paragraph content. It not only shows an error in the console, but also fails to render the editor, which is crashed.

johannbuscail:
i have the same issue, any update ?

DarioRega:
same issue here, it is when i set the the data dynamically with the editor.render({...}) and in fact when it is empty it throw that same error

justynjozwiak:
same trouble here! also `editor.render({...})`

devon2018:
I got this same issue and turns out it was because i had applied a tune to the paragraph component but the tune had not been registered (I commented it out while testing something else.)

Lubin7792:
[...]

PR #2123, opened 2022-08-07
[Fix] add default block for readonly toggler

kryptamine:
It fixes the problem when you toggle the read-only property with empty blocks attached to the editor.
https://github.com/codex-team/editor.js/issues/2090


neSpecc:
I don't think, that `readOnly.toggle()` should affect block list state. It would be better to find the reasong of error and add some statement there.

PR #2131, opened 2022-08-29
Fix (??) type of `pasteConfig`

RyanCavanaugh:
TypeScript 4.9 found something is wrong with this code, but it's unclear (to me) which line is wrong. This PR is a guess, do with it what you will.

In paste.ts there's a check to see if `pasteConfig === false`:
https://github.com/codex-team/editor.js/blob/next/src/components/modules/paste.ts#L287

However, this getter never returns false because if the LHS of the `||` is `false`, `{ }` is returned instead.

It seems like this meant to be `??` instead so that if `this.constructable[InternalBlockToolSettings.PasteConfig]` was `undefined` (missing), then `{}` would be returned instead. But maybe you meant `false` here - I don't know.


neSpecc:
Thanks, it really looks strange. We'll write a testcase for that and then include this fix to the next release

ilyamore88:
Hi, @RyanCavanaugh. Thank you for your help. I'm closing your request because your fix is merged in this request (https://github.com/codex-team/editor.js/pull/2322).

PR #2156, opened 2022-10-28
Adding support for conditional inline-tool 

cortantief:
### Use case
I need a way to make a custome element inside a block, but for it to work properly data validation was a must.
I didn't find a clean way to do it therefore I taught it might be a good idea an optional validation method.

I called the method "isValid" I don't know if it is the more idiomatic, maybe toShow ?
there is no breaking change since the method is optional, if inline tools implemented a "hacky" way to have the same feature it will still work.

I did my best to add some test

Im sharing this hoping It might help others

### DIFF
--- a/src/components/modules/toolbar/inline.ts
+++ b/src/components/modules/toolbar/inline.ts
@@ -196,6 +196,14 @@ export default class InlineToolbar extends Module<InlineToolbarNodes> {
 
     this.nodes.wrapper.style.left = Math.floor(newCoords.x) + 'px';
     this.nodes.wrapper.style.top = Math.floor(newCoords.y) + 'px';
+
+    const selection = SelectionUtils.get();
+
+    for (const button of Array.from(this.nodes.buttons.querySelectorAll('button'))) {
+      const instance = this.toolsInstances.get(button.dataset.tool);
+
+      button.hidden = instance.isValid && !instance.isValid(selection);
+    }
   }
 
   /**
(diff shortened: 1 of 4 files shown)


neSpecc:
Interesting idea. Should be designed more carefully.

The use case is still not clear to me. How user should know that some tool in the Inline Toolbar will appear only if it selects specific text fragments? It's not obvious from the user's perspective.

cortantief:
Good point, I guess showing it in a "disabled" state with a tooltip message telling explicitly what condition must be fulfilled ?
For my use case I needed to make an inline tool that only accept "IBAN" to format them in a certain way.
In fact I also needed some kind of inline-block

neSpecc:
> For my use case I needed to make an inline tool that only accept "IBAN" to format them in a certain way.

How the IBAN should be formatted?

cortantief:
It was a two step process first conditionnal inline to ensure the iban is valid and an inline block to display it.
I wanted given that iban to automatically  display for example the bank, the BIC etc bit disaying it onmy when the user click on that IBAN using a kind of dropdown button sorry if im not clear im at work.

cortantief:
It was a two step process first conditionnal inline to ensure the iban is valid and an inline block to display it.
I wanted given that iban to automatically  display for example the bank, the BIC etc bit disaying it onmy when the user click on that IBAN using a kind of dropdown button sorry if im not clear im at work.

PR #2158, opened 2022-11-01
Add support for copying and pasting of "strong" and "em" HTML elements

### LINKED ISSUE #1666: Editor JS erase can't read strong and em tags in renderFromHTML
I am trying to convert an HTML content into content for editorJS, however when it transforms it removes the <strong> and <em> tags, how can I fix this?

arsm100:
Fixes #1666 

Followed the [suggestion](https://github.com/codex-team/editor.js/issues/1666#issuecomment-1238587836) of @neSpecc to extend the sanitization config for both italic and bold formatting. 

I also added tests to cover the new use cases.

## Before:

https://user-images.githubusercontent.com/44353593/199292460-d22fd918-9ef2-45e2-ad97-599eafa3f9d9.mov

## After:

https://user-images.githubusercontent.com/44353593/199292392-6d9b7b99-f410-4c02-ae65-18565c5e458b.mov

### DIFF
--- a/src/components/inline-tools/inline-tool-bold.ts
+++ b/src/components/inline-tools/inline-tool-bold.ts
@@ -23,13 +23,14 @@ export default class BoldInlineTool implements InlineTool {
 
   /**
    * Sanitizer Rule
-   * Leave <b> tags
+   * Leave <b> and <strong> tags
    *
    * @returns {object}
    */
   public static get sanitize(): SanitizerConfig {
     return {
       b: {},
+      strong: {},
     } as SanitizerConfig;
   }
 
--- a/src/components/inline-tools/inline-tool-italic.ts
+++ b/src/components/inline-tools/inline-tool-italic.ts
@@ -23,13 +23,14 @@ export default class ItalicInlineTool implements InlineTool {
 
   /**
    * Sanitizer Rule
-   * Leave <i> tags
+   * Leave <i> and <em> tags
    *
    * @returns {object}
    */
   public static get sanitize(): SanitizerConfig {
     return {
       i: {},
+      em: {},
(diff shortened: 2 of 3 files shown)


arsm100:
Hi @neSpecc @gohabereg @TatianaFomina @ilyamore88! Do you mind reviewing this to get the linked issue resolved? Thanks!

neSpecc:
I think the Inline Tool should return only a single tag, not a set of one. It would be more straightforward. Maybe I'm wrong.

Anyway, the problem you're resolving is understandable. Maybe we could somehow support `strong` tags on paste, but convert them to `b`? Probably it will require a new API improvement, to make it similar to the `handleTags` like in Block Tools.

We need to think about it a little bit more.

varna:
Well, I think it has to accept both variations. The question is if we should replace one with another. `<strong>` and `<em>` are semantic elements that are given more importance on accessibility side (i.e. screen readers). But `<b>` and `<i>` elements are simply made for styling purposes and totally ignored.

My personal preference would be to convert any instance of `<b>` to `<strong>` and `<i>` to `<em>`. Also updating the Inline Tool to use these instead.

buranovite:
Please add `strong: {}` and `em: {}` to the respective sanitize rules as a quick fix for when pasting is broken. After that there will be enough time to worry about some details and difficulties that arise (for ex. user pasted `<strong>` tag, but can't add or remove it because the tool adds and deletes only `<b>`, not `<strong>`).

However, for now please add the sanitize rules, to save people efforts if they have to copy and paste large texts.

gorenburg:
can we, please, re-consider this PR?

un000:
bump very needed

PR #2164, opened 2022-11-10
Also check root node for mutationFree attribute

oskarer:
This allows to set `mutationFree` on any element added in the `render()` function, not just added or removed child nodes. For example this case which I encountered (not a full example, somewhat psuedo code);

```
render() {
  const container = document.createElement('div');
  // Also add some input field to set video URL here

  if (data.url) {
    const video = document.createElement('lite-youtube');
    video.dataset.mutationFree = 'true'; // Only works with this PR
    video.setAttribute('videoid', 'someID');
    container.appendChild(video);
  }

  return container;
}
```

This would trigger the `onChange` event since `lite-youtube` is a custom element that adds a lot of elements to the DOM after `render()` has run. But in reality nothing has really changed, the only thing that should trigger a change is change of the video URL in the `<input>` field.


src/components/block/index.ts
```diff
@@ -205,13 +205,18 @@ export default class Block extends EventsDispatcher<BlockEvents> {
...
+      removedNodes,
+      target,
+    }) => {
+      if ((target as HTMLElement).dataset?.mutationFree === 'true') {
+        return true;
+      }
+
+      return [...Array.from(addedNodes), ...Array.from(removedNodes)]
+        .some(node => $.isElement(node) && (node as HTMLElement).dataset.mutationFree === 'true');
+    });
```
neSpecc: This code should be separated into several parts with clear names. For now, it is hard to read and understand.

Also, we need to write a tests for the covered cases.

neSpecc (changes requested):

neSpecc:
It would be great to have a test case for that change.

I think we'll include it in the upcoming release.

oskarer:
> It would be great to have a test case for that change.
> 
> I think we'll include it in the upcoming release.

@neSpecc Thanks, can you give me any pointers where a great place would be to add a spec for it? At least I've rebased now so it's mergeable.

neSpecc:
> @neSpecc Thanks, can you give me any pointers where a great place would be to add a spec for it? At least I've rebased now so it's mergeable.

I guess `/test/cypress/tests/onChange.spec.ts` will be a nice place

Issue #2175, opened 2022-11-25
[Bug] Can not delete nested list first item


Rue-pro:
Can not delete nested list first item

Steps to reproduce:
1. Create nested list
2. Try to delete first item of the nested list

Expected behavior:
First element of the nested list is deleted

Demo:
![Alt Text](https://media.giphy.com/media/pxrg8JalGk6J0qn0jM/giphy.gif)

Device, Browser, OS:
PC, Chrome, Linux

Editor.js version:
2.26

Plugins you use with their versions:
@editorjs/nested-list 1.0.2

Rue-pro:
Would like to fix this bug

neSpecc:
It is related to the https://github.com/editor-js/nested-list so please open an issue in there

PR #2176, opened 2022-11-25
[Bug] Can not delete nested list first item (#2175)

Rue-pro:
#2175


neSpecc:
Looks like this problem should be resolved on the Nested List side. So please open a PR to the Nested List repo

PR #2186, opened 2022-11-27
fix(block-events): split block with removing of selected text fragment

neSpecc:
I've found a bug in the `next` branch introduced by https://github.com/codex-team/editor.js/pull/2091 ([flipper@line:131](https://github.com/codex-team/editor.js/pull/2091/files#diff-bc434f7301eed82d7c601195fb59e6ed259a5ad29bf3cbbc4b7f797f765256ffR131))
 - Select some text fragment (Inline Toolbar will be opened)
 - Press Enter
 - Nothing happened

The problem is caused by Flipper that is activated with the Inline Toolbar opening. That Flipper stops Enter keydown propagation. 

For now, the Flipper will stop the Enter propagation only if any flippable item is focused

### DIFF
--- a/src/components/flipper.ts
+++ b/src/components/flipper.ts
@@ -279,15 +279,17 @@ export default class Flipper {
     }
 
     if (this.iterator.currentItem) {
+      /**
+       * Stop Enter propagation only if flipper is ready to select focused item
+       */
+      event.stopPropagation();
+      event.preventDefault();
       this.iterator.currentItem.click();
     }
 
     if (_.isFunction(this.activateCallback)) {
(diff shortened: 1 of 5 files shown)


ilyamore88:
Could you update the changelog?

neSpecc:
> Could you update the changelog?

It is just a fix for one of the previous PRs that is already in the changelog:

```
- `Fix` — When Tunes Menu open keydown events can not be handled inside plugins.
```

ilyamore88 (approved):
👍

TatianaFomina (approved):

PR #2192, opened 2022-11-29
Don't minify bundle

oskarer:
Minifying a package published on NPM is generally not needed. The project that installs the package usually has a minifier that takes care of that.

Advantages of not minifying;
* Faster builds
* Easier for consumers to debug any errors in editor.js using browser devtools
* Better support for consumers patching packages using https://www.npmjs.com/package/patch-package

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,6 @@
     "postcss-preset-env": "^6.6.0",
     "rimraf": "^3.0.2",
     "stylelint": "^13.3.3",
-    "terser-webpack-plugin": "^2.3.6",
     "ts-loader": "^7.0.1",
     "tslint": "^6.1.1",
     "typescript": "3.8.3",
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -8,7 +8,6 @@
 
 module.exports = (env, argv) => {
   const path = require('path');
-  const TerserPlugin = require('terser-webpack-plugin');
   const { LicenseWebpackPlugin } = require('license-webpack-plugin');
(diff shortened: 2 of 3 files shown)


neSpecc:
What about bundling for CDN?

gkovats:
I like this, though it kinda goes against all the tools that are all minified. Honestly would be fine if they generated a source map for all builds rather than just development.

oskarer:
> What about bundling for CDN?

@neSpecc It is not needed to minify, jsdelivr.com does it for you when you request it. From https://www.jsdelivr.com/documentation:
image



> I like this, though it kinda goes against all the tools that are all minified. Honestly would be fine if they generated a source map for all builds rather than just development.

@gkovats Can you please elaborate? In my experience the vast majority of packages you install through npm is not minified. Or if you're thinking of using the package through a CDN, in that case jsdelivr is handling that, providing you with source maps. See above answer and this SO thread for more context; https://stackoverflow.com/a/48673965/2014510.

oskarer:
For example when going to the already published CDN version (https://cdn.jsdelivr.net/npm/@editorjs/editorjs@2.26.5/dist/editor.min.js) you see this;

image

gkovats:
Was just saying for our project, we did run into the same problem with bundling, ended up just concatenating editor.js files. In truth think it's best to include all in your dist - minified, expanded, source maps. Just that if they do this for core, it should / would need to be done for editor.js tools as well. They're all minified as well.

PR #2195, opened 2022-12-01
feat: Add toggle group support

TatianaFomina:
Added ability to specify string key as a value of block tune's `toggle` property. It allows to create toggle groups that behave like radio buttons 

https://user-images.githubusercontent.com/31101125/205109875-b8fd185c-033e-4cd4-b5b1-9c00bd640cab.mov

### DIFF
--- a/test/cypress/tests/utils/popover.spec.ts
+++ b/test/cypress/tests/utils/popover.spec.ts
@@ -234,7 +234,7 @@ describe('Popover', () => {
     });
   });
 
-  it('should hightlighted as active the item with toggle property set to unique string key once clicked', () => {
+  it('should toggle item if it is the only item in toggle group', () => {
     const items: PopoverItem[] = [
       {
         icon: 'Icon',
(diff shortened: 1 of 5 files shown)


ilyamore88 (approved):

test/cypress/tests/utils/popover.spec.ts
```diff
@@ -185,4 +185,79 @@ describe('Popover', () => {
...
+        .click()
+        .should('have.class', 'ce-popover__item--active');
+
+      /** Check first item became not active */
+      cy.get('[data-item-name=testItem1]')
+        .should('not.have.class', 'ce-popover__item--active');
+    });
+  });
+
+  it('should hightlighted as active the item with toggle property set to unique string key once clicked', () => {
```
neSpecc:
```suggestion
  it('should toggle item if it is the only item in toggle group', () => {
```

neSpecc (approved):

Issue #2153, opened 2022-10-21
💡 Toolbar position should be relative to block


sadikyalcin:
Problem:

I'm using `editorjs` with/for different layouts. In certain scenarios I have blocks side by side, grid, scrollable horizontal etc... When there are two blocks side by side, it's hard to get to the toolbar for the block you want as the focus changes the moment hover over the other block. I can get around it by not entering the block modal but I can't ask users to do that. Red bordered block is focused:

<img width="1557" alt="Screenshot 2022-10-21 at 16 24 35" src="https://user-images.githubusercontent.com/8109619/197231943-f7dad659-bd71-4614-a2ba-904e7be92deb.png">

Proposed solution:

The toolbar container position should be relative to the block offset rather than the `editorjs` cotainer (`codex-editor`).

<img width="1464" alt="Screenshot 2022-10-21 at 16 24 29" src="https://user-images.githubusercontent.com/8109619/197231960-776fc7fb-377a-4dea-a39c-5391280897c3.png">

navot-tomorrow:
any update on that one?

PR #2201, opened 2022-12-03
Feature/2153 relative toolbar position

sadikyalcin:
As described in #2153, in certain use cases, different layouts may be applied to blocks; i.e a grid layout. This will make it very difficult to get to the toolbar of the desired block because focused block changes on mouse enter:

https://user-images.githubusercontent.com/8109619/205445843-ff0cb2fd-a652-4bd2-8409-bab3110a8c06.mov

With the changes in this branch, the toolbar is absolutely position to the left of a block content as well as the top. The changes should not affect left positioning at all if the block layout is untouched.

Screenshot 2022-12-03 at 14 33 05

Screenshot 2022-12-03 at 14 32 38

Screenshot 2022-12-03 at 14 32 44

With a grid layout;

https://user-images.githubusercontent.com/8109619/205446258-e4ba5eb0-345b-4cb4-af5d-67836e7b6487.mov

### DIFF
--- a/src/components/modules/toolbar/index.ts
+++ b/src/components/modules/toolbar/index.ts
@@ -457,6 +457,13 @@ export default class Toolbar extends Module<ToolbarNodes> {
    * @param block - block to move Toolbar near it
    */
   private assignToolbarLeftPosition(block: Block = this.Editor.BlockManager.currentBlock): void {
+    /**
+     * If no Block is selected as a Current
+     */
+    if (!block) {
+      return;
+    }
+
     const targetBlockHolder = block.holder;
     const blockContentNode = targetBlockHolder.querySelector(`.${Block.CSS.content}`) as HTMLElement | null;
 
--- a/src/components/modules/toolbar/index.ts
+++ b/src/components/modules/toolbar/index.ts
@@ -265,6 +265,8 @@ export default class Toolbar extends Module<ToolbarNodes> {
      */
     this.nodes.wrapper.style.top = `${Math.floor(toolbarY)}px`;
 
+    this.assignToolbarLeftPosition(block);
+
     /**
      * Do not show Block Tunes Toggler near single and empty block
      */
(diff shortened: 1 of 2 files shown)


neSpecc:
Thanks, @sadikyalcin it works fine. Could you write tests for these cases:

1) Toolbar should be moved to the left coordinate of the hovered Block on the single-column layout
2) Toolbar should be moved to the left coordinate of the hovered Block on the multi-column layout
3) Toolbar should be moved to the left coordinate of the current Block on thin (mobile) screens

src/components/modules/toolbar/index.ts
```diff
@@ -492,6 +515,16 @@ export default class Toolbar extends Module<ToolbarNodes> {
 
         this.moveAndOpen(data.block);
       });
+
+      /**
+       * Need to move toolbar on resize because we are now moving it to te left of a Block
+       * This is needed because max-width is removed from Toolbar content to accommodate different block layouts
+       */
+      this.readOnlyMutableListeners.on(window, 'resize', () => {
+        this.assignToolbarLeftPosition(this.hoveredBlock);
```
neSpecc, line 524: why not `moveAndOpen()?`

src/components/modules/toolbar/index.ts
```diff
@@ -492,6 +515,16 @@ export default class Toolbar extends Module<ToolbarNodes> {
 
         this.moveAndOpen(data.block);
       });
+
+      /**
+       * Need to move toolbar on resize because we are now moving it to te left of a Block
+       * This is needed because max-width is removed from Toolbar content to accommodate different block layouts
+       */
+      this.readOnlyMutableListeners.on(window, 'resize', () => {
```
neSpecc, line 523: we already have the window resize handler in the `Ui` module. Maybe you could use it?

navot-tomorrow: any update on that one?

sadikyalcin:
I'm sorry peeps, I've been too busy with work and never had the chance to pick up on this. I have been using this in prod and have no issues. Would be great if someone could pick it up and implement the changes required and get it merged to main.

PR #2203, opened 2022-12-03
Confirmation delete label change to red color

julianramireze:
When confirmation delete is active, the text change to white and this can't be looked, i change this to red color(same of icon) and look good newly.


neSpecc:
I can't reproduce the problem.
image

neSpecc:
Not reproduced.

Issue #2209, opened 2022-12-09
[Bug] Typescript compiler errors when importing editorjs in SvelteKit project (preserveValueImports)


rajwilkhu:
Describe a bug.

Steps to reproduce:
1. Create a sveltekit project
2. install editorjs as an npm depenency
3. import editorjs as 
```typescript
import type EditorJS from '@editorjs/editorjs';
```
5. run linter with default rules and you get the following linter error
```
Error: 'BlockToolData' is a type and must be imported using a type-only import when 'preserveValueImports' and 'isolatedModules' are both enabled.
       import {BlockToolData} from '../tools';
```

Expected behavior:
Typescript compiler should be compile project without errors

Screenshots:

Device, Browser, OS:

Editor.js version: 2.26.2

Plugins you use with their versions: None yet

paolopiaggio:
+1 👍 
same error here, could be related to #1432

PR #2210, opened 2022-12-09
fix(types): fix typescript compiler errors under strict mode

rajwilkhu:
This is relate to issue #2209

### DIFF
--- a/types/configs/conversion-config.ts
+++ b/types/configs/conversion-config.ts
@@ -1,4 +1,4 @@
-import {BlockToolData} from '../tools';
+import type {BlockToolData} from '../tools';
 
 /**
  * Config allows Tool to specify how it can be converted into/from another Tool


rajwilkhu:
@here, any way to get this reviewed? It's a minor change? @neSpecc @gohabereg @TatianaFomina @ilyamore88

neSpecc:
I think it is not the only error under strict mode. We should enable it in our tsconfig and fix all errors complexly.

neSpecc:
But we can accept your change and release a patch anyway if your work relies on this.

neSpecc:
Could you add a changelog, please?

CODE-BR3AKER:
@neSpecc please merge this PR asap this issue is breaking our prod, even the --skipLibCheck has no effect on the .ts files , we have to disbale typecheck for the entire project which is bad !

neSpecc:
Thanks for the fix. Unfortunately, I was not able to push to your branch due to disabled access for maintainers.

So I'm focusing on the similar PR [#2264](https://github.com/codex-team/editor.js/pull/2264) and closing this PR as duplicate.

PR #2217, opened 2022-12-13
fix(paste): handle pasteConfig with tag names specified in upper case

### LINKED ISSUE #2208: [Bug] Pasted HTML lists are empty
Whenever I copy a HTML list (either a `<ul>` or `<ol>`) from the web, and paste it into an Editor.js instance, an empty list gets created. The list items are not populated.

This only seems to happen when pasting formatted text from an external application – i.e. where the clipboard data is in `text/html` format. Copying within or between Editor.js instances works as expected – presumably this is due to the clipboard data having the [custom mime type](https://github.com/codex-team/editor.js/blob/811a41366bbf8201105c12ef92f8c8abb907f028/src/components/modules/paste.ts#L115-L116) `application/x-editor-js` and therefore being processed differently.

### Steps to reproduce

1. Open the demo editor at [editorjs.io](https://editorjs.io/).
2. Copy this list directly from the GitHub issue.
3. Paste it into the editor. It will render as an empty ordered list.

### Expected behavior

I expect the `<li>` list items to be populated within the list block's `data.items` attribute. Instead, the `items` array is empty.

### Screenshots

https://user-images.githubusercontent.com/7735945/206200924-e4a22bd3-f797-4cce-8f7a-1b77f998ce25.mov

### Device, Browser, OS

I'm using macOS Monterey. I've been able to recreate the bug in Chrome, Brave, Firefox and Safari.

### Editor.js version

In local development, I've had issues with the current release, version 2.26.1. Presumably this is also what's running on the [Editor.js homepage demo](https://editorjs.io/).

### Plugins you use with their versions

I can recreate the issue in an Editor.js instance with only the default header, paragraph and list tools enabled.

robonetphy:
resolve #2208

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+### 2.26.3
+
+- `Fix` — *Paste Module* — `pasteConfig().tags` regression removed [#2208](https://github.com/codex-team/editor.js/issues/2208).
 
 ### 2.26.2
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### 2.26.3
 
-- `Fix` — *Paste Module* — `pasteConfig().tags` regression removed [#2208](https://github.com/codex-team/editor.js/issues/2208).
+- `Fix` — *Paste Module* — fix for a problem with specifying of `pasteConfig().tags` in upper case  [#2208](https://github.com/codex-team/editor.js/issues/2208).
 
 ### 2.26.2
 
--- a/test/cypress/tests/api/tools.spec.ts
+++ b/test/cypress/tests/api/tools.spec.ts
@@ -213,12 +213,12 @@ describe('Editor Tools Api', () => {
             icon: ICON,
             name: 'testToolTune',
 
-            onActivate: (): void => {},
+            onActivate: (): void => { },
           };
         }
 
         /** Save method stub */
-        public save(): void {}
+        public save(): void { }
 
         /** Renders a block */
         public render(): HTMLElement {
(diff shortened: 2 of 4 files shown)


test/cypress/tests/api/tools.spec.ts
```diff
@@ -526,6 +526,79 @@ describe('Editor Tools Api', () => {
...
+        let pastedElement;
+
+        /**
+         * Test tool with pasteConfig.tags specified
+         */
+        class TestListTool {
+          /** config specified handled tag */
+          public static get pasteConfig(): PasteConfig {
+            return {
+              tags: ['OL', 'LI'], // only tag name specified. Attributes should be sanitized
```
neSpecc:
```suggestion
              tags: ['OL', 'LI'], // tag names specified in upper case
```

test/cypress/tests/api/tools.spec.ts
```diff
@@ -526,6 +526,79 @@ describe('Editor Tools Api', () => {
...
+            testListTool: TestListTool,
+          },
+        });
+
+        cy.get('[data-cy=editorjs]')
+          .get('div.ce-block')
+          .click()
+          .paste({
+            // eslint-disable-next-line @typescript-eslint/naming-convention
+            'text/html': '<ol start="50"><li>Orderd List</li><li>Unorderd List</li></ol>', // all attributes should be sanitized
```
neSpecc:
```suggestion
            'text/html': '<ol start="50"><li>Orderd List</li><li>Unorderd List</li></ol>', // all attributes should be sanitized, <li> should be preserved
```

docs/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+### 2.26.3
+
+- `Fix` — *Paste Module* — `pasteConfig().tags` regression removed [#2208](https://github.com/codex-team/editor.js/issues/2208).
```
```suggestion
- `Fix` — *Paste Module* — fix for a problem with specifying of `pasteConfig().tags` in upper case  [#2208](https://github.com/codex-team/editor.js/issues/2208).
```

ilyamore88 (approved):

neSpecc (approved):

ollietreend:
Thanks for the fix @robonetphy! ✨🚀

Issue #2221, opened 2022-12-17
Define new "block-updated" event. 


krisnik:
The question.

> How to differentiate Block Update (vs) Block insert / delete. 

Why and how the question has come up.

When block is updated using `block.update(id, BlockToolData)` - EditorJS `onChange` api gets called twice. once with `blocked-removed` and later with `block-added` events. 

`block-removed` and `block-added` events are also fired when the blocks are deleted / added to the EditorJS instance. 

Hence, from the onChange response, it is not possible to differentiate if block's content is updated or a new block is added / existing block is removed.

lijie1129:
I think this is because the `update` function is implemented by calling the `insert` function and specifying the `replace` parameter as `true`. 

Therefore, there are two final actions: 

* delete the old block
* insert the block holding the new data in the original location.

This is why an update causes the `onChange` API gets called twice.


**_This is the definition of the update function:_**
https://github.com/codex-team/editor.js/blob/870e265af0e2114c0a1504a70148ea3c5e28d959/src/components/modules/api/blocks.ts#L291-L311

**_This is the definition of the insert function:_**
https://github.com/codex-team/editor.js/blob/870e265af0e2114c0a1504a70148ea3c5e28d959/src/components/modules/blockManager.ts#L266-L322


It seems that the simple way is:

1. Provide a new function in the **_blockManager.ts_** file to update the block's data according to the specified block ID. (This function locates the correct block in `this._blocks` by ID and directly replace it.)
2. Then, directly call the new function defined above in the update function of the **_api/blocks.ts_** file.
3.  Finally, trigger the `onChange` API once.

lijie1129:
## Why does it need to support a single event

Recently, I tried to make a binding between the [YJS](https://github.com/yjs/yjs) and EditorJS.

This is to realize an idea. Please refer to [here](https://discuss.yjs.dev/t/editor-js-bindings/1349/5?u=jarone).

### First
Since the synchronization EditorJS changes to the data model (YJS doc) is triggered by `onChange`.

It is necessary to minimize unnecessary events and avoid performance waste.

### Second
In the scenario of multi-person collaborative editing, directly removing and replacing the DOM of the entire block will cause the collaborator to lose the input focus, which will cause experience problems. 

Therefore, it is a good way to delegate the work of updating a specific block content to a specific plug-in.

lijie1129:
> ## Why does it need to support a single event
> 
> 
> 
[...]

PR #2234, opened 2023-01-09
feat: support new "block-updated" event. (#2221)

lijie1129:
#2221

### DIFF
--- a/src/components/block/api.ts
+++ b/src/components/block/api.ts
@@ -105,6 +105,16 @@ function BlockAPI(
     },
 
     /**
+     * Update Block data
+     *
+     * @param {BlockToolData} data - data to update
+     * @returns {Promise<boolean>}
+     */
+    update(data: BlockToolData): Promise<boolean> {
+      return block.update(data);
+    },
+
+    /**
      * Validate Block data
      *
      * @param {BlockToolData} data - data to validate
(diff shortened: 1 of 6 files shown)


lijie1129:
## Why does it need to support a single event

Recently, I tried to make a binding between the YJS and EditorJS.

This is to realize an idea. Please refer to [here](https://discuss.yjs.dev/t/editor-js-bindings/1349/5?u=jarone).

### First
Since the synchronization EditorJS changes to the data model (YJS doc) is triggered by `onChange`.

It is necessary to minimize unnecessary events and avoid performance waste.

### Second
In the scenario of multi-person collaborative editing, directly removing and replacing the DOM of the entire block will cause the collaborator to lose the input focus, which will cause experience problems. 

Therefore, it is a good way to delegate the work of updating a specific block content to a specific plug-in.

neSpecc:
Thanks, @lijie1129, it is nice feature and implementation. 

We're working on collaborative editing, but we have not designed Tools API change yet. So adding a `update` method should be agreed. 

Do you have a workaround for a while?

lijie1129:
Hi, @neSpecc 

Thank you for your suggestion. 🤝

I will fix this pr later.

lijie1129:
This issue has been resolved.

PR #2263, opened 2023-02-01
feat: toolbox, add to an item a "description" setting

alinceDev:
Need to have a description for each item in the toolbox like notion does.
Here is an example :
https://global-uploads.webflow.com/5e1331324b55b0fe5f41fe63/638dcb89ba38ef59fbdeed26_How%20to%20create%20a%20new%20page%20in%20Notion.png


neSpecc:
Nice to have feature. It requires some UI design investigation, then could be included

PR #2284, opened 2023-02-21
Give delete key a chance

gkovats:
Not a TS dev naturally, this is definitely a copy / paste from the backspace() method. Basically just adding "delete" key behavior to base editor, a variant of backspace.
- when multiple blocks are selected, it's the same behavior
- when on a blank line, same behavior
- when at the end of a block and the next block is of same type _and_ supports merge, merge the next block to current

Our environment is very Windows heavy, and delete key is more prominent in Win UIs, so this was missing from the experience. Perhaps the blockEvents `delete()` logic is a little repetitive now, but this helps bridge the gap for folks expecting standard line navigation / merging. Tossing up for what it's worth.


gorenburg:
Yes, bumped into the same feature request from users. Give it a chance, @codex-team!

neSpecc:
Resolved by #2402

PR #2300, opened 2023-03-13
Use Vite builder

talyguryn:
- use vite builder
- removed unused packages


index.html
```diff
@@ -0,0 +1,326 @@
...
+  <script src="https://cdn.jsdelivr.net/npm/@editorjs/marker@latest"></script><!-- Marker -->
+  <script src="https://cdn.jsdelivr.net/npm/@editorjs/inline-code@latest"></script><!-- Inline Code -->
+
+  <!-- Load Editor.js's Core -->
+  <!-- <script src="../dist/editor.js" onload="document.getElementById('hint-core').hidden = true;"></script> -->
+
+
+  <!-- Initialization -->
+  <script type="module">
+    import EditorJS from './src/codex.ts';
```
neSpecc, line 99: looks like this file is a mix of entry point and example.html. Maybe we don't need index.html as entry point?

package.json
```diff
@@ -40,64 +38,51 @@
...
   },
   "devDependencies": {
-    "@babel/core": "^7.9.0",
-    "@babel/plugin-transform-runtime": "^7.9.0",
-    "@babel/polyfill": "^7.8.7",
-    "@babel/preset-env": "^7.9.5",
-    "@babel/preset-typescript": "^7.13.0",
-    "@babel/register": "^7.9.0",
-    "@babel/runtime": "^7.9.2",
+    "@babel/plugin-proposal-decorators": "^7.21.0",
```
neSpecc: do we need `@babel/plugin-proposal-decorators` ?

package.json
```diff
@@ -40,64 +38,51 @@
...
+    "@babel/register": "^7.21.0",
+    "@codexteam/icons": "^0.3.0",
     "@codexteam/shortcuts": "^1.1.1",
     "@cypress/code-coverage": "^3.9.2",
     "@cypress/webpack-preprocessor": "^5.6.0",
     "@editorjs/code": "^2.7.0",
     "@editorjs/delimiter": "^1.2.0",
     "@editorjs/header": "^2.7.0",
+    "@editorjs/paragraph": "^2.9.0",
```
neSpecc, line 47: it is a prod dependency

package.json
```diff
@@ -40,64 +38,51 @@
...
     "@editorjs/simple-image": "^1.4.1",
+    "@originjs/vite-plugin-require-context": "^1.0.9",
     "@types/node": "^14.14.35",
-    "@types/webpack": "^4.41.12",
-    "@types/webpack-env": "^1.15.2",
-    "babel-loader": "^8.1.0",
-    "babel-plugin-add-module-exports": "^1.0.0",
-    "babel-plugin-class-display-name": "^2.1.0",
-    "babel-plugin-istanbul": "^6.0.0",
+    "@vitejs/plugin-react": "^3.1.0",
```
neSpecc: why do we need this?

.postcssrc.yml
```diff
@@ -22,20 +18,6 @@ plugins:
       - 'last 2 versions'
       - '> 1%'
 
-    # Instruct all plugins to omit pre-polyfilled CSS
-    # https://github.com/csstools/postcss-preset-env#preserve
-    preserve: false
```
neSpecc, line 27: If we won't replace css variables, we need to move them from :root {} to the editor scope to prevent confllicts

package.json
```diff
@@ -40,64 +38,51 @@
...
     "eslint": "^8.28.0",
     "eslint-config-codex": "^1.7.1",
-    "eslint-loader": "^4.0.2",
     "eslint-plugin-chai-friendly": "^0.7.2",
     "eslint-plugin-cypress": "^2.12.1",
-    "extract-text-webpack-plugin": "^3.0.2",
-    "license-webpack-plugin": "^2.1.4",
+    "html-janitor": "^2.0.4",
     "mkdirp": "^1.0.4",
+    "nanoid": "^4.0.1",
```
neSpecc: prod dependency

package.json
```diff
@@ -40,64 +38,51 @@
...
     "cypress": "^6.8.0",
     "cypress-intellij-reporter": "^0.0.6",
     "eslint": "^8.28.0",
     "eslint-config-codex": "^1.7.1",
-    "eslint-loader": "^4.0.2",
     "eslint-plugin-chai-friendly": "^0.7.2",
     "eslint-plugin-cypress": "^2.12.1",
-    "extract-text-webpack-plugin": "^3.0.2",
-    "license-webpack-plugin": "^2.1.4",
+    "html-janitor": "^2.0.4",
```
neSpecc, line 59: prod dependency

src/components/core.ts
```diff
@@ -292,9 +292,15 @@ export default class Core {
...
+
+    // console.log(`[this.moduleInstances]`, this.moduleInstances);
+
+
     await modulesToPrepare.reduce(
       (promise, module) => promise.then(async () => {
-        // _.log(`Preparing ${module} module`, 'time');
+        _.log(`Preparing ${module} module`, 'time');
+
+        // console.log(`[modulesToPrepare] this.moduleInstances[module]`, this.moduleInstances[module]);
```
neSpecc: pls remove logs

src/components/modules/api/blocks.ts
```diff
@@ -10,6 +10,8 @@ import Block from '../../block';
  * provides with methods working with Block
  */
 export default class BlocksAPI extends Module {
+  public static readonly displayName: string = 'BlocksAPI';
```
neSpecc: Can we get rid of this? It has no sense from the architecture perspective

vite.config.js
```diff
@@ -0,0 +1,92 @@
...
+
+  plugins: [
+    banner(BANNER),
+    ViteRequireContext(),
+    cssInjectedByJsPlugin(),
+    react({
+      babel: {
+        plugins: [
+          [
+            '@babel/plugin-proposal-decorators',
```
neSpecc: why do we need this?

index.html
ilyamore88: It's an entry point for `yarn dev`. Anyway, @talyguryn, why don't you use existing files?

package.json
ilyamore88: +1

package.json
```diff
@@ -40,64 +38,49 @@
...
+    "vite": "^4.1.4",
+    "vite-plugin-babel": "^1.1.3",
+    "vite-plugin-banner": "^0.7.0",
+    "vite-plugin-css-injected-by-js": "^3.0.1"
   },
   "collective": {
     "type": "opencollective",
     "url": "https://opencollective.com/editorjs"
-  },
-  "dependencies": {
```
ilyamore88, line 96: Dependencies field helps user to understand what he will get in a bundle

vite.config.js
```diff
@@ -0,0 +1,79 @@
...
+const NODE_ENV = process.argv.mode || 'development';
+const VERSION = pkg.version;
+const BANNER = `
+Editor.js
+
+@version ${VERSION}
+@licence Apache-2.0
+@author CodeX <https://codex.so>
+
+@uses html-janitor
```
ilyamore88: Will you always add new packages with Apache licenses to this banner manually?

vite.config.js
```diff
@@ -0,0 +1,79 @@
+import path from 'path';
+
+import cssInjectedByJsPlugin from 'vite-plugin-css-injected-by-js';
+import ViteRequireContext from '@originjs/vite-plugin-require-context';
+import license from 'rollup-plugin-license';
+import banner from 'vite-plugin-banner';
```
ilyamore88: Is it possible to use https://github.com/mjeanroy/rollup-plugin-license#banner?

package.json
talyguryn: i don't know about custom interpretations, but "dependencies" used first of all to install additional packages.

editor.js is used as bundle which has no dependencies. that's why you can use in on the page with a single script required.

at this time "dependencies" section provides installing a bunch of packages that will not be used:

image

index.html
ilyamore88: Also, is it possible to provide the name of a template for this command? Like `yarn dev example-rtl.html`. And by default use `example-dev.html`.

ilyamore88: I think it's possible because as I remember you can get this variable in a config file.

index.html
ilyamore88, line 1: I still don't understand why you don't use example files. Why do you create one more example file?

ilyamore88 (commented):
Update your branch pls.

src/components/core.ts
```diff
@@ -5,27 +5,22 @@ import { EditorModules } from '../types-internal/editor-modules';
...
-   * - does not starts with _
-   */
-  if (filename.match(/^\.\/[^_][\w/]*\.([tj])s$/)) {
-    modules.push(contextRequire(filename));
-  }
+const modules: { name: string, module: Module }[] = Object.keys(Modules).map((key) => {
+  return {
+    name: key,
+    module: Modules[key],
+  };
```
neSpecc: redundant wrapper, you can use `Modules` directly

src/types-internal/editor-modules.d.ts
neSpecc, line 1: this file is not needed anymore since it duplicated `src/modules/index.ts`

neSpecc (changes requested):

neSpecc (approved):

ilyamore88 (changes requested):
Could you please add a changelog?

ilyamore88 (approved):

PR #2302, opened 2023-03-15
Add block tool name to block class wrapper

wmnnd:
This PR adds the tool name to block wrapper elements. This makes it easy to apply custom styles to existing/third-party blocks where it’s not possible to add such classes to the block itself.

In the case of the paragraph block, this would mean that the block receives the `ce-block--type-paragraph` class in addition to `ce-block`.

I’ve found this change tremendously helpful when working on integrating Editor.js into [Keila](https://github.com/pentacent/keila).


src/components/block/index.ts
```diff
@@ -821,7 +821,7 @@ export default class Block extends EventsDispatcher<BlockEvents> {
    * @returns {HTMLDivElement}
    */
   private compose(): HTMLDivElement {
-    const wrapper = $.make('div', Block.CSS.wrapper) as HTMLDivElement,
```
neSpecc, line 824: I'd suggest to use the `data-tool` attribute instead of class

neSpecc:
Useful thing. Please, add a changelog as well

neSpecc:
> This makes it easy to apply custom styles to existing/third-party blocks where it’s not possible to add such classes to the block itself.

Could you provide some real examples? Why not using tools' class names?

PR #2303, opened 2023-03-15
Make sure hovered block belongs to current instance of Editor.js

wmnnd:
This change ensures that the returned block belongs to the current instance of the editor, avoiding issues with nested instances.

When building a layout block for [Keila](https://github.com/pentacent/keila) (you can see it in action [here](https://fosstodon.org/@keila/109712314246003046)), I noticed issues with hover detection when there were nested instances of Editor.js. This PR resolves this by making sure detected blocks are always part of the current instance.


neSpecc:
Hi, @wmnnd. Thanks for the PR.

Editor.js does not support a multicolumn layout at the moment. The nesting of the editor instances seems like a messy workaround, so I'm not sure that we should add an additional logic for that. Moreover, if it reduces performance like looping through all the blocks and their children on each mousemove.

It would be better to implement a multicolumn layout inside the Editor. I think this feature could be added in the future.

Bytheway Keila looks cool, will be glad to test it.

Issue #2205, opened 2022-12-04
💡 Optional mutation observer for block tools


medzhidov:
Sometimes it needs to stop mutation observer on blocks, because they are not created to be observed. For example, i create an custom tool block that has several logic, attributes/classes/styles and i want to save only content that i dispatched via `block.dispatchChange();`, because mutation observer trigger "didMutated" when content is not really edited (for example trigger class on image by hovering).

I already solved this by edited code, so there is my suggested solution:

```typescript
// src/components/tools/base.ts:80 - add
/**
   * Is mutation observer enabled for Tool
   */
  IsEnabledMutationObserver = 'isEnabledMutationObserver',
```

```typescript
// src/components/tools/block.ts:70 - add
/**
   * Returns true if Tool supports linebreaks
   */
  public get isEnabledMutationObserver(): boolean {
    return this.constructable[InternalBlockToolSettings.IsEnabledMutationObserver] !== false;
  }
```

```typescript
//src/components/block/index.ts:267 - replace
    if (tool.isEnabledMutationObserver) {
      this.mutationObserver = new MutationObserver(this.didMutated);
    }

//src/components/block/index.ts:295 - replace
this.mutationObserver?.observe(

//src/components/block/index.ts:716 - replace
this.mutationObserver?.disconnect();
```

So now we can add in our block tool a static method and mutation will be ignored for this block:
```typescript
export default class SimpleImage {
   constructor(){...}
   
   static get isEnabledMutationObserver() {
        return false;
   }
   
   render() {...}
```

It's 2 am so i may forget something but it works perfectly in my project now.

Bobsans:
I have related problem too. I write tool with map and map script update block class many times per second. So if i call `block.dispatchChange()` from tool - debouncer ignore this call, and data saved only if block lose focus (observer unsubscribes). Please add possibility to disable observer for tool

lucas-varela:
@medzhidov and @Bobsans, I've implemented a way to disable the mutation observation, check out this PR: #2317 

Also, I've created a branch that includes the `dist` folder built with these changes: [https://github.com/lucas-varela/editor.js/tree/build/optional-mutation-observer](https://github.com/lucas-varela/editor.js/tree/build/optional-mutation-observer)

This branch can be used on an `npm install` or `yarn add`, like:

```
npm install https://github.com/lucas-varela/editor.js.git#build/optional-mutation-observer
```

[...]

PR #2317, opened 2023-03-23
feat(block): Add option to disable mutation observation

lucas-varela:
This PR adds the option to disable updates when DOM mutations are observed or inputs are changed inside the element tree of a rendered block. Disabling this behavior is as simple as adding a getter to the block's class:

```typescript
class ExampleBlock extends BlockTool {
  public static get shouldUpdateOnMutation(): boolean {
    return false;
  }
}
```

After this, the block should use `dispatchChange()` whenever it needs to trigger an update.

### Context

I'm currently implementing custom Blocks on a project using React. It uses a portal to render React components inside the wrapper element returned by the `render()` method, and it works quite well.

However, every time React updates the DOM inside this portal, the EditorJS instance triggers an update since its MutationObserver is always listening for mutations inside the block. This is undesirable for performance reasons, and this behavior can actually cause loops when interacting with React states.

To fix this, I've added `shouldUpdateOnMutation` and every time my internal React component updates its internal data state, it calls `dispatchChange()` to tell EditorJS that a change occurred.

The code in this PR is heavily based on the solution found in #2205, created by @medzhidov (thanks, by the way!). I've only changed the name of the option, since my implementation also disables listening for input changes in the block's element tree.

Side note: I intend to release the code that allows users to implement Blocks as React components as a library in the near future. I'll definitely let people know when I do so.

### DIFF
--- a/src/components/block/index.ts
+++ b/src/components/block/index.ts
@@ -748,8 +748,15 @@ export default class Block extends EventsDispatcher<BlockEvents> {
    * Is fired when Block will be unselected
    */
   public willUnselect(): void {
-    this.mutationObserver?.disconnect();
-    this.removeInputEvents();
+    /**
+     * There's no need to disconnect the observer or
+     * remove listeners if shouldUpdateOnMutation is false,
+     * since they weren't set up in the first place.
+     */
+    if (this.tool.shouldUpdateOnMutation) {
+      this.mutationObserver.disconnect();
+      this.removeInputEvents();
+    }
   }
 
   /**
--- a/src/components/block/index.ts
+++ b/src/components/block/index.ts
@@ -199,11 +199,11 @@ export default class Block extends EventsDispatcher<BlockEvents> {
   private mutationObserver: MutationObserver | undefined;
 
   /**
-   * Debounce Timer
+   * Minimum time to wait between mutation events, in milliseconds.
    *
    * @type {number}
    */
-  private readonly modificationDebounceTimer = 450;
+  private readonly modificationDebounceTimer = _.modificationDebounceTimer;
 
   /**
    * Is fired when DOM mutation has been happened
@@ -309,7 +309,7 @@ export default class Block extends EventsDispatcher<BlockEvents> {
     this.tool = tool;
     this.toolInstance = tool.create(data, this.blockAPI, readOnly);
 
-    if (tool.isMutationObserverEnabled) {
+    if (tool.shouldUpdateOnMutation) {
       this.mutationObserver = new MutationObserver(this.didMutated);
     }
 
--- a/src/components/block/index.ts
+++ b/src/components/block/index.ts
@@ -194,9 +194,9 @@ export default class Block extends EventsDispatcher<BlockEvents> {
   /**
    * Mutation observer to handle DOM mutations
    *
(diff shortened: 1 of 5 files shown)


neSpecc:
> However, every time React updates the DOM inside this portal, the EditorJS instance triggers an update since its MutationObserver is always listening for mutations inside the block.

Have you tried adding the `data-mutation-free` to some elements that are changing in this case?

medzhidov:
> > However, every time React updates the DOM inside this portal, the EditorJS instance triggers an update since its MutationObserver is always listening for mutations inside the block.
> 
> Have you tried adding the `data-mutation-free` to some elements that are changing in this case?

That didn't work for me, so this is the reason why I wrote this code

neSpecc:
Could you provide an example of exact dom changes in your tool that triggers unwanted `onChange`? Maybe we can improve logic related to the `data-mutation-free` to handle that.

lucas-varela:
> Have you tried adding the `data-mutation-free` to some elements that are changing in this case?

@neSpecc `data-mutation-free` works, but since **all** mutated nodes need to have this property and React might mutate various nodes when rendering, including nodes unrelated to rendered content, the solution would be adding this property to a lot of elements depending on the block's complexity. I believe this is undesirable since it has the potential to pollute an application code, specially in complex blocks.

> Could you provide an example of exact dom changes in your tool that triggers unwanted `onChange`? Maybe we can improve logic related to the `data-mutation-free` to handle that.

Absolutely. I'll create an example in the next few days, and I'll mention you in the comment when I do so.

Regarding improving `data-mutation-free`'s logic, it's a possibility. In a way, this PR adds a `data-mutation-free="deep"` option of sorts.

Instead of using a class property (like I did in this PR), we could simply update `data-mutation-free` to allow developers to choose between ignoring mutations on a single element (which could be the default for backwards compatibility) or ignoring mutations on an element and all of their children (like the example above, `data-mutation-free="deep"`). 

We could also just add a new property to allow the second option, like `data-deep-mutation-free`.

Which do you think is the better option?

bettysteger:
That would be so nice if that worked, i am rendering a few custom Vue components as Block Tools, for example a Video.js player, when i just play the video the editorjs gets a "block-changed" event and that's simply not true.

bettysteger:
@lucas-varela can you please update your PR and resolve the conflicts - so i can use your fork ;)

lucas-varela:
@bettysteger I'm currently busy working with infrastructure matters on my project, and I sadly don't have the time necessary to update this PR in the foreseeable future, since there's been plenty of changes since I've opened it.

However, one of my peers recently fixed an unrelated bug on our project by updating editor.js to version 2.28.2, and by doing so, they switched from my fork (with the mutation observer fix) to the official package. Everything seems to be working fine, and DOM mutations caused by React aren't triggering dozens of editor.js updates or loops like it did before.

I believe some editor.js update might have fixed the behavior, but I didn't have an opportunity to take a closer look at the changes to make sure this behavior is 100% fixed and won't cause trouble when interacting with libs like React and Vue in the future, but I'm planning to do so. But, again, I'm not sure when I'll do this.

For now, please try to update to v2.28.2 if you didn't already do so because it might fix this behavior. Good luck!

PR #2318, opened 2023-03-23
Updated all devDependancies in package.json and webpack to support node 18

rogueturnip:
The same work needs to be done with all the plugins.  I've done this and can submit for each.

Only thing not updated was eslint since it was giving issues.


rogueturnip:
found some bugs.  closing for now.

PR #2322, opened 2023-03-26
deps(TypeScript) - upgrade to v5, upgrade ts-loader, fix types error, fix pasteConfig getter wrapper

ilyamore88:
I've upgraded the TypeScript package to the latest version. Also, fixed a bug from #2131.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,8 +7,10 @@
 - `Improvement` — The `onChange` callback won't be triggered only if all mutations contain nodes with the `data-mutation-free` attributes.
 - `Fix` — Resolve compiler error from importing the BlockToolData type.
 - `Fix` — Resolved a problem when document was being scrolled to the beginning after moving up a Block above the viewport.
-- `Improvement` — Package size reduced by removing redundant files
-- `Fix`- Several bugs caused by random browser extensions
+- `Improvement` — Package size reduced by removing redundant files.
+- `Fix`- Several bugs caused by random browser extensions.
+- `Improvement` — *Dependencies* — Upgrade TypeScript to v5.
+- `Fix` — *BlockTool* — `pasteConfig` getter provide `false` into the core if you need to disable preventing default behavior for paste event.
 
 ### 2.26.5
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,7 +10,7 @@
 - `Improvement` — Package size reduced by removing redundant files.
 - `Fix`- Several bugs caused by random browser extensions.
 - `Improvement` — *Dependencies* — Upgrade TypeScript to v5.
-- `Fix` — *BlockTool* — `pasteConfig` getter provide `false` into the core if you need to disable preventing default behavior for paste event.
+- `Fix` — *ToolsAPI* — `pasteConfig` getter with `false` value could be used to disable paste handling by Editor.js core. Could be useful if your tool has its own paste handler.
 
 ### 2.26.5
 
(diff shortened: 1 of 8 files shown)


ilyamore88:
I still need to think about a test case for the fix.

src/components/tools/block.ts
```diff
@@ -157,8 +157,8 @@ export default class BlockTool extends BaseTool<IBlockTool> {
   /**
    * Returns Tool paste configuration
    */
-  public get pasteConfig(): PasteConfig {
-    return this.constructable[InternalBlockToolSettings.PasteConfig] || {};
+  public get pasteConfig(): PasteConfig | false {
```
gohabereg: I'd change PasteConfig itself to include false, just for convenience of reusage

gohabereg (approved):

types/configs/paste-config.d.ts
```diff
@@ -31,3 +31,8 @@ export interface PasteConfig {
    */
   files?: {extensions?: string[], mimeTypes?: string[]};
 }
+
+/**
+ * Alias for PasteConfig with false
+ */
+export type PasteConfig = IPasteConfig | false;
```
neSpecc: We're not using the "I" prefix with interfaces, and also the difference between `pasteConfig` and `iPasteConfig` is not clear. I'd suggest `PasteConfigSpecified` here.

```suggestion
export type PasteConfig = PasteConfigSpecified | false;
```

test/cypress/tests/copy-paste.spec.ts
```diff
@@ -1,13 +1,48 @@
...
+      return {};
+    }
+  }
+
   beforeEach(function () {
     cy.createEditor({
       tools: {
         header: Header,
         image: Image,
+        blockToolWithPasteHandler: BlockToolWithPasteHandler,
```
neSpecc: I think it would be better to create an instance with this tool inside the corresponding test, not before each test case since it is not related to other tests.

docs/CHANGELOG.md
```diff
@@ -7,8 +7,10 @@
 - `Improvement` — The `onChange` callback won't be triggered only if all mutations contain nodes with the `data-mutation-free` attributes.
 - `Fix` — Resolve compiler error from importing the BlockToolData type.
 - `Fix` — Resolved a problem when document was being scrolled to the beginning after moving up a Block above the viewport.
-- `Improvement` — Package size reduced by removing redundant files
-- `Fix`- Several bugs caused by random browser extensions
+- `Improvement` — Package size reduced by removing redundant files.
+- `Fix`- Several bugs caused by random browser extensions.
+- `Improvement` — *Dependencies* — Upgrade TypeScript to v5.
+- `Fix` — *BlockTool* — `pasteConfig` getter provide `false` into the core if you need to disable preventing default behavior for paste event.
```
```suggestion
- `Fix` — *ToolsAPI* — `pasteConfig` getter with `false` value could be used to disable paste handling by Editor.js core. Could be useful if your tool has its own paste handler.
```

test/cypress/tests/copy-paste.spec.ts
```diff
@@ -139,6 +141,75 @@ describe('Copy pasting from Editor', function () {
...
+        });
+
+      cy.get('@onPaste')
+        .should('have.been.calledWithMatch', {
+          defaultPrevented: false,
+        });
+
+      cy.get('@editorInstanceWithBlockToolWithPasteHandler')
+        .then((editorInstance: unknown) => (editorInstance as EditorJS).destroy());
+    });
```
neSpecc: I think it shouldn't be in a test case. Probably, it's not required at all

neSpecc (approved):

PR #2331, opened 2023-04-06
fix: split custom event

### LINKED ISSUE #2230: [Bug] Splitting a block does not create "block-changed" event on the source block.
Describe a bug.

If I click enter in somewhere in the block, a new block gets created from the caret position. 

However for the source block which got split, does not get a "block-changed" event.  

Steps to reproduce:
1. Go to … Add some text to paragraph block. 
2. Click on … Click enter somewhere in middle of the text. A new block with content from the caret position get created
3. … 

Expected behavior:
- Block Added event for new block. ✅ 
- Block Changed event for the block which got split. ❌ 

From the source code, only a insert operation is invoked which results in "block-added" event. 

https://github.com/codex-team/editor.js/blob/43032ebaac6a8b9a67c9bf2ada5adc0f8592d546/src/components/modules/blockManager.ts#L479 
Screenshots:

Device, Browser, OS:

Editor.js version:
2.26.4

HyTekCoop:
Resolves #2230 
Add custom event of changes when block is split.

The problem was that when you press enter in the block and the block was splitted. The message appeared about "block-added", but there was no message about changing the block

Picture 1. Block initial state
image

Picture 2. Bug example
image

Picture 3. After fix
image


neSpecc:
Please, add a test case and a changelog.

Also, attach an issue you're resolving

gohabereg (approved):

HyTekCoop:
It's no longer relevant. Part of the code has been rewritten.

PR #2346, opened 2023-04-22
adding block id for in the class list and as data

slimfadi:
this is useful for external manipulation, for example my use case was to move blocks between different documents

### DIFF
--- a/src/components/block/index.ts
+++ b/src/components/block/index.ts
@@ -814,6 +814,9 @@ export default class Block extends EventsDispatcher<BlockEvents> {
         contentNode = $.make('div', Block.CSS.content),
         pluginsContent = this.toolInstance.render();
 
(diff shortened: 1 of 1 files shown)


src/components/block/index.ts
```diff
@@ -814,6 +814,9 @@ export default class Block extends EventsDispatcher<BlockEvents> {
         contentNode = $.make('div', Block.CSS.content),
         pluginsContent = this.toolInstance.render();
 
+    wrapper.classList.add(Block.CSS.wrapper + "-"+this.id);
+    wrapper.dataset.blockId = this.id;
```
neSpecc, line 818: the more preferred way is to add `data-id` attribute instead.

Also, please add a change log and a test case in `block-ids.cy.ts`

neSpecc:
Resolved by #2404

PR #2349, opened 2023-05-03
chore(onChange): bugfix, batching, improvements of ModificationsObserver

### LINKED ISSUE #2325: Pasting header is broken
After updating the version from 2.27.0-rc.3 to 2.27.0-rc.4, pasting header is not working correctly. Not like other blocks, "text" field in the header gets an empty string when `save(toolsContent)` of the Header plugin is called.

Steps to reproduce:
1. Copy text, including headers. (I attached the example below.)
2. Paste it to EditorJS.
3. Save the content and check if the headers are still there.

Example copy text:

#### header example

paragraph example.

Expected behavior:
The header should be saved.

Screenshots:
image
image

Device, Browser, OS: Macbook Air M2, Chrome, OSX

Editor.js version: 2.27.0-rc.4

Header, Paragraph

neSpecc:
### This PR contains several improvements of ModificationObserver and related code.

1. Mutation Observer moved from each Block to the common editor element.
    - Now observer belongs to the ModificationObserver module.  
    - It is connected to `.ce-editor__redactor` blocks wrapper.
    - The ModificationObserver emits the `RedactorDomChanged` event to the EventBus.
    - Each Block now have access to the Editor's Event Bus
    - Each Block listens for the `RedactorDomChanged` event. Then, they filter those mutations, which are related to them. And pass them to the `didMutated` handler.
    - Block's `didMutated` handled now will be called on every block change (including background changes). Previously, it worked only if Block has been focused.

2. Batching in `onChange`
    - If several changes will be made in short period of time, they will be batched under the single `onChange` call
    - The second argument of `onChange` callback now can be an array of `CustomEvent`. It is a braking change, but not tragic.
    - Debounce have been removed from Block's mutation handler. The batching in `ModificationObserver` used instead.

3. Pressing "Enter" at the end of the Block won't lead to `split()` call, just to appending a new block. Redundant `block-changed` call is removed.

4. The `onChange` won't be triggered with Block Tunes opening and closing.
    - Mutex is used to disable observer on `Fake Cursor` setting/removing.
    - Mutex is implemented through EventBus: `FakeCursorAboutToBeSet` and `FakeCursorHaveBeenSet` is used.
    - Redundant calls of `addFakeCursor` removed. Now we firstly check whether Caret is inside a block or not. Same with `removeFakeCursor`.

5. `EventDispatcher` types improved. 
    - Now we can pass `EventMap` via generic to specify a map of event names and their payloads.
    - All events in common editor Event Bus now have own type declarations.

6. Types for Block mutation events added.

7. Block `didMutated` now catches Tool's element change and updates a link in memory. Fixes #2325


src/components/block/index.ts
```diff
@@ -84,9 +87,11 @@ export enum BlockToolAPI {
 }
 
 /**
- * Names of events supported by Block class
+ * Names of events used in Block
  */
-type BlockEvents = 'didMutated';
+interface BlockEvents extends Record<string, unknown>{
+  'didMutated': Block,
+}
```
gohabereg, line 94: Why not enum then?

src/components/block/index.ts
```diff
@@ -296,18 +227,17 @@ export default class Block extends EventsDispatcher<BlockEvents> {
...
-  }: BlockConstructorOptions) {
+  }: BlockConstructorOptions, eventBus?: EventsDispatcher<EditorEventMap>) {
     super();
 
     this.name = tool.name;
     this.id = id;
     this.settings = tool.settings;
     this.config = tool.settings.config || {};
     this.api = api;
+    this.editorEventBus = eventBus || null;
```
gohabereg, line 238: Maybe we could make api instance of EventDispatcher, so we dont need to pass two parameters. Or maybe have a property of the api object which is EventDispatcher instance

src/components/block/index.ts
```diff
@@ -505,14 +446,21 @@ export default class Block extends EventsDispatcher<BlockEvents> {
...
+    this.holder.classList.toggle(Block.CSS.selected, state);
 
-      SelectionUtils.addFakeCursor(this.holder);
-    } else {
-      this.holder.classList.remove(Block.CSS.selected);
+    const fakeCursorWillBeAdded = state === true && SelectionUtils.isRangeInsideContainer(this.holder);
+    const fakeCursorWillBeRemoved = state === false && SelectionUtils.isElementContainsFakeCursor(this.holder);
+
+    if (fakeCursorWillBeAdded || fakeCursorWillBeRemoved) {
+      this.editorEventBus?.emit(FakeCursorAboutToBeSet, { state }); // mutex
```
gohabereg: Why those constants are not enum?

gohabereg: Also why event is called `AboutToBeSet` even if cursor will be removed

src/components/block/index.ts
```diff
@@ -910,7 +829,138 @@ export default class Block extends EventsDispatcher<BlockEvents> {
       input.removeEventListener('focus', this.handleFocus);
 
       if ($.isNativeInput(input)) {
-        input.removeEventListener('input', this.didMutated);
+        input.removeEventListener('input', this.didMutated as EventListener);
```
gohabereg, line 832: You could try to use function overload for didMutated method to not cast here

src/components/block/index.ts
```diff
@@ -910,7 +829,138 @@ export default class Block extends EventsDispatcher<BlockEvents> {
...
+     * Block API have dispatchChange() method. In this case, mutations list will be undefined.
+     */
+    const isManuallyDispatched = mutationsOrInputEvent === undefined;
+
+    /**
+     * If tool updates its own root element, we need to renew it in our memory
+     */
+    if (!isManuallyDispatched && 'length' in mutationsOrInputEvent) {
+      this.detectToolRootChange(mutationsOrInputEvent);
+    }
```
gohabereg: Why not just check something like `!(mutationsOrInputEvent instanceof InputEvent)`?

src/components/block/index.ts
```diff
@@ -910,7 +829,138 @@ export default class Block extends EventsDispatcher<BlockEvents> {
...
+
+    /**
+     * We won't fire a Block mutation event if mutation contain only nodes marked with 'data-mutation-free' attributes
+     */
+    let shouldFireUpdate;
+
+    if (mutationsOrInputEvent === undefined) {
```
gohabereg:
```suggestion
    if (isManuallyDispatched) {
```

src/components/block/index.ts
```diff
@@ -910,7 +829,138 @@ export default class Block extends EventsDispatcher<BlockEvents> {
...
+
+          return (node as HTMLElement).dataset.mutationFree === 'true';
+        });
+      });
+
+      if (everyRecordIsMutationFree) {
+        shouldFireUpdate = false;
+      } else {
+        shouldFireUpdate = true;
+      }
```
gohabereg:
```suggestion
      shouldFireUpdate = !everyRecordIsMutationFree;
```

src/components/block/index.ts
```diff
@@ -910,7 +829,138 @@ export default class Block extends EventsDispatcher<BlockEvents> {
...
+       */
+      const everyRecordIsMutationFree = mutationsOrInputEvent.length > 0 && mutationsOrInputEvent.every((record) => {
+        const { addedNodes, removedNodes } = record;
+        const changedNodes = [
+          ...Array.from(addedNodes),
+          ...Array.from(removedNodes),
+        ];
+
+        return changedNodes.some((node) => {
+          if ($.isElement(node) === false) {
```
gohabereg:
```suggestion
          if (!$.isElement(node)) {
```

src/components/block/index.ts
```diff
@@ -910,7 +829,138 @@ export default class Block extends EventsDispatcher<BlockEvents> {
...
+     * In case some mutation free elements are added or removed, do not trigger didMutated event
+     */
+    if (!shouldFireUpdate) {
+      return;
+    }
+
+    /**
+     * Drop cache
+     */
+    this.cachedInputs = [];
```
gohabereg: Maybe make a separate method for that in case it is used somewhere else

src/components/block/index.ts
```diff
@@ -910,7 +829,138 @@ export default class Block extends EventsDispatcher<BlockEvents> {
...
+
+    /**
+     * Update current input
+     */
+    this.updateCurrentInput();
+
+    this.call(BlockToolAPI.UPDATED);
+
+    this.emit('didMutated', this);
```
gohabereg, line 919: Magic string

src/components/block/index.ts
```diff
@@ -910,7 +829,138 @@ export default class Block extends EventsDispatcher<BlockEvents> {
...
+      const { mutations } = payload;
+
+      const mutationBelongsToBlock = mutations.some(record => isMutationBelongsToElement(record, this.toolRenderedElement));
+
+      if (mutationBelongsToBlock) {
+        this.didMutated(mutations);
+      }
+    };
+
+    this.editorEventBus?.on(RedactorDomChanged, this.redactorDomChangedCallback);
```
gohabereg, line 941: Also why not enum value?

src/components/block/index.ts
```diff
@@ -910,7 +829,138 @@ export default class Block extends EventsDispatcher<BlockEvents> {
...
+   * We need to detect such changes and update a link to tools main element with the new one
+   *
+   * @param mutations - records of block content mutations
+   */
+  private detectToolRootChange(mutations: MutationRecord[]): void {
+    mutations.forEach(record => {
+      const toolRootHasBeenUpdated = Array.from(record.removedNodes).includes(this.toolRenderedElement);
+
+      if (toolRootHasBeenUpdated) {
+        const newToolElement = record.addedNodes[record.addedNodes.length - 1];
```
gohabereg, line 962: Would it always be the last added node?

src/components/block/index.ts
```diff
@@ -910,7 +829,138 @@ export default class Block extends EventsDispatcher<BlockEvents> {
...
+
+        console.warn('Original element changed', newToolElement);
```
gohabereg: Is it necessary here?

src/components/events/index.ts
```diff
@@ -0,0 +1,26 @@
...
+  RedactorDomChanged,
+  BlockChanged,
+  FakeCursorAboutToBeSet,
+  FakeCursorHaveBeenSet
+};
+
+/**
+ * Event name -> Event payload
+ */
+export interface EditorEventMap extends Record<string, unknown> {
```
gohabereg: Do you really need extends here? It is not [type-safe](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBASwHY2FAZgQwMbDgUQBMEZp8A3YFAWUzDlFSUIGc4AlYbaQgHhZhRkAcwA0cAK5IA1kggB3JAD44AbwBQcOACEANhGzSAEhEpRghAFxwA5HoPHTaCwAVMAT32ZCNgNyaOCxxSKAARCABbAGEAC0wkYQtrG05CYOhw6LiE1w8vH38te0NY+MSrW2LpUpzCN08Ibz8AgDFMaWAoiSgWaABBACMICRgAFQhtYABlYBhkto6unv6hkfHJmZh6-OatBc7u3qgjTEpJqk359oPl49Pgc6RN7caC9QBfdXVuJAE4TGsRBIZEoNDocAAvGoAkV9IYTGYknAAERVBHOOp5V7I0QwwJpbAhTI1crWZGpdJhSIk3INbw4vFVGkVVFw6rZcovem4rR7a5LI6DYZjCbTWZk-YClbC9ZirZY7l4yWHaAnM7AC7ilHK25qh4ap6zLmEZEffzqTAAbRsvQieCg8UIkTgYCgEDANgAur44AB6X1wXgAWhDcDkDCgbqgQA)

src/components/events/index.ts
```diff
@@ -0,0 +1,26 @@
...
+  [BlockHovered]: BlockHoveredPayload,
+  [RedactorDomChanged]: RedactorDomChangedPayload,
+  [BlockChanged]: BlockChangedPayload,
+  [FakeCursorAboutToBeSet]: FakeCursorAboutToBeSetPayload,
+  [FakeCursorHaveBeenSet]: FakeCursorHaveBeenSetPayload,
```
gohabereg: It is an interface

```suggestion
  [BlockHovered]: BlockHoveredPayload;
  [RedactorDomChanged]: RedactorDomChangedPayload;
  [BlockChanged]: BlockChangedPayload;
  [FakeCursorAboutToBeSet]: FakeCursorAboutToBeSetPayload;
  [FakeCursorHaveBeenSet]: FakeCursorHaveBeenSetPayload;
```

src/components/modules/blockManager.ts
```diff
@@ -851,17 +848,19 @@ export default class BlockManager extends Module {
...
-   * @param details - additional data to pass with change event
+   * @param detailData - additional data to pass with change event
    */
-  private blockDidMutated(mutationType: BlockMutationType, block: Block, details: Record<string, unknown> = {}): Block {
+  private blockDidMutated<Type extends BlockMutationType>(mutationType: Type, block: Block, detailData: Omit<BlockMutationEventMap[Type]['detail'], 'target'>): Block {
     const event = new CustomEvent(mutationType, {
       detail: {
         target: new BlockAPI(block),
-        ...details,
+        ...detailData as Omit<BlockMutationEventMap[Type]['detail'], 'target'>,
```
gohabereg: Looks like you could use type alias for that

src/components/modules/modificationsObserver.ts
```diff
@@ -10,17 +14,82 @@ export default class ModificationsObserver extends Module {
...
+   * Timeout used to batched several events in a single onChange call
+   */
+  private batchingTimeout: null | ReturnType<typeof setTimeout> = null;
+
+  /**
+   * Array of onChange events used to batch them
+   *
+   * Map is used to filter duplicated events related to the same block
+   */
+  private batchingOnChangeQueue: Map<`block:${BlockId}:event:${BlockMutationType}`, BlockMutationEvent> = new Map();
```
gohabereg: I'd use a type alias for that as well. Looks like magic const imho

gohabereg: Actually, much better like this:

```suggestion
  private batchingOnChangeQueue = new Map<`block:${BlockId}:event:${BlockMutationType}`, BlockMutationEvent>();
```

src/components/modules/modificationsObserver.ts
```diff
@@ -29,11 +98,45 @@ export default class ModificationsObserver extends Module {
...
    * @param event - some of our custom change events
    */
-  public onChange(event: CustomEvent): void {
+  private particularBlockChanged(event: BlockMutationEvent): void {
     if (this.disabled || !_.isFunction(this.config.onChange)) {
       return;
     }
 
-    this.config.onChange(this.Editor.API.methods, event);
+    this.batchingOnChangeQueue.set(`block:${event.detail.target.id}:event:${event.type as BlockMutationType}`, event);
```
gohabereg, line 111: Why would you need to cast here, isn't BlockMutationEvent has `BlockMutationType` as type for `type` prop?

src/components/selection.ts
```diff
@@ -353,24 +353,47 @@ export default class SelectionUtils {
...
+  public static isRangeInsideContainer(container: HTMLElement): boolean {
     const range = SelectionUtils.range;
-    const fakeCursor = $.make('span', 'codex-editor__fake-cursor');
 
-    fakeCursor.dataset.mutationFree = 'true';
+    if (range === null) {
+      return false;
+    }
+
+    return container.contains(range.startContainer);
```
gohabereg, line 367: Shouldn't you also check `endContainer`?

src/components/selection.ts
```diff
@@ -353,24 +353,47 @@ export default class SelectionUtils {
...
     range.collapse();
     range.insertNode(fakeCursor);
   }
 
+  /**
+   * Check if passed element contains a fake cursor
+   *
+   * @param el - where to check
+   */
+  public static isElementContainsFakeCursor(el: HTMLElement): boolean {
```
gohabereg: Not sure it's better but grammatically correct

```suggestion
  public static doesElementContainFakeCursor(el: HTMLElement): boolean {
```

src/components/ui/toolbox.ts
```diff
@@ -33,6 +33,19 @@ export enum ToolboxEvent {
   BlockAdded = 'toolbox-block-added',
 }
 
+/**
+ * Events fired by the Toolbox
+ *
+ * Event name -> payload
+ */
+export interface ToolboxEventMap extends Record<string, unknown> {
```
gohabereg: extends here is also not type safe

src/components/utils/events.ts
```diff
@@ -1,31 +1,34 @@
 import { isEmpty } from '../utils';
 
 /**
- * @class EventDispatcher
+ * Mapped type with subscriptions list
  *
- * Has two important methods:
+ * event name -> array of callbacks
+ */
+type Subscriptions<EventMap> = {
```
gohabereg, line 13:
```suggestion
type Subscriptions<EventMap extends Record<string, any>> = {
```

src/components/utils/events.ts
```diff
@@ -1,31 +1,34 @@
...
+  [name in keyof EventMap]: Array<(data?: EventMap[name]) => unknown>
```
gohabereg:
```suggestion
  [Key in keyof EventMap]: Array<(data?: EventMap[Key]) => void>
```

gohabereg: Also you could make a type alias for `Listener`:

```
type Listener<Data extends any> = (data?: Data) => void;
```

So you could use it here like `Listener<EventMap[Key]>[]`

src/components/utils/events.ts
```diff
@@ -1,31 +1,34 @@
...
+ * Provides methods for working with Event Bus:
  *    - {Function} on - appends subscriber to the event. If event doesn't exist - creates new one
  *    - {Function} emit - fires all subscribers with data
  *    - {Function off - unsubscribes callback
- * @version 1.0.0
- * @typedef {Events} Events
- * @property {object} subscribers - all subscribers grouped by event name
  */
-export default class EventsDispatcher<Events extends string = string> {
+export default class EventsDispatcher<EventMap extends Record<string, unknown> = undefined> {
```
gohabereg: Why undefined as default value?

src/components/utils/events.ts
```diff
@@ -1,31 +1,34 @@
...
   /**
+   * All subscribers grouped by event name
    * Object with events` names as key and array of callback functions as value
-   *
-   * @type {{}}
    */
-  private subscribers: {[name: string]: Array<(data?: object) => unknown>} = {};
+  private subscribers: Subscriptions<EventMap> = <Subscriptions<EventMap>>{};
```
gohabereg: Please don't use this cast syntax, use `as`

src/components/utils/events.ts
```diff
@@ -1,31 +1,34 @@
...
   /**
    * Subscribe any event on callback
    *
-   * @param {string} eventName - event name
-   * @param {Function} callback - subscriber
+   * @param eventName - event name
+   * @param callback - subscriber
    */
-  public on(eventName: Events, callback: (data: object) => unknown): void {
+  public on<Name extends keyof EventMap>(eventName: Name, callback: (data: EventMap[Name]) => unknown): void {
```
gohabereg: `void` type is better to indicate *you don't expect any return value*

```suggestion
  public on<Name extends keyof EventMap>(eventName: Name, callback: (data: EventMap[Name]) => void): void {
```

src/components/utils/events.ts
```diff
@@ -37,15 +40,15 @@ export default class EventsDispatcher<Events extends string = string> {
   /**
    * Subscribe any event on callback. Callback will be called once and be removed from subscribers array after call.
    *
-   * @param {string} eventName - event name
-   * @param {Function} callback - subscriber
+   * @param eventName - event name
+   * @param callback - subscriber
    */
-  public once(eventName: Events, callback: (data: object) => unknown): void {
+  public once<Name extends keyof EventMap>(eventName: Name, callback: (data: EventMap[Name]) => unknown): void {
```
gohabereg:
```suggestion
  public once<Name extends keyof EventMap>(eventName: Name, callback: (data: EventMap[Name]) => void): void {
```

gohabereg: Also could make use of Listener alias type here and above

src/components/utils/events.ts
```diff
@@ -64,10 +67,10 @@ export default class EventsDispatcher<Events extends string = string> {
   /**
    * Emit callbacks with passed data
    *
-   * @param {string} eventName - event name
-   * @param {object} data - subscribers get this data when they were fired
+   * @param eventName - event name
+   * @param data - subscribers get this data when they were fired
    */
-  public emit(eventName: Events, data?: object): void {
+  public emit<Name extends keyof EventMap>(eventName: Name, data?: EventMap[Name]): void {
```
gohabereg, line 78: What is the point of `EventMap` if `data` is optional?

src/components/utils/events.ts
```diff
@@ -37,15 +40,15 @@ export default class EventsDispatcher<Events extends string = string> {
...
+   * @param callback - subscriber
    */
-  public once(eventName: Events, callback: (data: object) => unknown): void {
+  public once<Name extends keyof EventMap>(eventName: Name, callback: (data: EventMap[Name]) => unknown): void {
     if (!(eventName in this.subscribers)) {
       this.subscribers[eventName] = [];
     }
 
-    const wrappedCallback = (data: object): unknown => {
+    const wrappedCallback = (data: EventMap[typeof eventName]): unknown => {
```
gohabereg:
```suggestion
    const wrappedCallback = (data: EventMap[typeof eventName]): void => {
```

src/components/utils/events.ts
```diff
@@ -82,10 +85,10 @@ export default class EventsDispatcher<Events extends string = string> {
   /**
    * Unsubscribe callback from event
    *
-   * @param {string} eventName - event name
-   * @param {Function} callback - event handler
+   * @param eventName - event name
+   * @param callback - event handler
    */
-  public off(eventName: Events, callback: (data: object) => unknown): void {
+  public off<Name extends keyof EventMap>(eventName: Name, callback: (data: EventMap[Name]) => unknown): void {
```
gohabereg:
```suggestion
  public off<Name extends keyof EventMap>(eventName: Name, callback: Listener<EventMap[Name]>): void {
```

src/components/utils/popover/index.ts
```diff
@@ -64,11 +64,18 @@ export enum PopoverEvent {
   Close = 'close'
 }
 
+/**
+ * Events fired by the Popover
+ */
+interface PopoverEventMap extends Record<string, unknown> {
```
gohabereg: Not type safe as well

types/events/block/index.ts
```diff
@@ -0,0 +1,44 @@
...
+import { BlockRemovedEvent, BlockRemovedMutationType } from './BlockRemoved';
+
+/**
+ * Map for Custom Events related to block mutation types
+ */
+export interface BlockMutationEventMap {
+  /**
+   * New Block added
+   */
+  [BlockAddedMutationType]: BlockAddedEvent,
```
gohabereg: Semicolon for interfaces. Something might be wrong with your IDE setup

gohabereg (changes requested):

src/components/block/index.ts
neSpecc: do you want to expose editor internal event bus to users? I think, it's better to do design it first and implement via a separate feature

src/components/block/index.ts
neSpecc: renamed to `FakeCursorAboutToBeToggled`. 

Enum is not suitable here, because all the information about the single event stored in isolated file. There is no need to describe a enum with all available event names — EventMap used instead

image

src/components/block/index.ts
neSpecc: Technically I can do that, but in that case we'll need to create an `HTMLInputElement` extension with overloading of `removeEventListener` and `addEventListener` methods telling that in our custom `BlockInputHTMLElement` those handlers can accept not only `InputEvent` but `MutationRecord[]` as well. 

```ts
interface BlockInputHTMLElement extends HTMLInputElement {
  addEventListener(
    type: 'input',
    listener: (ev: InputEvent) => any,
    options?: boolean | AddEventListenerOptions
  ): void;
  addEventListener(
    type: 'input',
    listener: (ev: InputEvent | MutationRecord[]) => any,
    options?: boolean | AddEventListenerOptions
  ): void;
}
```

It has no sense because we're not creating a custom `HTMLInputElement` implementation and its handler will never get `MutationRecord[]`. We just use the same function for both cases: "input" event and mutation observer callback. So using a casting in our case we just telling a compiler that in this line it should treat handler as a regular "input" event listener because that is what it actually is.

neSpecc: and also we'll need to cast our input to `BlockInputHTMLElement` anyway

src/components/block/index.ts
neSpecc: It's not a new code, it was existed before my change. I've added a comment to that statement

src/components/block/index.ts
neSpecc: There is usually just one removed and added node when Tool is replacing its wrapper element, so I'm taking the last item which will work fine for most cases.

If Tool will replace its wrapper with several nodes, it won't work anyway because it's this behaviour is not supported

src/components/modules/modificationsObserver.ts
neSpecc: no, since `event` is a `CustomEvent` here where `type` is a `string`. I've [tried](https://github.com/codex-team/editor.js/pull/2349/commits/ce6299e72c66a4ca0d6f4255c74dba17b9b05461) to implement CustomEvent extensions, but it was much more over-engineered way with breaking API change (onChange handler accepts not `CustomEvent`, but custom classes), so I've [got rid](https://github.com/codex-team/editor.js/pull/2349/commits/f298a088907ae2a2a3628cb3259ad1ce35235e09) of that logic. 

So this cast is telling a compiler that `event.type` here is a known event name.

image

src/components/selection.ts
neSpecc: I think it's enough to know that a container has a range starting node. And this statement was used previously inside `addFakeCursor`, I've just moved it to a separate method

src/components/selection.ts
neSpecc: renamed with `isFakeCursorInsideContainer` to make it consistent with `isRangeInsideContainer`

src/components/utils/events.ts
neSpecc: I don't see a problem here. EventMap can explicitly describe events without a payload: 

```ts
/**
 * Events fired by the Toolbox
 *
 * Event name -> payload
 */
export interface ToolboxEventMap {
  [ToolboxEvent.Opened]: undefined;
  [ToolboxEvent.Closed]: undefined;
  [ToolboxEvent.BlockAdded]: {
    block: BlockAPI
  };
}
```

src/components/block/index.ts
neSpecc: done

src/components/block/index.ts
neSpecc: done

src/components/block/index.ts
neSpecc: done

src/components/block/index.ts
neSpecc: done

src/components/block/index.ts
neSpecc: done

src/components/block/index.ts
neSpecc: answered

src/components/block/index.ts
neSpecc: removed

src/components/events/index.ts
neSpecc: removed

src/components/events/index.ts
neSpecc: fixed

src/components/ui/toolbox.ts
neSpecc: removed

ilyamore88 (approved):
I don't have questions from my side. But please, make smaller PRs next time🥺

src/components/utils/events.ts
gohabereg: This seems to be still relevant

gohabereg (approved):

src/components/utils/events.ts
neSpecc: why do we need it?

neSpecc: it will lead returning extension of `Record<string, any>` in all event maps

gohabereg: To specify EventMap is an object

PR #2364, opened 2023-05-26
Change calculating toolbar right orientation

### LINKED ISSUE #2268: Toolbar gets cut off
When highlighting a text that's on the right end of the editor, the toolbar gets cut off:

image

Do you have any pointer where to fix this?

yasanov:
use contentRect.width when calculating toolbar orientation. Fix #2268

newCoords are calculated relative to wrapper 
image

so we need to compare realRightCoord with wrapper width not it's right coordinate
image


neSpecc:
Seems like it was fixed in 2.29

PR #2396, opened 2023-06-25
set import type when 'preserveValueImports' and 'isolatedModules' are both enabled.

mphouse16:
When 'preserveValueImports' and 'isolatedModules' are both enabled, compile error occured.
So, I updated these.


neSpecc (approved):

neSpecc:
Add a changelog please

types/events/block/index.ts
```diff
@@ -1,7 +1,11 @@
-import { BlockAddedEvent, BlockAddedMutationType } from './BlockAdded';
-import { BlockChangedEvent, BlockChangedMutationType } from './BlockChanged';
-import { BlockMovedEvent, BlockMovedMutationType } from './BlockMoved';
-import { BlockRemovedEvent, BlockRemovedMutationType } from './BlockRemoved';
+import { BlockAddedMutationType } from './BlockAdded';
+import type { BlockAddedEvent, BlockAddedMutationType } from './BlockAdded';
```
neSpecc, line 2: There is a TS error:
```
Duplicate identifier 'BlockAddedMutationType'.ts(2300)
```

neSpecc (changes requested):

types/events/block/index.ts
mphouse16: Really Sorry about that. You should remove 'BlockAddedMutationType' from type import.
like this.
from - import type { BlockAddedEvent, BlockAddedMutationType } from './BlockAdded';
to - import type { BlockAddedEvent } from './BlockAdded';

mphouse16:
I apologize for our recent misbehavior on github.com. Since I don't have access to a local PC these days, I've been working directly on github.com, but I'm not used to issuing commands on the web, so I think I made a lot of mistakes. Please check the above bug with the link below.

https://github.com/codex-team/editor.js/pull/2396/commits/b09af9a8c87b2cd29c15d94dcaeb3114b73c8238

PR #2402, opened 2023-07-05
feat(block-events): delete keydown handling added

neSpecc:
Currently, we have only Backspace handler that (if caret at the start) merges current block with **previous**.

This PR adds similar logic for the Delete keydown. If pressed at the end of the block, the block will be merged with the **next** block.

Also, code of both handlers was refactored and simplified. Tests added.


test/cypress/tests/modules/BlockEvents.cy.ts
```diff
@@ -1,5 +1,25 @@
+import type EditorJS from '../../../../types/index';
```
gohabereg: Maybe move some helper functions and mocks to a separate file and apply refactoring where possible (extract common fragments). Too hard too read atm

gohabereg (commented):
Fix CI please

test/cypress/tests/modules/BlockEvents.cy.ts
neSpecc: I decided to leave small mock classes inside tests for better understanding of test implementation. If I'll move them to externall files, they won't be such self-descriptive.

I can separate this file to several files to make them smaller

src/components/modules/blockEvents.ts
```diff
@@ -277,104 +282,166 @@ export default class BlockEvents extends Module {
...
 
       return;
     }
 
     /**
-     * Don't handle Backspaces when Tool sets enableLineBreaks to true.
-     * Uses for Tools like <code> where line breaks should be handled by default behaviour.
-     *
-     * But if caret is at start of the block, we allow to remove it by backspaces
+     * If current Block is empty, just just remove it and set cursor to the previous Block (like we're removing line break char)
```
ilyamore88:
```suggestion
     * If current Block is empty, just remove it and set cursor to the previous Block (like we're removing line break char)
```

src/components/modules/blockEvents.ts
```diff
@@ -277,104 +282,166 @@ export default class BlockEvents extends Module {
...
+     * If next Block is empty, it should be removed just like a character
+     */
+    if (nextBlock.isEmpty) {
+      BlockManager.removeBlock(nextBlock);
+
+      return;
+    }
+
+    /**
+     * If current Block is empty, just just remove it and set cursor to the next Block (like we're removing line break char)
```
ilyamore88:
```suggestion
     * If current Block is empty, just remove it and set cursor to the next Block (like we're removing line break char)
```

src/components/modules/blockEvents.ts
```diff
@@ -277,104 +282,166 @@ export default class BlockEvents extends Module {
...
+      Caret.setToBlock(BlockManager.currentBlock, Caret.positions.START);
+
+      return;
+    }
+
+    const bothBlocksMergeable = currentBlock.mergeable && nextBlock.mergeable;
+
+    /**
+     * If Blocks could be merged, do it
+     * Otherwise, just navigate previous block
```
ilyamore88:
```suggestion
     * Otherwise, just navigate to the next block
```

src/components/modules/blockEvents.ts
```diff
@@ -277,104 +282,166 @@ export default class BlockEvents extends Module {
...
+    } else {
+      Caret.setToBlock(
+        BlockManager.nextBlock,
+        Caret.positions.START
+      );
+    }
+  }
+
+  /**
+   * Merge current and previous Blocks if they have the same type
```
ilyamore88:
```suggestion
   * Merge two Blocks if they have the same type
```

src/components/modules/blockEvents.ts
```diff
@@ -277,104 +282,166 @@ export default class BlockEvents extends Module {
...
+   *
+   * @param blockToMerge - what Block we want to merge
+   * @param targetBlock - to which Block we want to merge
+   */
+  private mergeBlocks(blockToMerge: Block, targetBlock: Block): void {
```
ilyamore88: I think it's more clear to have a first argument as a target, and a second argument as data to merge it to the target block (as in the BlockManager.mergeBlocks)

```suggestion
  private mergeBlocks(targetBlock: Block, blockToMerge: Block): void {
```

test/cypress/tests/modules/BlockEvents/Backspace.cy.ts
```diff
@@ -0,0 +1,363 @@
...
+
+        container.append(input, input2);
+
+        return container;
+      }
+
+      /**
+       * Saving logic is not necessary for this test
+       */
+      // eslint-disable-next-line @typescript-eslint/no-empty-function
```
ilyamore88, line 85: FYI: you can add a comment to ignore a rule like this

```suggestion
      // eslint-disable-next-line @typescript-eslint/no-empty-function -- saving logic is not necessary for this test
```

gohabereg (approved):

test/cypress/tests/modules/BlockEvents/Backspace.cy.ts
```diff
@@ -0,0 +1,380 @@
...
+
+    /**
+     * Toolbox has been closed
+     */
+    cy.get('[data-cy=editorjs]')
+      .find('.ce-toolbar')
+      .should('not.have.class', 'ce-toolbar--opened');
+  });
+
+  it('should merge current Block with the previous one if Caret at the start of the Block and both Blocks are mergeable. Also, should close the Toolbox. Caret should be places in a place of glue', function () {
```
ilyamore88, line 228: Just thinking about cases when there are two mergeable blocks with different type. Which type you'll leave for backspace and for delete events? Maybe add test cases?

neSpecc: There will be a target block type. `mergeBlock` just rerenders it with new data. 

It should be covered by Block Manager unit test, but currently we don't have unit tests. I think, it would be covered in Document Model.

ilyamore88 (approved):

test/cypress/tests/modules/BlockEvents/Backspace.cy.ts
neSpecc: I was wrong, currently we can merge only similar types of blocks. I thought that we're using our `ConversionConfig` feature for merging, but seems like it has not been implemented yet. 

I'll add a statement for `type` checking for now. Later I'll implement Conversion, and remove the `type` restriction. And we'll have an ability to merge any blocks.

PR #2419, opened 2023-07-17
feat(shortcuts): convert block by tools shortcut

neSpecc:
If Tool has a `conversionConfig` and a `shortcut`, it will be converted by shortcut press instead of appending the new block below.

- Conversion logic moved from Conversion Toolbar to the Block Manager and some utilities.
- The new `blocks.convert(id, type)` API method added. It is used by a Toolbox module in shortcuts.
- Shortcuts displayed in the Conversion Toolbar

Requirements for conversion:
- Original block has `conversionConvig.export` implemented
- Target block has `conversionConvig.import` implemented

... in all other cases shortcuts will append a new block below

image

### DIFF
--- a/test/cypress/tests/api/blocks.cy.ts
+++ b/test/cypress/tests/api/blocks.cy.ts
@@ -157,7 +157,7 @@ describe('api.blocks', () => {
     });
   });
 
-  describe.only('.convert()', function () {
+  describe('.convert()', function () {
     it('should convert a Block to another type if original Tool has "conversionConfig.export" and target Tool has "conversionConfig.import"', function () {
       /**
        * Mock of Tool with conversionConfig
--- a/test/cypress/tests/api/blocks.cy.ts
+++ b/test/cypress/tests/api/blocks.cy.ts
@@ -241,13 +241,9 @@ describe('api.blocks', () => {
           const fakeId = 'WRNG_ID';
           const { convert } = editor.blocks;
 
-          try {
-            convert(fakeId, 'convertableTool');
+          const exec = (): void => convert(fakeId, 'convertableTool');
 
-            expect(true).to.be.false; // we should not pass here
-          } catch (error) {
-            expect(error.message).to.eq(`Block with id "${fakeId}" not found`);
-          }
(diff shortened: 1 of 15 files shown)


test/cypress/tests/api/blocks.cy.ts
```diff
@@ -141,4 +156,179 @@ describe('api.blocks', () => {
...
+          const fakeId = 'WRNG_ID';
+          const { convert } = editor.blocks;
+
+          try {
+            convert(fakeId, 'convertableTool');
+
+            expect(true).to.be.false; // we should not pass here
+          } catch (error) {
+            expect(error.message).to.eq(`Block with id "${fakeId}" not found`);
+          }
```
ilyamore88: Why not use `.to.throw()`?

ilyamore88:
As I understand, now I can't create a new block with shortcuts if the selected block might be converted. What if we will change converting shortcuts? For example, add the Option (Alt) button to the shortcut.

test/cypress/tests/api/blocks.cy.ts
neSpecc: Tried, but it is not working. Error fails Cypress test process.

neSpecc:
> As I understand, now I can't create a new block with shortcuts if the selected block might be converted. 

You can create a new block by pressing shortcut in an empty paragraph. It will be replaced by a new block, just like in the current version.

> What if we will change converting shortcuts? For example, add the Option (Alt) button to the shortcut.

For now, it's up to user to specify any shortcut (including `ALT+...`), we can't change it so simply. And it won't be obvious that we changed it. 

I think it is not a problem, because you still can make a new block below using "Enter" + "shortcut". For example, if you want to create a header below a non-empty paragraph, you can focus the Paragraph, press Enter, then press a shortcut. It's a more intuitive solution

test/cypress/tests/api/blocks.cy.ts
gohabereg: You need to wrap original function with another one:

```
const exec = () => functionThroesError();

expect(exec).to.throw...
```

gohabereg (approved):

ilyamore88 (approved):

PR #2443, opened 2023-08-17
chore(api): blocks.update(id, data) method improved

neSpecc:
1. `blocks.update(id, data)` previously triggers onChange with `block-removed`, `block-added`, `block-changed` events. Now it will trigger only `block-changed` 
2. `blocks.update(id, data)` now can accept partial data to update. It will be merged with existing data
3. `blocks.update(id, data)` now will return `Promise<BlockAPI>`
4. Custom Chai assertion `be.calledWithBatchedEvents` implemented
5. `BlockMutationType` and `BlockMutationEvent` types exported

### DIFF
--- a/src/components/modules/api/blocks.ts
+++ b/src/components/modules/api/blocks.ts
@@ -297,26 +297,19 @@ export default class BlocksAPI extends Module {
    * @param id - id of the block to update
    * @param data - the new data
    */
-  public update = (id: string, data: BlockToolData): void => {
+  public update = async (id: string, data: Partial<BlockToolData>): Promise<BlockAPIInterface> => {
     const { BlockManager } = this.Editor;
     const block = BlockManager.getBlockById(id);
 
-    if (!block) {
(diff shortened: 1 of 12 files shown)


src/components/modules/api/blocks.ts
```diff
@@ -297,26 +297,19 @@ export default class BlocksAPI extends Module {
...
-      id: block.id,
-      tool: block.name,
-      data,
-      index: blockIndex,
-      replace: true,
-      tunes: block.tunes,
-    });
+    // we cast to any because our BlockAPI has no "new" signature
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    return new (BlockAPI as any)(updatedBlock);
```
gohabereg, line 312: Why not just `BlockAPI(updateBlock)` then? It is intentionally a function, not a class

neSpecc: because it does not return anything

gohabereg (approved):

ilyamore88 (approved):

PR #2454, opened 2023-08-24
fix(renderer): handle empty array as data.blocks

### LINKED ISSUE #2084: [Bug] editor.render with "blocks":[] breaks the editor
(Not sure if exact duplicate of https://github.com/codex-team/editor.js/pull/1741 )

Steps to reproduce:
1. call editor.render with this json: {"time":1655483036373,"blocks":[],"version":"2.24.3"}
2. Observe editor gives errors in console, and it is now impossible to add blocks or type into the editor.

Expected behavior:
Editor operates as normal, with either no blocks present or with one default block.

Workaround:
If you make a blank document and save it, it returns "blocks":[]. However you cannot load this to get back your blank document.
I've had to manually change blank documents to have instead: "blocks":[{"id":"dvSFGiwGIy","type":"paragraph","data":{"text":""}}]
With at least one block, the file loads correctly.

Screenshots:
I get three different errors when clicking around on the broken editor:

`Editor.js 2.24.3 There is no block at index -1 8 editor.js:1:3853`
`Uncaught TypeError: this.Editor.BlockManager.lastBlock is undefined`
`Uncaught TypeError: s is undefined
    value editor.js:1
    value editor.js:1
    value editor.js:1
    on editor.js:1
    value editor.js:1
    value editor.js:1
    t editor.js:1
    l editor.js:1
    _invoke editor.js:1
    e editor.js:1
    n editor.js:1
    s editor.js:1
    exports editor.js:1
    exports editor.js:1
    value editor.js:1
    t editor.js:1
    l editor.js:1
    _invoke editor.js:1
    e editor.js:1
    n editor.js:1
    s editor.js:1
    exports editor.js:1
    exports editor.js:1
    value editor.js:1
    t editor.js:1
    l editor.js:1
    _invoke editor.js:1
    e editor.js:1
    n editor.js:1
    s editor.js:1
    exports editor.js:1
    exports editor.js:1
    promise callback*t/</< editor.js:1
    t editor.js:1
    l editor.js:1
    _invoke editor.js:1
    e editor.js:1
    n editor.js:1
    s editor.js:1
    exports editor.js:1
    exports editor.js:1
    value editor.js:1
    t editor.js:1
    l editor.js:1
    _invoke editor.js:1
    e editor.js:1
    n editor.js:1
    s editor.js:1
    promise callback*n editor.js:1
    s editor.js:1
    promise callback*n editor.js:1
    s editor.js:1
    exports editor.js:1
    exports editor.js:1
    promise callback*t editor.js:1
    t editor.js:1
    <anonymous> book.js:1
    <anonymous> book.js:1
    <anonymous> book.js:1`
Device, Browser, OS:

Device, Browser, OS:
Desktop/Firefox/Windows 10

Editor.js version: 2.24.3

paragraph@2.8.0
header@2.6.2
simple-image@1.4.1
delimiter@1.2.0

neSpecc:
Currently, if we pass `[]` as initial data or via `blocks.render()` editor will contain no-blocks at all. It will break the UX since `redactorClicked` will throw error:

```
Can not read property "holder" of undefined
```

https://github.com/codex-team/editor.js/blob/f8e94142177a9a49d3bebb4e3e28628ac21a7e0d/src/components/modules/ui.ts#L726-L727

Actually `.getBlockByIndex()` could return `undefined` (if we pass an unexisting index) — so I've added `undefined` as the return type. But in case of getting the last block by `-1` it always should return a Block since it is expected that we always have at least one Block. So I've added overload for that case instead of checking on `undefined` in `redactorClicked`

Resolves #2084

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### 2.29.0
+
+- `Fix` — Passing an empty array via initial data or `blocks.render()` won't break the editor
+
 ### 2.28.0
 
 - `New` - Block ids now displayed in DOM via a data-id attribute. Could be useful for plugins that want access a Block's element by id.
--- a/src/components/blocks.ts
+++ b/src/components/blocks.ts
@@ -323,7 +323,7 @@ export default class Blocks {
    * @param {number} index — Block index
    * @returns {Block}
    */
-  public get(index: number): Block {
+  public get(index: number): Block | undefined {
     return this.blocks[index];
   }
 
--- a/src/components/modules/blockManager.ts
+++ b/src/components/modules/blockManager.ts
@@ -590,10 +590,25 @@ export default class BlockManager extends Module {
   /**
    * Returns Block by passed index
    *
+   * If we pass -1 as index, the last block will be returned
+   * There shouldn't be a case when there is no blocks at all — at least one always should exist
+   */
+  public getBlockByIndex(index: -1): Block;
+
+  /**
+   * Returns Block by passed index.
+   *
+   * Could return undefined if there is no block with such index
+   */
+  public getBlockByIndex(index: number): Block | undefined;
+
+  /**
+   * Returns Block by passed index
+   *
    * @param {number} index - index to get. -1 to get last
    * @returns {Block}
    */
-  public getBlockByIndex(index): Block {
+  public getBlockByIndex(index: number): Block | undefined {
     if (index === -1) {
       index = this._blocks.length - 1;
     }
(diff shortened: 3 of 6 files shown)


docs/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+### 2.29.0
```
ilyamore88, line 3: Don't you want to publish it as a patch version?

neSpecc: maybe. But I don't want to include it in 2.28.0 since it has already been tested via RC.

neSpecc: It's not so critical since it has workaround

ilyamore88 (approved):

gohabereg (approved):

docs/CHANGELOG.md
MussaratAziz: what's the workaround? @neSpecc

PR #2465, opened 2023-09-02
Optimize data-mutation-free for mutation observer

medzhidov:
Optimization from this issue:
https://github.com/codex-team/editor.js/issues/2205

### DIFF
--- a/src/components/block/index.ts
+++ b/src/components/block/index.ts
@@ -911,7 +911,7 @@ export default class Block extends EventsDispatcher<BlockEvents> {
             return false;
           }
 
-          return (node as HTMLElement).dataset.mutationFree === 'true';
+          return !!(node as HTMLElement).closest('[data-mutation-free="true"]');
         });
       });
 


neSpecc (commented):
Please, implement a test for this case

bettysteger:
@medzhidov can you write a test for this case? The test could be similar to [this one](https://github.com/codex-team/editor.js/pull/2317/files#diff-0229d9e7d89b5f70373d7d562dbd380858019e0c47442229ae9215b7987cdcddR329).

bettysteger:
@medzhidov so this will also resolve PR https://github.com/codex-team/editor.js/pull/2317 ?

medzhidov:
@bettysteger Yes, but the solution suggested in #2317 is just another approach. I propose making the data mutation more user-friendly, while in #2317, it is suggested to create a new property for the block instance that will disable any mutations inside it. I believe both solutions are good, but the second one (#2317) is more fundamental because we can manipulate the block from its class instance. "data-mutation-free" is more versatile; it may be needed when we want to "hide" certain parts of our block from the onchange callback, rather than the entire block.

bettysteger:
@medzhidov thanks for your answer! much appreciated!

I just did the "same" PR, but added a test: https://github.com/codex-team/editor.js/pull/2548

neSpecc:
Resolved by #2548

PR #2477, opened 2023-09-13
fix: inline toolbar incorrectly calculates orientation

### LINKED ISSUE #2268: Toolbar gets cut off

### LINKED ISSUE #2476: InlineToolbar incorrectly overflows on right edge
When a parent of the editor is using position: fixed or absolute, the inline toolbar will overflow on the right side.

Expected behaviour:
Inline toolbar should position itself to avoid right-side overflow if there is space.

Screenshots:
image

Device, Browser, OS: Desktop, Chrome, Windows 10

Editor.js version: 2.28.0

SneakiBrina:
### Summary

When calculating the inline toolbar orientation, the selection coordinates are being adjusted for the offset of the wrapper element however, the content coordinates are not. 

https://github.com/codex-team/editor.js/blob/77eb320203e5a25e9b8699de6fe0a0cc285da73e/src/components/modules/toolbar/inline.ts#L164-L167

https://github.com/codex-team/editor.js/blob/77eb320203e5a25e9b8699de6fe0a0cc285da73e/src/components/modules/toolbar/inline.ts#L186

https://github.com/codex-team/editor.js/blob/77eb320203e5a25e9b8699de6fe0a0cc285da73e/src/components/modules/toolbar/inline.ts#L199

### Changes

- take offset into account for comparison against content rects
- added regression test

Resolves: #2476
Resolves: #2268

fixed-editorjs

### DIFF
--- a/src/components/modules/toolbar/inline.ts
+++ b/src/components/modules/toolbar/inline.ts
@@ -191,12 +191,12 @@ export default class InlineToolbar extends Module<InlineToolbarNodes> {
      */
     this.nodes.wrapper.classList.toggle(
       this.CSS.inlineToolbarLeftOriented,
-      realLeftCoord < this.Editor.UI.contentRect.left
+      realLeftCoord < this.Editor.UI.contentRect.left - wrapperOffset.x
     );
 
     this.nodes.wrapper.classList.toggle(
       this.CSS.inlineToolbarRightOriented,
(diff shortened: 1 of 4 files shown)


neSpecc:
Probably fixed via #2550. Please, test it in 2.29.0-rc.8

PR #2514, opened 2023-10-17
fix(onchange): blocks.render would not call onChange in safari

neSpecc:
This PR actually pulls the latest version of "Paragraph" tool with the fix of unwanted `onChange` in `blocks.render()`

See https://github.com/editor-js/paragraph/pull/76 for details


ilyamore88 (approved):

gohabereg (approved):

collimarco:
The link to the related issue seems wrong... Can you provide a link to the original issue that is solved by this PR?

My guess is that the issue could affect other browsers and not just Safari

PR #2548, opened 2023-12-05
chore(on-change):  check data-mutation-free of parent nodes

### LINKED ISSUE #2205: 💡 Optional mutation observer for block tools

bettysteger:
- The `data-mutation-free` attribute now being checked for all parents of changed node
- "characterData" mutations also now could be muted

Resolves #2205

> I am rendering Vue components as Block Tools, for example a Video.js player. W I just play the video the editorjs gets a "block-changed" event and that's simply not true. The Video.js player creates lots of divs, where i have no "power" to set the data-mutation-free attribute!

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -18,6 +18,7 @@
 - `Improvement` — Now you can set focus via arrows/Tab to "contentless" (decorative) blocks like Delimiter which have no inputs.
 - `Improvement` — Inline Toolbar sometimes opened in an incorrect position. Now it will be aligned by the left side of the selected text. And won't overflow the right side of the text column.
 - `Improvement` - Now the `data-mutation-free` supports deep nesting, so you can mark some element with it to prevent the onChange call caused by child element mutating
+- `Improvement` - Now the `data-mutation-free` also allows to skip "characterData" mutations (eg. text content change)
 - `Refactoring` — `ce-block--focused` class toggling removed as unused.
 
 ### 2.28.2
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -12,7 +12,7 @@
 - `Fix` — Editor wrapper element growing on the Inline Toolbar close
 - `Fix` — Fix errors thrown by clicks on a document when the editor is being initialized
 - `Fix` — Inline Toolbar sometimes opened in an incorrect position. Now it will be aligned by the left side of the selected text. And won't overflow the right side of the text column.
-- `Feat` - Add deep check of `data-mutation-free` for mutation observer
+- `Improvement` - Now the `data-mutation-free` supports deep nesting, so you can mark some element with it to prevent the onChange call caused by child element mutating
 
 ### 2.28.2
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -12,6 +12,7 @@
 - `Fix` — Editor wrapper element growing on the Inline Toolbar close
 - `Fix` — Fix errors thrown by clicks on a document when the editor is being initialized
 - `Fix` — Inline Toolbar sometimes opened in an incorrect position. Now it will be aligned by the left side of the selected text. And won't overflow the right side of the text column.
+- `Feat` - Add deep check of `data-mutation-free` for mutation observer
 
 ### 2.28.2
 
--- a/src/components/block/index.ts
+++ b/src/components/block/index.ts
@@ -911,7 +911,7 @@ export default class Block extends EventsDispatcher<BlockEvents> {
             return false;
           }
 
-          return (node as HTMLElement).dataset.mutationFree === 'true';
+          return (node as HTMLElement).dataset.mutationFree === 'true' || !!node.closest('[data-mutation-free="deep"]');
         });
       });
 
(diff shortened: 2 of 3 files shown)


src/components/block/index.ts
```diff
@@ -911,7 +911,7 @@ export default class Block extends EventsDispatcher<BlockEvents> {
             return false;
           }
 
-          return (node as HTMLElement).dataset.mutationFree === 'true';
+          return (node as HTMLElement).dataset.mutationFree === 'true' || !!node.closest('[data-mutation-free="deep"]');
```
neSpecc: Do we really need this "deep" case? Why not always check for `.closest()`?

Is there any case when an element marked as `mutation-free` contains mutations that are expected to lead onChange call?

test/cypress/tests/onchange.cy.ts
```diff
@@ -544,6 +544,73 @@ describe('onChange callback', () => {
...
+     * Emulate tool's internal attribute mutation
+     */
+    cy.wait(100).then(() => {
+      toolChild.setAttribute('some-changed-attr', 'some-new-value');
+    });
+
+    /**
+     * Check that onChange callback was not called
+     */
+    cy.wait(500).then(() => {
```
neSpecc, line 609: It would be better to decrease waiting as much as possible to not throttle CI

src/components/block/index.ts
bettysteger: no we do not really need the "deep", i can change that!

neSpecc: I'm just asking, maybe there is some case that I don't see

test/cypress/tests/onchange.cy.ts
bettysteger: @neSpecc i just copied the previous test and changed it a little bit, see https://github.com/bettysteger/editor.js/commit/9e64003c0a37d1ba47bc912857e0d0afe5bb2260 

but i can decrease

src/components/block/index.ts
bettysteger: in my case no, i just did it because of backwards compatibility.

test/cypress/tests/onchange.cy.ts
```diff
@@ -539,7 +539,74 @@ describe('onChange callback', () => {
     /**
      * Check that onChange callback was not called
      */
-    cy.wait(500).then(() => {
+    cy.wait(200).then(() => {
+      cy.get('@onChange').should('have.callCount', 0);
+    });
+  });
+
+  it('should not be fired when child element of "data-mutation-free=deep" mark changes some attribute', () => {
```
neSpecc:
```suggestion
  it('should not be fired when mutation happened in a child of element with the "data-mutation-free" mark', () => {
```

src/components/block/index.ts
```diff
@@ -911,7 +911,7 @@ export default class Block extends EventsDispatcher<BlockEvents> {
             return false;
           }
 
-          return (node as HTMLElement).dataset.mutationFree === 'true';
+          return !!node.closest('[data-mutation-free="true"]');
```
neSpecc:
```suggestion
          return node.closest('[data-mutation-free="true"]') !== null;
```

neSpecc (commented):
Add a line to the CHANGELOG.md please

docs/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 - `Fix` — Editor wrapper element growing on the Inline Toolbar close
 - `Fix` — Fix errors thrown by clicks on a document when the editor is being initialized
 - `Fix` — Inline Toolbar sometimes opened in an incorrect position. Now it will be aligned by the left side of the selected text. And won't overflow the right side of the text column.
+- `Feat` - Add deep check of `data-mutation-free` for mutation observer
```
```suggestion
- `Improvement` - Now the `data-mutation-free` supports deep nesting, so you can mark some element with it to prevent the onChange call caused by child element mutating
```

neSpecc (approved):

gohabereg (approved):

ilyamore88 (approved):

PR #2555, opened 2023-12-09
feat: Skip workflow and do not bump version if it was changed already

GoldenJaden:
It now doesn't throw an error. Instead it sets the workflow status to "skipped'

### DIFF
--- a/.github/workflows/bump-version-on-merge-next.yml
+++ b/.github/workflows/bump-version-on-merge-next.yml
@@ -31,13 +31,12 @@ jobs:
         id: packageOld
         uses: codex-team/action-nodejs-package-info@v1
 
-      # Stop workflow and do not bump version if it was changed already
-      - name: Stop workflow and do not bump version if it was changed already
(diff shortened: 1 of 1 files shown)


.github/workflows/bump-version-on-merge-next.yml
```diff
@@ -32,12 +32,11 @@ jobs:
         uses: codex-team/action-nodejs-package-info@v1
 
       # Stop workflow and do not bump version if it was changed already
-      - name: Stop workflow and do not bump version if it was changed already
-        uses: actions/github-script@v3
+      - name: Stop workflow if version was not changed
```
neSpecc: Comment and name say opposite. What's the correct?

GoldenJaden: Changed the action name, check now

neSpecc (approved):

ilyamore88 (approved):

PR #2565, opened 2023-12-17
remve enter new blocks and remove css path svg

cloudy95:


maxheyn:
This PR seems like nonsense.

PR #2569, opened 2023-12-21
feat(ui): native-like tab behaviour, slash for toolbox

### LINKED ISSUE #927: Tab lock
Mouse is not required, you can do all manipulations from the keyboard only.

Enter — to create a new block.
Shift+Arrows — to select fragment
Tab — to open Toolbox at empty block
Tab — to open Block Settings at non-empty block
Tab (and Shift+Tab) — to switch focused button at the Toolbox, Inline Toolbar or Block Settings panels

and so on.

If we are missed something that you can not do without mouse, please create a separate Issue describing a problem.

_Originally posted by @neSpecc in https://github.com/codex-team/editor.js/issues/672#issuecomment-545424112_

All of this works, but I can't figure out how I can get out of the editor and to the next element in my form. I tried the arrows in the last element and tabbing. Both on an empty block and a block with content

neSpecc:
## Problem

1. Tab is not working for native-like controls focusing.
2. We can't step out of Editor using keyboard only

## Cause

Historically, we were using Tab to open Toolbox/Block Tunes. In empty block in was opening the Toolbox, and in non-empty block — the Block Tunes.

### Slash

- Now the `/` char typed in empty block will open the Toolbox
- `CMD+/` could be used to open the Block Tunes

### New Tab behaviour

- By pressing Tab, we will navigate to the next input/Block
- By pressing Shift+Tab, we will navigate to the previous input/Block
- By pressing Tab at the last Block, the next control out of the Editor could be focused. And the Shift+Tab at the first Block will focus a previous one.
- If the Block that is about to be focused does not contain inputs (like [Delimiter](https://github.com/editor-js/delimiter)), the whole Block will be highlighted as focused

https://github.com/codex-team/editor.js/assets/3684889/8aa50d9d-9282-4f59-a4d0-80f02ecd0842

Resolves #927

### Arrow navigation

Arrow navigation now also allows to focus Block that does not contain inputs, like Delimiter

### Other changes

- "block.focused" state has been removed since it has not been used for a long time. We use `block.selected` for focused state, and the "focused" property just was misleading.

### DIFF
--- a/src/components/modules/blockEvents.ts
+++ b/src/components/modules/blockEvents.ts
@@ -217,7 +217,6 @@ export default class BlockEvents extends Module {
       return;
     }
 
-    this.Editor.BlockSelection.clearSelection();
     this.activateBlockSettings();
   }
 
@@ -227,6 +226,10 @@ export default class BlockEvents extends Module {
   private slashPressed(): void {
     const currentBlock = this.Editor.BlockManager.currentBlock;
     const canOpenToolbox = currentBlock.isEmpty;
+
+    /**
+     * @todo Handle case when slash pressed when several blocks are selected
+     */
 
     /**
      * Toolbox will be opened only if Block is empty
--- a/src/components/modules/blockEvents.ts
+++ b/src/components/modules/blockEvents.ts
@@ -52,6 +52,13 @@ export default class BlockEvents extends Module {
       case _.keyCodes.TAB:
         this.tabPressed(event);
         break;
+      case _.keyCodes.SLASH:
+        if (event.ctrlKey || event.metaKey) {
+          this.commandSlashPressed();
+        } else {
+          this.slashPressed();
+        }
+        break;
     }
   }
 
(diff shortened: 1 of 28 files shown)


ilyamore88 (approved):

src/components/modules/blockEvents.ts
```diff
@@ -52,6 +52,13 @@ export default class BlockEvents extends Module {
       case _.keyCodes.TAB:
         this.tabPressed(event);
         break;
+      case _.keyCodes.SLASH:
+        if (event.ctrlKey || event.metaKey) {
```
gohabereg, line 56: Do we need both ctrl and meta keys?
On windows meta key is Win button

src/components/modules/toolbar/index.ts
```diff
@@ -411,13 +412,16 @@ export default class Toolbar extends Module<ToolbarNodes> {
...
-    tooltip.onHover(
-      this.nodes.settingsToggler,
-      I18n.ui(I18nInternalNS.ui.blockTunes.toggler, 'Click to tune'),
-      {
-        hidingDelay: 400,
-      }
-    );
+    const blockTunesTooltip = $.make('div');
+
+    blockTunesTooltip.appendChild(document.createTextNode(I18n.ui(I18nInternalNS.ui.blockTunes.toggler, 'Click to tune')));
```
gohabereg: Can you maybe extract something here into variable. Very hard to read

test/cypress/tests/modules/BlockEvents/Tab.cy.ts
```diff
@@ -0,0 +1,365 @@
...
+        ],
+      },
+    });
+
+    cy.get('[data-cy=editorjs]')
+      .find('.ce-paragraph')
+      .first()
+      .click()
+      .trigger('keydown', { keyCode: 9 })
+      .wait(100);
```
gohabereg: You could use constants here

src/components/modules/blockEvents.ts
neSpecc: yes, it's a standard statement to Cmd/Ctrl key

gohabereg (approved):

neSpecc:
> Great work here! When will this be pushed to npm?

it's already available under `2.29.0-rc.8`. The main release of 2.29 will be published soon (in 2-3 days)

PR #2579, opened 2024-01-02
Add `editor.tools` API, to update the tools' user provided config

### LINKED ISSUE #2397: Updating list of tools after editor is initialized
Hi there,

I was wondering, what would be the preferred method of updating a list of tools once the editor is already active and visible for users? I saw that there is a setter for configuration that handles new config, but I am not sure how to make `EditorJS` object utilize that new info. Right now setting new value won't update the editor.

My use case revolves around adding or removing toolbar items based on some external user config. I know that some blocks may not render after such a thing happens, but that is something I don't have a problem with as I know users won't mind losing a block when they disable some tool in their config.

Is destruction and reinitialization the only way to achieve this?

ShravanKarnati:
- Fixes #2397
- Adds a new API `tools` on the editor instance, which can now be used to update the tools' config without having to destroy and reinitialize the editor.


ShravanKarnati:
@neSpecc Could you take a look at this please?

PR #2581, opened 2024-01-04
Add a check that allows inline toolboxes to work with nested editorsjs

loucass003:
In our project, we have cases where we can have nested editorjs.
For instance, we have a block tool that will create an editorjs instance inside of it.

We ran into a situation where the inline toolbox would close when opening the link tool inside a nested editor.
This is due to the editor checking for the parent block only to verify if we clicked outside the toolbox.

We added a check that should prevent this issue (at least it does for us) and also keep the standard functionality of editorjs.
We hope that this change can be merged so we can get rid of our fork and make everyone else benefit from it.

Respectfully, Lucas Lelievre

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,8 @@
     "tools:update": "yarn _tools:checkout && yarn _tools:pull && yarn _tools:make",
     "test:e2e": "yarn build:test && cypress run",
     "test:e2e:open": "yarn build:test && cypress open",
-    "devserver:start": "yarn build && node ./devserver.js"
+    "devserver:start": "yarn build && node ./devserver.js",
+    "prepare": "npm run build"
   },
   "author": "CodeX",
   "license": "Apache-2.0",
@@ -75,5 +76,8 @@
   "collective": {
     "type": "opencollective",
     "url": "https://opencollective.com/editorjs"
+  },
+  "dependencies": {
+    "yarn": "^1.22.21"
   }
 }
--- a/src/components/modules/ui.ts
+++ b/src/components/modules/ui.ts
@@ -820,9 +820,13 @@ export default class UI extends Module<UINodes> {
 
     /**
      * Event can be fired on clicks at non-block-content elements,
-     * for example, at the Inline Toolbar or some Block Tune element
+     * for example, at the Inline Toolbar or some Block Tune element.
+     * We also make sure that the closest block belongs to the current editor and not a parent
      */
-    const clickedOutsideBlockContent = focusedElement.closest(`.${Block.CSS.content}`) === null;
+    const closestBlock = focusedElement.closest(`.${Block.CSS.content}`);
+    const clickedOutsideBlockContent =
+      closestBlock === null
+      || (closestBlock.closest(`.${Selection.CSS.editorWrapper}`) !== this.nodes.wrapper);
 
     if (clickedOutsideBlockContent) {
       /**


2tanhamon:
Hello @loucass003, i also have nested EditorJs and also have the same problem. But i have and another broblem with nested Editor js, maybe did u meet the same problem too? https://github.com/codex-team/editor.js/issues/2585

loucass003:
> Hello @loucass003, i also have nested EditorJs and also have the same problem. But i have and another broblem with nested Editor js, maybe did u meet the same problem too? #2585

Sorry we have not. We do not use react for our project maybe something related with that?

calumk:
It would be really good if this could be merged - It would fix an ongoing issue with a popular plugin : 

https://github.com/calumk/editorjs-columns/issues/3

sureshHARDIYA:
Could we please prioritize this?

package.json
neSpecc, line 1: please, remove unrelated changes

src/components/modules/ui.ts
```diff
@@ -820,9 +820,13 @@ export default class UI extends Module<UINodes> {
 
     /**
      * Event can be fired on clicks at non-block-content elements,
-     * for example, at the Inline Toolbar or some Block Tune element
+     * for example, at the Inline Toolbar or some Block Tune element.
+     * We also make sure that the closest block belongs to the current editor and not a parent
      */
-    const clickedOutsideBlockContent = focusedElement.closest(`.${Block.CSS.content}`) === null;
+    const closestBlock = focusedElement.closest(`.${Block.CSS.content}`);
+    const clickedOutsideBlockContent =
```
neSpecc, line 827: add corresponding test case

vamidi:
Would love to see this fixed =D

calumk:
@loucass003 do you have time to write a test for this and submit a clean PR? 

@neSpecc I can submit a clean PR but im not familiar with cypress so would probably take me a while to work out how to write the test for this

calumk:
@loucass003 - sorry to bump, but as you originally proposed this, do you have time to write a test for this and submit a clean PR?

perssonrichard:
We're also working with nested EditorJS instances so this would be more than great.

calumk:
Hi @neSpecc  - Can you please provide some guidance as to how exactly to test for this?

The issue as i see it is this specific feature (nested editor.js) doesnt exist in any of the default plugins. 

So to test with cypress, unless I misunderstand, we would need to add a 3rd-party example plugin to the package devDependencies

This might be something you are unwilling to merge

Either we would need to create a new plugin to test the simpleist intergration, or use an exisiting one, such as my own [@calumk/editor-js-columns](https://github.com/calumk/editorjs-columns) 

Thoughts?

gorenburg:
this one is blocker for us as well. what approach should be used for the tests?

neSpecc:
What about testing, you can create a Tool Mock what will render editor.js inside own render method. Example https://github.com/codex-team/editor.js/blob/next/test/cypress/tests/modules/BlockEvents/Backspace.cy.ts#L66

gorenburg:
created new PR to fix this issue with tests: https://github.com/codex-team/editor.js/pull/2780

calumk:
@gorenburg amazing 🤩

calumk:
@neSpecc Are the tests @gorenburg provided sufficiant?

gorenburg:
This PR can be closed since it was fixed with https://github.com/codex-team/editor.js/pull/2780

PR #2599, opened 2024-01-27
fix(slash): do not handle / + shift/alt, support for ascii keyboard

### LINKED ISSUE #2597: Slash doesn't open toolbox
With v2.29.0 Tab doesn't open the Toolbox anymore since it was changed to slash.

However, when typing a "/" in an empty line, the toolbox doesn't appear, only the slash gets typed.

neSpecc:
## The problem

1. Right now Toolbox is opening even when Slash is pressed in a combination with Shift or Alt.
2. Toolbox is not opening for Slash press in non-default keyboard layouts 

https://github.com/codex-team/editor.js/assets/3684889/5b324819-69f4-49b4-86cd-210b12e0ee17

## Cause

1. We're checking all keyboard keys using `event.keyCode` property with is not working properly in some keyboard layout.
2. There was no check for Shift or Alt

## Solution

1. To open Toolbox we now use `event.key=/` to check exact "/" typed. So check for Shift or Alt is not needed, since `key` will be different in that case.
2. To open Block Tunes we now use `event.code=Slash` to check if slash-like key pressed with `cmd`. For example, in Spanish keyboard it will be `cmd+-`

Resolves #2597

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+### 2.29.1
+
+- `Fix` — Toolbox wont be shown when Slash pressed with along with Shift or Alt
+- `Fix` — Toolbox will be opened when Slash pressed in ASCII-capable keyboard layout.
+
 ### 2.29.0
 
 - `New` — Editor Config now has the `style.nonce` attribute that could be used to allowlist editor style tag for Content Security Policy "style-src"
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -3,7 +3,7 @@
 ### 2.29.1
 
 - `Fix` — Toolbox wont be shown when Slash pressed with along with Shift or Alt
-- `Fix` — Toolbox will be opened when Slash pressed in ASCII-capable keyboard layout.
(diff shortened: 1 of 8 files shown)


gohabereg (approved):

docs/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+### 2.29.1
+
+- `Fix` — Toolbox wont be shown when Slash pressed with along with Shift or Alt
```
ilyamore88, line 5: Control or meta keys

neSpecc: `CMD+/` and `CTRL+/` open Block Tunes — this behaviour stays the same.
Problem was in `Shift+/` and `Alt+/` — Toolbox won't be shown now, since "key" will be  "?" or "÷" — we don't need to check Shift or Alt separately. Previously we checked `keyCode` which is the same for those chars.

ilyamore88: Oh, okay. Shift is a bit weird for me here, because in some layouts I can't type / without shift.

ilyamore88 (approved):

PR #2639, opened 2024-02-27
fix issue #2523

### LINKED ISSUE #2523: Bug: editor.onChange is not triggered in case of removing the whole text inside block, if the contenteditable element is not the root node of a block
https://github.com/codex-team/editor.js/blob/ee188bfe5de9c99592396250f6fce90f63b2957e/src/components/utils/mutations.ts#L15 - 
this check won't help if the change happened within some descendant of the 'element'.

For example, if we remove the text inside one of the 'li' items of '@editorjs/nested-list' block, onChange won't be triggered.

VikhorKonstantin:
fix #2523:

fixing the following function: https://github.com/codex-team/editor.js/blob/ee188bfe5de9c99592396250f6fce90f63b2957e/src/components/utils/mutations.ts#L7 
Made this function return `true` if the whole text node is removed inside of some descendant of the passed `element`.

This PR addresses the issue where the onChange is not triggered if the text inside one of the 'li' items of the '@editorjs/nested-list' block is removed.

### DIFF
--- a/src/components/utils/mutations.ts
+++ b/src/components/utils/mutations.ts
@@ -7,29 +7,20 @@
 export function isMutationBelongsToElement(mutationRecord: MutationRecord, element: Element): boolean {
   const { type, target, addedNodes, removedNodes } = mutationRecord;
 
-  /**
-   * In case of removing the whole text in element, mutation type will be 'childList',
-   * 'removedNodes' will contain text node that is not existed anymore, so we can't check it with 'contains' method
-   * But Target will be the element itself, so we can detect it.
-   */
-  if (target === element) {
+  if (element.contains(target)) {
     return true;
   }
 
-  /**
-   * Check typing and attributes changes
-   */
-  if (['characterData', 'attributes'].includes(type)) {
-    const targetElement = target.nodeType === Node.TEXT_NODE ? target.parentNode : target;
-
-    return element.contains(targetElement);
+  if (type === 'childList') {
+    const elementAddedItself = Array.from(addedNodes).some(node => node.contains(element));
+    if (elementAddedItself) {
+      return true;
+    }
+    const elementRemovedItself = Array.from(removedNodes).some(node => node.contains(element));
+    if (elementRemovedItself) {
+      return true;
+    }
   }
 
-  /**
-   * Check new/removed nodes
-   */
-  const addedNodesBelongsToBlock = Array.from(addedNodes).some(node => element.contains(node));
-  const removedNodesBelongsToBlock = Array.from(removedNodes).some(node => element.contains(node));
-
-  return addedNodesBelongsToBlock || removedNodesBelongsToBlock;
+  return false;
 }
--- a/src/components/utils/mutations.ts
+++ b/src/components/utils/mutations.ts
@@ -11,6 +11,9 @@ export function isMutationBelongsToElement(mutationRecord: MutationRecord, eleme
     return true;
   }
 
+  /**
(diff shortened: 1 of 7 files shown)


src/components/utils/mutations.ts
```diff
@@ -7,29 +7,20 @@
 export function isMutationBelongsToElement(mutationRecord: MutationRecord, element: Element): boolean {
   const { type, target, addedNodes, removedNodes } = mutationRecord;
 
-  /**
-   * In case of removing the whole text in element, mutation type will be 'childList',
-   * 'removedNodes' will contain text node that is not existed anymore, so we can't check it with 'contains' method
-   * But Target will be the element itself, so we can detect it.
-   */
```
neSpecc: why comments are removed?

neSpecc (changes requested):

src/components/utils/mutations.ts
neSpecc, line 1: add corresponding tests please

src/components/utils/mutations.ts
VikhorKonstantin: @neSpecc Hi, I've restructured the code a little bit and replaced the part this comment was addressed to.

src/components/utils/mutations.ts
VikhorKonstantin: Done

src/components/utils/mutations.ts
neSpecc: So you need to write the new comments explaining all logical parts.

P.S. Please, do not resolve other people's comments. Reply instead.

test/cypress/tests/onchange.cy.ts
```diff
@@ -1,5 +1,6 @@
 import Header from '@editorjs/header';
 import Code from '@editorjs/code';
+import Checklist from '@editorjs/checklist';
```
neSpecc: Is this dependency really required? It would be better to use mock instead, check `fixtures/tools/ToolMock.ts`

VikhorKonstantin: Replaced with a ToolMock

src/components/utils/mutations.ts
VikhorKonstantin: Added some comments

src/components/utils/mutations.ts
```diff
@@ -7,29 +7,23 @@
...
-  if (target === element) {
+  if (element.contains(target)) {
     return true;
   }
 
   /**
-   * Check typing and attributes changes
+   * In case of removing/adding the element itself, mutation type will be 'childList' and 'removedNodes'/'addedNodes' will contain the element.
    */
-  if (['characterData', 'attributes'].includes(type)) {
```
neSpecc, line 22: Am I understanding it right that `characterData` and `attributes` mutations covered by `element.contains(target)` statement? Docs are not clear

VikhorKonstantin: Yes, `element.contains(target)` covers all mutations happened to the element or it's descendants with the only one exception - removing/adding the element itself, because in this case target will be the element's parent and the element itself would be inside 'removedNodes'/'addedNodes'.

VikhorKonstantin: Added more explanation + fixed formatting.

neSpecc (commented):
Seems ok, please, drop a line at the `docs/CHANGELOG.md` under the new 2.30.0 section

VikhorKonstantin:
> Seems ok, please, drop a line at the `docs/CHANGELOG.md` under the new 2.30.0 section

Done

neSpecc (approved):

gohabereg (approved):

PR #2644, opened 2024-02-29
Clear highlights and pointer on BlockManager even if selection was inside of the editor before clicking outside

### LINKED ISSUE #2643: Bug: clicking outside of an editor should clear currentBlock (call dropPointer) even if selection was inside of editor before the click
Describe a bug.

Steps to reproduce:
1. Go to example_dev.html
2. Click on somewhere outside of the editor
3. Press '/'

Expected behavior:
Toolbox shouldn't be opened, because I've already went out of the editor by clicking outside

Actual:
Toolbox is still opened because we haven't cleared currentBlock

Device, Browser, OS:
Any

Editor.js version:
2.29.1

VikhorKonstantin:
Fix: #2643 

Clear highlights and pointer on BlockManager even if selection was inside of the editor before clicking outside

### DIFF
--- a/src/components/modules/ui.ts
+++ b/src/components/modules/ui.ts
@@ -620,7 +620,7 @@ export default class UI extends Module<UINodes> {
      * Do not fire check on clicks at the Inline Toolbar buttons
      */
     const target = event.target as HTMLElement;
-    const clickedInsideOfEditor = this.nodes.holder.contains(target) || Selection.isAtEditor;
+    const clickedInsideOfEditor = this.nodes.holder.contains(target);
 
     if (!clickedInsideOfEditor) {
       /**
--- a/src/components/modules/ui.ts
+++ b/src/components/modules/ui.ts
@@ -611,10 +611,12 @@ export default class UI extends Module<UINodes> {
     /**
      * Sometimes we emulate click on some UI elements, for example by Enter on Block Settings button
      * We don't need to handle such events, because they handled in other place.
+     *
+     * PS: Commented this out because it makes this method behave differently in tests. Cypress-generated clicks are having isTrusted == false.
      */
-    if (!event.isTrusted) {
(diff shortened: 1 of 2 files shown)


src/components/modules/ui.ts
```diff
@@ -620,7 +620,7 @@ export default class UI extends Module<UINodes> {
      * Do not fire check on clicks at the Inline Toolbar buttons
      */
     const target = event.target as HTMLElement;
-    const clickedInsideOfEditor = this.nodes.holder.contains(target) || Selection.isAtEditor;
```
neSpecc, line 623: I can't remember why it was added. Anyway, I've tested it and it looks like everything is ok. So please add a test (at the `Ui.cy.ts`) and a change log

VikhorKonstantin: @neSpecc, I just tried to write some tests and encountered an issue with the `ui.ts:documentClicked` function:
image
It appears that this check causes clicks made by Cypress `click()` to behave differently from user-generated clicks (because Cypress-generated clicks will always have isTrusted == false). I'm concerned that this might break other tests as well.

VikhorKonstantin: @neSpecc, I've commented this out with an explanation in a comment. Seems like we're not triggering Click events anywhere in the code so it shouldn't break anything.

neSpecc: I've found a bug with Block Tunes pressing by Enter. It sometimes removes the block itself (like Enter pressed when block is selected). It can be reproduces in the `next` so I'm gonna dig in it soon. 

They can be related

neSpecc: Hey @VikhorKonstantin. I've resolved several problems with Enter press. Please, pull the `next` and check if the problem still actual

PR #2649, opened 2024-03-09
feat: nested popover

TatianaFomina:
🏆 Popover now has support for nested items
🎁 Available in both mobile and desktop versions
💅 Any level of nesting is supported

image

image

image


src/components/dom.ts
```diff
@@ -56,6 +56,7 @@ export default class Dom {
     const el = document.createElement(tagName);
 
     if (Array.isArray(classNames)) {
+      classNames = classNames.filter(className => className !== undefined);
```
neSpecc: pls leave a comment about where it is needed

src/components/utils/classnames.ts
```diff
@@ -0,0 +1,21 @@
...
+const MODIFIER_DELIMITER = '--';
+
+/**
+ * Utility function that allows to construct class names from block and element names
+ *
+ * @param blockName - string with block name
+ * @param elementName - string with element name
+ * @param modifier - modifier to be appended
+ */
+export function cn(blockName: string) {
```
neSpecc: lets call it `bem()`?

src/components/utils/popover/components/popover-header/popover-header.const.ts
```diff
@@ -0,0 +1,15 @@
+import { cn } from '../../../classnames';
+
+/**
+ * Popover header block CSS class constructor
+ */
+const popoverHeaderCn = cn('ce-popover-header');
```
neSpecc:
```suggestion
const popoverHeaderClassName = cn('ce-popover-header');
```

src/components/utils/popover/index.ts
```diff
@@ -112,24 +51,42 @@ export default class Popover extends EventsDispatcher<PopoverEventMap> {
...
+   * History of popover states for back navigation.
+   * Is used for mobile version of popover,
+   * where we can not display nested popover of the screen and
+   * have to render nested items in the same popover switching to new state
+   */
+  private history = new PopoverStatesHistory();
+
+  /**
+   * Last hovered item inside popover.
+   * Is used to track hovered item changes.
```
neSpecc: > Is used to track hovered item changes

pls add more details. 

And describe when it can be `undefined` and `null`

src/components/utils/popover/index.ts
```diff
@@ -373,15 +359,39 @@ export default class Popover extends EventsDispatcher<PopoverEventMap> {
...
+    }
+
     if (item.isDisabled) {
       return;
     }
 
+    if (item.children.length > 0) {
+      if (isMobileScreen()) {
+        /** Show nested items */
+        this.applyPopoverState(item.children, item.title);
```
neSpecc: its unclear what this method is doing. I'd suggest improving naming

src/components/utils/bem.ts
```diff
@@ -0,0 +1,21 @@
+const ELEMENT_DELIMITER = '__';
+const MODIFIER_DELIMITER = '--';
+
+/**
+ * Utility function that allows to construct class names from block and element names
+ *
+ * @param blockName - string with block name
+ * @param elementName - string with element name
+ * @param modifier - modifier to be appended
+ */
```
neSpecc, line 14: add `@example` pls

src/components/utils/popover/components/popover-header/popover-header.const.ts
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Popover header block CSS class constructor
+ */
+const popoverHeaderCn = bem('ce-popover-header');
+
+/**
+ * CSS class names to be used in popover header class
+ */
+export const cls = {
```
neSpecc:
```suggestion
export const css = {
```

src/components/utils/popover/components/popover-item/popover-item.ts
```diff
@@ -10,7 +11,7 @@ export class PopoverItem {
    * True if item is disabled and hence not clickable
    */
   public get isDisabled(): boolean {
-    return this.params.isDisabled;
+    return !!this.params.isDisabled;
```
neSpecc:
```suggestion
    return this.params.isDisabled === true;
```

src/components/utils/bem.ts
TatianaFomina: ➕

src/components/utils/popover/components/popover-header/popover-header.const.ts
TatianaFomina: ➕

src/components/utils/popover/components/popover-item/popover-item.ts
TatianaFomina: ➕

src/components/modules/toolbar/blockSettings.ts
```diff
@@ -56,8 +58,8 @@ export default class BlockSettings extends Module<BlockSettingsNodes> {
    *
    * @todo remove once BlockSettings becomes standalone non-module class
    */
-  public get flipper(): Flipper {
-    return this.popover?.flipper;
+  public get flipper(): Flipper | undefined {
+    return 'flipper' in this.popover ? this.popover?.flipper : undefined;
```
gohabereg, line 65: `this.popover?.flipper` should work the same here

src/components/ui/toolbox.ts
```diff
@@ -128,44 +136,21 @@ export default class Toolbox extends EventsDispatcher<ToolboxEventMap> {
...
+  public getElement(): HTMLElement {
     return this.nodes.toolbox;
   }
 
   /**
    * Returns true if the Toolbox has the Flipper activated and the Flipper has selected button
    */
   public hasFocus(): boolean | undefined {
-    return this.popover?.hasFocus();
+    return 'hasFocus' in this.popover ? this.popover?.hasFocus() : undefined;
```
gohabereg: Same hear, previous version should work just fine

src/components/ui/toolbox.ts
```diff
@@ -83,7 +91,7 @@ export default class Toolbox extends EventsDispatcher<ToolboxEventMap> {
   /**
    * Popover instance. There is a util for vertical lists.
    */
-  private popover: Popover | undefined;
+  private popover: PopoverDesktop | PopoverMobile | undefined;
```
gohabereg: Might worth it to create a union type as `PopoverDesktop | PopoverMobile` used several times across the codebase

src/components/utils/popover/components/popover-item/popover-item.ts
```diff
@@ -166,40 +138,53 @@ export class PopoverItem {
     this.disableSpecialHoverAndFocusBehavior();
   }
 
+  /**
+   * Returns list of item children
+   */
+  public get children(): PopoverItemParams[] {
+    return 'children' in this.params && this.params.children?.items !== undefined ? this.params.children.items : [];
```
gohabereg, line 149: `this.params.children?.items ?? []`

src/components/ui/toolbox.ts
gohabereg: Or just describe an interface with common props and methods

gohabereg (approved):
Would be great to make popover elements accessible for screen readers, could be done as a separate PR though

src/components/modules/toolbar/blockSettings.ts
TatianaFomina: Nah, there will be TypeScript error. Property `flipper` exists only in PopoverDesktop and not in PopoverMobile, that's why we need to check weather the property is present

src/components/ui/toolbox.ts
TatianaFomina: there also will be TypeScript error. `hasFocus` exists only in PopoverDesktop and not in PopoverMobile, that's why we need to check weather the property is present

src/components/utils/popover/components/popover-item/popover-item.ts
TatianaFomina: 🥲
image

src/components/ui/toolbox.ts
TatianaFomina: done

ilyamore88 (approved):

src/components/modules/toolbar/blockSettings.ts
```diff
@@ -7,7 +7,9 @@ import { I18nInternalNS } from '../../i18n/namespace-internal';
 import Flipper from '../../flipper';
 import { TunesMenuConfigItem } from '../../../../types/tools';
 import { resolveAliases } from '../../utils/resolve-aliases';
-import Popover, { PopoverEvent } from '../../utils/popover';
+import { Popover, PopoverDesktop, PopoverMobile } from '../../utils/popover';
+import { PopoverEvent } from '../../utils/popover/popover.typings';
```
neSpecc: use `import type` when only types are imported
```suggestion
import { type Popover, PopoverDesktop, PopoverMobile } from '../../utils/popover';
import type { PopoverEvent } from '../../utils/popover/popover.typings';
```

src/components/modules/toolbar/blockSettings.ts
```diff
@@ -68,8 +72,7 @@ export default class BlockSettings extends Module<BlockSettingsNodes> {
   /**
    * Popover instance. There is a util for vertical lists.
```
neSpecc, line 74: Describe when it can be null

src/components/modules/toolbar/index.ts
```diff
@@ -436,6 +445,8 @@ export default class Toolbar extends Module<ToolbarNodes> {
      * Append toolbar to the Editor
      */
     $.append(this.Editor.UI.nodes.wrapper, this.nodes.wrapper);
+
+    this.listeners.on(window, 'resize', this.handleResize);
```
neSpecc: we already have a window resize handler in `ui.ts`. It calls `checkIsMobile()`. We can add event triggering in that method: when `isMobile` value changed, event will be fired in our Event Bus. Thus you can subscribe to that event directly which will make a logic more clear. What do you think?

src/components/modules/toolbar/index.ts
```diff
@@ -104,6 +104,14 @@ export default class Toolbar extends Module<ToolbarNodes> {
    */
   private toolboxInstance: Toolbox | null = null;
 
+  /**
+   * Handles window resize. Closes block tunes and toolbox in case they are open
+   */
+  private handleResize = _.debounce((): void => {
+    this.Editor.BlockSettings.close();
+    this.toolboxInstance.forceRemountPopover();
```
neSpecc: seems a little bit strange that BlockSettings and Toolbox has a different logic. 

I guess you wanted to remount popover in both cases, so it would be more expected if logic will be similar: both close or both forceRemount

src/components/ui/toolbox.ts
```diff
@@ -83,7 +91,7 @@ export default class Toolbox extends EventsDispatcher<ToolboxEventMap> {
   /**
    * Popover instance. There is a util for vertical lists.
```
neSpecc, line 93: describe when it can be nulll

src/components/ui/toolbox.ts
```diff
@@ -226,6 +216,67 @@ export default class Toolbox extends EventsDispatcher<ToolboxEventMap> {
...
+      scopeElement: this.api.ui.nodes.redactor,
+      searchable: true,
+      messages: {
+        nothingFound: this.i18nLabels.nothingFound,
+        search: this.i18nLabels.filter,
+      },
+      items: this.toolboxItemsToBeDisplayed,
+    });
+
+    this.popover?.on(PopoverEvent.Close, this.onPopoverClose);
```
neSpecc: `?` seems redundant here

src/components/utils/popover/components/popover-header/popover-header.const.ts
```diff
@@ -0,0 +1,15 @@
+import { bem } from '../../../bem';
+
+/**
+ * Popover header block CSS class constructor
+ */
+const popoverHeaderCn = bem('ce-popover-header');
```
neSpecc:
```suggestion
const popoverHeaderClassName = bem('ce-popover-header');
```

src/components/ui/toolbox.ts
```diff
@@ -128,44 +136,25 @@ export default class Toolbox extends EventsDispatcher<ToolboxEventMap> {
     this.api = api;
     this.tools = tools;
     this.i18nLabels = i18nLabels;
+    this.make();
```
neSpecc: Can't test it at the moment, but we need to check if the Toolbox will be created in read-only mode. It should not be created.

src/components/utils/popover/components/popover-item/popover-item.ts
```diff
@@ -211,6 +196,10 @@ export class PopoverItem {
    * @param newState - new popover item params that should be applied
    */
   private enableConfirmationMode(newState: PopoverItemParams): void {
+    if (this.nodes.root === null) {
```
neSpecc, line 203: How it could be possible? Add a jsdoc for this case, please

src/components/utils/popover/components/popover-item/popover-item.ts
```diff
@@ -230,10 +219,13 @@ export class PopoverItem {
    * Returns item to its original state
    */
   private disableConfirmationMode(): void {
+    if (this.nodes.root === null) {
```
neSpecc, line 226: same

src/components/utils/popover/components/search-input/search-input.ts
```diff
@@ -132,6 +115,10 @@ export default class SearchInput {
     this.wrapper.appendChild(this.input);
 
     this.listeners.on(this.input, 'input', () => {
+      if (this.input === undefined) {
```
neSpecc: seems strange, like a code that will never used

src/components/utils/popover/components/search-input/search-input.typings.ts
neSpecc, line 1: please use `.types.ts` instead

src/components/utils/popover/components/search-input/search-input.typings.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Item that could be searched
+ */
+export interface SearchableItem {
+  title?: string;
```
neSpecc: prop description missed

src/components/utils/popover/popover-abstract.ts
```diff
@@ -0,0 +1,318 @@
...
+  /**
+   * Retrieves popover item that is the target of the specified event
+   *
+   * @param event - event to retrieve popover item from
+   */
+  protected getTargetItem(event: Event): PopoverItem | undefined {
+    return this.items.find(el => {
+      const itemEl = el.getElement();
+
+      if (itemEl === null) {
```
neSpecc, line 167: when it can be possible? jsdoc would be helpful

src/components/utils/popover/popover-abstract.ts
```diff
@@ -0,0 +1,318 @@
...
+      textContent: this.messages.nothingFound,
+    });
+
+    this.nodes.popoverContainer.appendChild(this.nodes.nothingFoundMessage);
+    this.nodes.items = Dom.make('div', [ css.items ]);
+
+    this.items.forEach(item => {
+      const itemEl = item.getElement();
+
+      if (itemEl === null) {
```
neSpecc: same

src/components/utils/popover/popover-abstract.ts
```diff
@@ -0,0 +1,318 @@
...
+  private addSearch(): void {
+    this.search = new SearchInput({
+      items: this.items,
+      placeholder: this.messages.search,
+      onSearch: this.onSearch,
+    });
+
+    const searchElement = this.search.getElement();
+
+    if (searchElement === undefined) {
```
neSpecc: can it be a real-world case? or just a ts crutch? If second, it would be better to resolve it by typescript instead of adding a redundant code in a bundle.

src/components/utils/popover/popover-abstract.ts
```diff
@@ -0,0 +1,318 @@
...
+   *
+   * - Performs radiobutton-like behavior if the item has property 'toggle' set to string key.
+   * (All the other items with the same key get inactive, and the item gets active)
+   *
+   * @param clickedItem - popover item that was clicked
+   */
+  private toggleItemActivenessIfNeeded(clickedItem: PopoverItem): void {
+    if (clickedItem.toggle === true) {
+      clickedItem.toggleActive();
+    }
```
neSpecc, line 266: return?

src/components/utils/popover/popover-desktop.ts
```diff
@@ -0,0 +1,362 @@
...
+import { Popover as PopoverBase } from './popover-abstract';
+import { PopoverItem, css as popoverItemCls } from './components/popover-item';
+import { PopoverParams } from './popover.typings';
+import { keyCodes } from '../../utils';
+import { css } from './popover.const';
+import { SearchableItem } from './components/search-input';
+import { cacheable } from '../../utils';
+
+/**
+ * Class responsible for rendering popover and handling its behaviour on desktop
```
neSpecc:
```suggestion
 * On desktop devices popover behaves like a floating element. Nested popover appears at right or left side.
```

src/components/utils/popover/popover-desktop.ts
```diff
@@ -0,0 +1,362 @@
+import Flipper from '../../flipper';
+import { Popover as PopoverBase } from './popover-abstract';
```
neSpecc: Right now a class, a file and an import uses different naming. I'd suggest choosing a single name: Popover, PopoverAbstract or PopoverBase.

src/components/utils/popover/popover-desktop.ts
```diff
@@ -0,0 +1,362 @@
...
+  }
+
+  /**
+   * Clears memory
+   */
+  public destroy(): void {
+    super.destroy();
+    this.flipper?.deactivate();
+    this.destroyNestedPopoverIfExists();
+  }
```
neSpecc, line 161: `this.previouslyHoveredItem = null;` ?

src/components/utils/popover/popover-abstract.ts
```diff
@@ -0,0 +1,318 @@
...
+      });
+    }
+  }
+
+  /**
+   * Handles displaying nested items for the item. Behaviour differs depending on platform.
+   *
+   * @param item – item to show nested popover for
+   */
+  protected abstract handleShowingNestedItems(item: PopoverItem): void;
```
neSpecc: name and description are a bit confusing. What does it mean "handle displaying" or "handle showing"? Maybe something like `showNestedListIfNeeded`?

src/components/utils/popover/popover-desktop.ts
```diff
@@ -0,0 +1,362 @@
...
+  };
+
+  /**
+   * Handles displaying nested items for the item.
+   *
+   * @param item – item to show nested popover for
+   */
+  protected override handleShowingNestedItems(item: PopoverItem): void {
+    if (this.nestedPopover == null || this.nestedPopover === undefined) {
+      this.showNestedPopoverForItem(item);
```
neSpecc: It is a bit confusing: `if there is no nested popover, show nested popover`?

src/components/utils/popover/popover-desktop.ts
```diff
@@ -0,0 +1,362 @@
...
+    }
+  }
+
+  /**
+   * Checks if popover should be opened bottom.
+   * It should happen when there is enough space below or not enough space above
+   */
+  private get shouldOpenBottom(): boolean {
```
neSpecc, line 198: can it be moved to some utility?

src/components/utils/popover/popover-desktop.ts
```diff
@@ -0,0 +1,362 @@
...
+      nestingLevel: this.nestingLevel + 1,
+    });
+
+    const nestedPopoverEl = this.nestedPopover.getElement();
+
+    this.nodes.popover?.appendChild(nestedPopoverEl);
+    const itemEl =  item.getElement();
+    const itemOffsetTop = (itemEl ? itemEl.offsetTop : 0) - this.scrollTop;
+    // eslint-disable-next-line @typescript-eslint/no-magic-numbers
+    const topOffset = this.offsetTop + itemOffsetTop - 4;
```
neSpecc: magic number is unclear here. Lets move it to a const

src/components/utils/popover/popover-mobile.ts
```diff
@@ -0,0 +1,146 @@
...
+import { PopoverHeader } from './components/popover-header';
+import { PopoverStatesHistory } from './utils/popover-states-history';
+import { PopoverParams } from './popover.typings';
+import { PopoverItem } from './components/popover-item';
+import { PopoverItem as PopoverItemParams } from '../../../../types';
+import { css } from './popover.const';
+import Dom from '../../dom';
+
+/**
+ * Class responsible for rendering popover and handling its behaviour on mobile screens
```
neSpecc:
```suggestion
 * On mobile devices Popover behaves like a fixed panel at the bottom of screen. Nested item appears like "pages" with the "back" button
```

src/components/utils/popover/popover-mobile.ts
```diff
@@ -0,0 +1,146 @@
...
+
+    /** Re-render items */
+    this.items.forEach(item => item.getElement()?.remove());
+
+    this.items = items.map(params => new PopoverItem(params));
+
+    this.items.forEach(item => {
+      const itemEl = item.getElement();
+
+      if (itemEl === null) {
```
neSpecc, line 136: same

src/components/utils/popover/popover.typings.ts
```diff
@@ -0,0 +1,74 @@
...
+ * Params required to render popover
+ */
+export interface PopoverParams {
+  /**
+   * Popover items config
+   */
+  items: PopoverItemParams[];
+
+  /**
+   * Element of the page that creates 'scope' of the popover
```
neSpecc: Please, describe a bit more, what "scope" means here

src/components/utils/popover/popover.typings.ts
```diff
@@ -0,0 +1,74 @@
...
+   * When popover closes
+   */
+  Close = 'close'
+}
+
+/**
+ * Events fired by the Popover
+ */
+export interface PopoverEventMap {
+  [PopoverEvent.Close]: undefined;
```
neSpecc: pls add descriptions for events

src/components/utils/popover/popover-mobile.ts
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  /**
+   * Clears memory
+   */
+  public destroy(): void {
+    super.destroy();
+
+    this.scrollLocker.unlock();
+  }
```
neSpecc, line 83: do we need to clear a state history on `hide` and `destroy`?

src/styles/popover.css
```diff
@@ -134,6 +154,33 @@
   &__custom-content--hidden {
     display: none;
   }
+
+  &--nested {
+    .ce-popover__container {
+      /* Variable --nesting-level is set via js in showNestedPopoverForItem() method */
+      --popover-left: calc(var(--nesting-level) * (var(--width) - 4px));
```
neSpecc: please move 4px to a variable.

test/cypress/tests/utils/popover.cy.ts
```diff
@@ -257,4 +258,187 @@ describe('Popover', () => {
...
+
+    /** Check child item displayed */
+    cy.get('[data-cy=editorjs]')
+      .get('.ce-popover--nested .ce-popover__container')
+      .get('[data-item-name="nested-test-item"]')
+      .should('be.visible');
+  });
+
+
+  it('should dysplay nested popover (mobile)', () => {
```
neSpecc:
```suggestion
  it('should display nested popover (mobile)', () => {
```

test/cypress/tests/utils/popover.cy.ts
```diff
@@ -257,4 +258,187 @@ describe('Popover', () => {
...
+      .get('[data-item-name="nested-test-item"]')
+      .should('not.exist');
+
+    /** Check back button is not displayed */
+    cy.get('[data-cy=editorjs]')
+      .get('.ce-popover__container')
+      .get('.ce-popover-header__back-button')
+      .should('not.exist');
+
+    /** Check header is not displayed */
```
neSpecc, line 439: pls mention these checks in a test case name

src/components/modules/toolbar/blockSettings.ts
TatianaFomina: It is `null` until popover is not initialized

src/components/modules/toolbar/index.ts
TatianaFomina: Block tunes destroys popover on each close and creates new one on open. It's because each block might have different tunes. So, no need to add `forceRemountPopover()` method to BlockSettings. 

Toolbox does not recreate popover on open, so, we need to do it manually when screen size changes

src/components/modules/toolbar/index.ts
TatianaFomina: done

src/components/ui/toolbox.ts
TatianaFomina: It is null by default before it is initialized

src/components/ui/toolbox.ts
TatianaFomina: seems to be ok

src/components/ui/toolbox.ts
TatianaFomina: `?` is needed because `this.popover` might be `null`

src/components/utils/popover/components/popover-header/popover-header.const.ts
TatianaFomina: done

src/components/utils/popover/components/popover-item/popover-item.ts
TatianaFomina: `this.nodes.root` is `null` by default before initalization. This check is needed for typescript validation

src/components/utils/popover/components/search-input/search-input.ts
TatianaFomina: it's done for ts validation

src/components/utils/popover/components/search-input/search-input.typings.ts
TatianaFomina: done

src/components/utils/popover/components/search-input/search-input.typings.ts
TatianaFomina: done

src/components/utils/popover/popover-abstract.ts
TatianaFomina: it's not possible, it's just a check that prevents ts warnings

src/components/utils/popover/popover-abstract.ts
TatianaFomina: huh?

src/components/modules/toolbar/index.ts
TatianaFomina: ➕

src/components/utils/popover/popover-desktop.ts
TatianaFomina: ➕

src/components/utils/popover/popover-abstract.ts
TatianaFomina: To handle showing nested items means to do what is necessary to make them visible. This behaviour differs depending on platform: on desktop nested popover is displayed and on mobile nested items are rendered in the same popover

src/components/utils/popover/popover-desktop.ts
TatianaFomina: not this time

src/components/utils/popover/popover-mobile.ts
TatianaFomina: for now we don't – block tunes recreate popover on each open. In future, when we will need to integrate nested popover to toolbox, I will add clearing history

src/components/modules/toolbar/blockSettings.ts
neSpecc: move it to jsdoc please

src/components/events/WindowResize.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Fired when window is resized
+ */
+export const WindowResize = 'window resize';
```
neSpecc: event name should be in a past form
```suggestion
export const WindowResized = 'window resized';
```

src/components/events/WindowResize.ts
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Payload that will be passed with the event
+ */
+export interface WindowResizePayload {
+  /**
+   * True, if new window size is mobile
```
neSpecc:
```suggestion
   * True, if editor toggled to the mobile view
```

src/components/modules/ui.ts
```diff
@@ -427,6 +428,13 @@ export default class UI extends Module<UINodes> {
      * Detect mobile version
      */
     this.checkIsMobile();
+
+    /**
+     * Dispatch global event
+     */
+    this.eventsDispatcher.emit(WindowResize, {
```
neSpecc: I think it would be better to trigger `EditorMobileLayoutToggled` event instead of just "window resize":
- `isMobile` is not actually mobile screens, it just meant that editor now in a mobile-view
- users can use own window resize listeners, and it's not obvious why they should use ours. In case of `EditorMobileLayoutToggled` it became more clear and explicit.

src/components/ui/toolbox.ts
neSpecc: it should be placed in jsdoc

src/components/modules/toolbar/index.ts
neSpecc: Let's subscribe on that event separately in Block Settings and Toolbox

src/components/ui/toolbox.ts
neSpecc: it is assigned in a line above, so it could not be null. Seems like there is a TS problem somewhere that should be resolved

src/components/utils/popover/components/popover-header/popover-header.ts
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * Destroys the instance
+   */
+  public destroy(): void {
+    this.nodes.root.remove();
+    this.nodes.root = null;
+    this.nodes.backButton = null;
+    this.nodes.text = null;
+    this.listeners.removeAll();
+    this.listeners.destroy();
```
neSpecc: destroy includes removeAll internally
```suggestion
    this.listeners.destroy();
```

src/components/utils/popover/components/search-input/search-input.ts
neSpecc: That means that TS types declared wrongly. I'd suggest to get rid of `render()` method and create all elements in a constructor. So they will never be null

src/components/utils/popover/popover-abstract.ts
neSpecc: same, it could be resolved by removing `make()` method.

src/components/utils/popover/popover-abstract.ts
neSpecc: there should be return since if `clickedItem.toggle === true`, so check for `typeof clickedItem.toggle === 'string'` will never passed, that means that code below is not reachable. And its better to return here if it is not supposed to continue execution

src/components/utils/popover/popover-abstract.ts
neSpecc: why not just `showNestedItems()?`

src/components/utils/popover/popover-desktop.ts
```diff
@@ -0,0 +1,364 @@
...
+  }
+
+  /**
+   * Clears memory
+   */
+  public destroy(): void {
+    super.destroy();
+    this.flipper?.deactivate();
+    this.destroyNestedPopoverIfExists();
+    this.previouslyHoveredItem = null;
```
neSpecc:
```suggestion
    this.hide();
    super.destroy();
```

src/components/utils/popover/popover-mobile.ts
neSpecc: It would be better to add it now since it is a part of memory clearing of the new class

test/cypress/tests/utils/popover.cy.ts
```diff
@@ -257,4 +258,187 @@ describe('Popover', () => {
...
+
+    /** Check child item displayed */
+    cy.get('[data-cy=editorjs]')
+      .get('.ce-popover--nested .ce-popover__container')
+      .get('[data-item-name="nested-test-item"]')
+      .should('be.visible');
+  });
+
+
+  it('should correctly display children items and switch between states (mobile)', () => {
```
neSpecc: please describe a whole scenario and expected results in test case name

src/components/modules/toolbar/index.ts
```diff
@@ -599,5 +601,6 @@ export default class Toolbar extends Module<ToolbarNodes> {
     if (this.toolboxInstance) {
       this.toolboxInstance.destroy();
     }
+    this.listeners.destroy();
```
neSpecc: Why this line has been added? Seems like listeners are not used in this file

src/components/modules/ui.ts
```diff
@@ -423,10 +424,21 @@ export default class UI extends Module<UINodes> {
...
     this.checkIsMobile();
+
+    if (prevIsMobile !== this.isMobile) {
+      /**
+       * Dispatch global event
+       */
+      this.eventsDispatcher.emit(EditorMobileLayoutToggled, {
+        isEnabled: this.isMobile,
+      });
+    }
```
neSpecc: let's move it inside a `checkIsMobile()`?

src/components/ui/toolbox.ts
```diff
@@ -75,15 +79,21 @@ export default class Toolbox extends EventsDispatcher<ToolboxEventMap> {
    */
   public opened = false;
 
+  /**
+   * Listeners util instance
+   */
+  protected listeners: Listeners = new Listeners();
+
```
neSpecc, line 86: seems unused

src/components/utils/popover/popover-desktop.ts
```diff
@@ -0,0 +1,357 @@
...
+   */
+  public flipper: Flipper;
+
+  /**
+   * List of html elements inside custom content area that should be available for keyboard navigation
+   */
+  private customContentFlippableItems: HTMLElement[] | undefined;
+
+  /**
+   * Reference to nested popover if exists
```
neSpecc: describe when it can be undefined and null

src/components/utils/popover/popover-mobile.ts
```diff
@@ -0,0 +1,142 @@
...
+
+    /* Save state to history for proper navigation between nested and parent popovers */
+    this.history.push({ items: params.items });
+  }
+
+  /**
+   * Open popover
+   */
+  public show(): void {
+    this.nodes.overlay?.classList.remove(css.overlayHidden);
```
neSpecc:
```suggestion
    this.nodes.overlay!.classList.remove(css.overlayHidden);
```

or pass `PopoverNodes` as generic to `PopoverAbstract` to explicitly tell ts, that `PopoverMobile` has overlay

src/components/utils/popover/popover-desktop.ts
```diff
@@ -0,0 +1,357 @@
...
+  }
+
+  /**
+   * Clears memory
+   */
+  public destroy(): void {
+    super.destroy();
+    this.flipper.deactivate();
+    this.destroyNestedPopoverIfExists();
+    this.previouslyHoveredItem = null;
```
neSpecc:
```suggestion
    this.hide();
    super.destroy();
```

src/components/modules/ui.ts
TatianaFomina: `checkIsMobile()` is also called in `prepare()`. There will be extra `EditorMobileLayoutToggled` event firing

src/components/utils/popover/components/search-input/search-input.ts
```diff
@@ -67,13 +47,41 @@ export default class SearchInput {
...
+       * so Tab in the last block will focus this hidden input if this property is not set)
+       */
+      tabIndex: -1,
+    }) as HTMLInputElement;
+
+    this.wrapper.appendChild(iconWrapper);
+    this.wrapper.appendChild(this.input);
+
+    this.listeners.on(this.input, 'input', () => {
+      if (this.input === undefined) {
```
neSpecc: seems redundant

src/components/utils/popover/components/search-input/search-input.ts
```diff
@@ -94,7 +102,10 @@ export default class SearchInput {
    * Clears search query and results
    */
   public clear(): void {
-    this.input.value = '';
+    if (this.input !== undefined) {
```
neSpecc: seems redundant

neSpecc (approved):
add a Changeling please

PR #2650, opened 2024-03-09
fix(block-tunes): enter keydown problems

### LINKED ISSUE #2637: Remove block on SP with keyboard
Describe a bug.
SP size and still moving a block with Keyboard operation deletes the block.

Steps to reproduce:
1. Go to https://editorjs.io/
2. Click on "Play With Demo"
3. Set browser width to 360px (height can be anything)
4. Type "Hello" as Text
5. Type NewLine (enter)
6. Type "World" as Heading
7. Move "World" up one place

Expected behavior:
"World", NewLine, and "Hello" on the screen in that order.

Movie: 

https://github.com/codex-team/editor.js/assets/38370581/cbb43000-556d-464e-8216-f47511633974

Device: macbook air
Browser: chrome
OS: 14.2.1（23C71）

Editor.js version: unknown (maybe always latest)

neSpecc:
Right now we have several problems with Enter keydown.

# Unexpected new line on Enter press with selected block without caret

The problem can be reproduced when there is no caret set, but the block is selected. 

https://github.com/codex-team/editor.js/assets/3684889/ac3320ae-a4a4-46fd-9715-1ef0168f39dc

Since there is no caret in a Block, the Ui.ts module's `enterKeydown` is involved. We call `BlockManager.insert()` and `Caret.setToBlock()`, but we're not calling the `event.preventDefault()` to disable default "Enter" behaviour. There was no problem until the 2.29 because there was a delay between block adding and caret setting. This delay was removed by #2570 — now block addition follows by immediate focusing so Enter handler ends with caret in a new block and default behaviour (a linebreak) is happening.

### Solution 

Addition of `event.preventDefault()` after block insertion resolves the issue.

# Autofocus loosing after Block Tunes opening.

Block Tunes have autofocus on search field. But the caret disappears in a moment:

https://github.com/codex-team/editor.js/assets/3684889/c2ecae88-7663-4dc4-9f36-3eebec69e387

After the focusing of the search field, the `ui.ts@selectionChanged()` callback is called. It checks if `isCrossBlockSelectionStarted()` to remove all ranges. 

The problem was in the `isCrossBlockSelectionStarted()` it was treating the single selected block as a CBS. The `requestAnimationFrame` in input focusing was redundant and just hides the problem.

# Block removing while Enter press on Block Tunes 

It is related to the previous problem: after caret loosing, the `enterKeydown` `!Selection.isSelectionExists` leads us to the `BlockSelection.clearSelection(event);` where block was deleting.

https://github.com/codex-team/editor.js/assets/3684889/9881f006-3a2c-45a8-aeef-e0fcdfc23867

Fix #2637

### DIFF
--- a/test/cypress/support/commands.ts
+++ b/test/cypress/support/commands.ts
@@ -248,7 +248,7 @@ Cypress.Commands.add('keydown', {
    * but by default cypress' KeyboardEvent is not an instance of the native KeyboardEvent,
    * so real-world and Cypress behaviour were different.
    *
-   * To make it work we need to trigger Cypres event with "eventConstructor: 'KeyboardEvent'",
+   * To make it work we need to trigger Cypress event with "eventConstructor: 'KeyboardEvent'",
    *
    * @see https://github.com/cypress-io/cypress/issues/5650
    * @see https://github.com/cypress-io/cypress/pull/8305/files
--- a/test/cypress/support/commands.ts
+++ b/test/cypress/support/commands.ts
@@ -234,3 +234,28 @@ Cypress.Commands.add('getLineWrapPositions', {
 
   return cy.wrap(lineWraps);
 });
+
+/**
+ * Dispatches keydown event on subject
+ * Uses the correct KeyboardEvent object to make it work with our code (see below)
+ */
+Cypress.Commands.add('keydown', {
+  prevSubject: true,
+}, (subject, keyCode: number) => {
+  /**
+   * We use the "reason instanceof KeyboardEvent" statement in blockSelection.ts
+   * but by default cypress' KeyboardEvent is not an instance of the native KeyboardEvent
+   *
+   * To make it work we need to trigger Cypres event with "eventConstructor: 'KeyboardEvent'",
+   *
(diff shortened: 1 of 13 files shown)


gohabereg (approved):

test/cypress/support/commands.ts
```diff
@@ -234,3 +234,29 @@ Cypress.Commands.add('getLineWrapPositions', {
...
+ */
+Cypress.Commands.add('keydown', {
+  prevSubject: true,
+}, (subject, keyCode: number) => {
+  /**
+   * We use the "reason instanceof KeyboardEvent" statement in blockSelection.ts
+   * but by default cypress' KeyboardEvent is not an instance of the native KeyboardEvent,
+   * so real-world and Cypress behaviour were different.
+   *
+   * To make it work we need to trigger Cypres event with "eventConstructor: 'KeyboardEvent'",
```
ilyamore88:
```suggestion
   * To make it work we need to trigger Cypress event with "eventConstructor: 'KeyboardEvent'",
```

ilyamore88 (approved):

Issue #2613, opened 2024-02-12
An empty block is added in readOnly mode on top


bettysteger:
After the update to 2.29.0 in readonly mode there is a empty block added with an invisible (opacity: 0) placeholder text. The empty line is breaking the layout. 

I temporarily fixed it with CSS: 

```
.ce-paragraph[data-placeholder][contenteditable="false"]:empty {
  display: none;
}
```

Expected behavior:
No additional line on top should be rendered, like in v2.28.0 in readOnly mode.

Screenshots:
![image](https://github.com/codex-team/editor.js/assets/244475/b2667d5e-d6ac-4065-a446-737115f70f11)

Device, Browser, OS:
Macbook, Chrome, MacOS

Editor.js version: 2.29.0

neSpecc:
Thanks, @bettysteger , would you try to research that?

conradfr:
It seems to only appears if you initiate editorjs with `readOnly: true`, if you toggle to edit and then back to readonly, or if you initiate with `readOnly: false` and toggle it, the empty blocks does not come back.

bettysteger:
I temporarily fixed it via CSS:

```
// need this fix since update of editor.js to 2.29.0
// @see https://github.com/codex-team/editor.js/issues/2613
.codex-editor__redactor > .ce-block:first-child .ce-paragraph[data-placeholder][contenteditable="false"]:empty {
  display: none;
}
  ```

PR #2670, opened 2024-03-29
fix issue #2613

egg-:
Fix issue #2613 

Changes:
* Add exception code to prevent adding blocks when in read-only mode.

Please Checks:
* #2454


neSpecc (commented):
Looks good. Please, add a line to the changelog. And add a corresponded test case

PR #2564, opened 2023-12-16
Feature possibilities to merge blocks of different types

GuillaumeOnepilot:
For some context, I have a need for potential custom blocks to accept other different types of blocks to be merged in them (think Notion behavior), here is an example from Notion:

https://github.com/codex-team/editor.js/assets/97881811/faf0377b-6fc1-4db8-a1cc-d8a407f4eb07

My solution is to add a new static property which contains the block's type that can be merged inside the target block, I also added the type to the merge functions in case some blocks would want to make conditional merging.

Here is the result of my implementation (which I agree is a bit weird, in some sense because who would want to merge a paragraph into a title, but who knows): 

https://github.com/codex-team/editor.js/assets/97881811/ead89a79-7857-4606-9f6d-87bedfc92a0b

and with my specific use case (custom blocks): 

https://github.com/codex-team/editor.js/assets/97881811/257b465c-0044-45e4-8d27-bb40f54f1583

PR #2671, opened 2024-03-30
feat(merge): blocks of different types can be merged

neSpecc:
This PR based on https://github.com/codex-team/editor.js/pull/2564 made by @GuillaumeOnepilot

---

# Feature

Now it is possible to merge blocks of different types.

If blocks provides the `conversionConfig`, we use it for merging:

```
BlockA -> string -> BlockB
```

https://github.com/codex-team/editor.js/assets/3684889/6a037d8f-ca09-476e-ae64-6057709a8fc6

# Changes

- `requestAnimationFrame` removed from `BlockEvents@mergeBlocks`. It was added because Paragraph was having `RAF` in `merge()` method. It was removed by https://github.com/editor-js/paragraph/pull/86
- `normalize()` removed as well since it creates issues with caret restoring — nodes text length changes and caret is being set in wrong position. It's better to call `normalize()` at the tool's `merge()` method if needed.
- `mergeBlocks` and `areBlocksMergeable` logic improved:
  - Target block tool should have the `merge()` method (be `.mergeable`)
  - Merging blocks should be the same type
  - or Merging blocks should have the `conversionConfig`
- In case of using conversion:
  1. We convert `blockToMerge` data to string
  2. Sanitize string
  3. Convert that string to the target's block data 
  4. Call `targetBlock.merge()`  
- Sanitising added to the merging content.
- Cypress updated to fix CI which is broken on all last PRs

### DIFF
--- a/src/components/modules/blockManager.ts
+++ b/src/components/modules/blockManager.ts
@@ -682,32 +682,6 @@ export default class BlockManager extends Module {
   }
 
   /**
-   * Remove selection from all Blocks then highlight only Current Block
-   */
-  public highlightCurrentNode(): void {
-    /**
-     * Remove previous selected Block's state
-     */
-    this.clearFocused();
-
-    /**
-     * Mark current Block as selected
-     *
-     * @type {boolean}
-     */
-    this.currentBlock.focused = true;
-  }
-
-  /**
-   * Remove selection from all Blocks
-   */
-  public clearFocused(): void {
-    this.blocks.forEach((block) => {
-      block.focused = false;
-    });
-  }
-
-  /**
    * 1) Find first-level Block from passed child Node
    * 2) Mark it as current
    *
@@ -891,7 +865,6 @@ export default class BlockManager extends Module {
    */
   public dropPointer(): void {
     this.currentBlockIndex = -1;
-    this.clearFocused();
   }
 
   /**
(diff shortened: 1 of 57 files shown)


GuillaumeOnepilot (approved):
Thanks for the changes 👍

himself65 (approved):

src/components/modules/blockManager.ts
```diff
@@ -471,12 +471,40 @@ export default class BlockManager extends Module {
...
+      }
+
+      const [ cleanData ] = sanitizeBlocks([ blockToMergeDataRaw ], targetBlock.tool.sanitizeConfig);
+
+      blockToMergeData = cleanData;
+
+    /**
+     * 2) Blocks with different Tools if they provides conversionConfig
+     */
+    } else if (targetBlock.mergeable && isBlockConvertable(blockToMerge, 'export') && isBlockConvertable(targetBlock, 'import')) {
```
gohabereg, line 496: It seems `areBlocksMergable`  can be used here

gohabereg (approved):

src/components/modules/blockManager.ts
neSpecc: `areBlockMergeable` is checked before `mergeBlocks` call. I've written these 2 statements explicitly for clarity: first about same-tool merge, second about merging through the conversion

TatianaFomina (approved):

PR #2686, opened 2024-04-15
Fix/add tunes data in mutation events

GuillaumeOnepilot:
My usage of EditorJs rely mainly on the Mutation events to make granular actions on my backend, instead of pushing the whole lists of blocks. 

I'm currently working on an annotation feature which require the tunes data to be sent, but unfortunately the different events don't have the tunes data sent with them, which mean I have to change my logic (which would be complicated)

So I propose to add the tunes data inside the different mutations events.


neSpecc:
> I'm currently working on an annotation feature which require the tunes data to be sent

Could you provide some examples?

GuillaumeOnepilot:
So basically I want my tune to act a some kind of metadata on the blocks, but the events don't hold the tune's data and I don't use the saver api in my case.

https://github.com/codex-team/editor.js/assets/97881811/abcbf9ef-64c7-40d3-a63b-aed9f0b883b1

neSpecc:
Why not just call `await event.detail.target.save()`? Tunes data will be there

PR #2692, opened 2024-04-21
feat(block tunes): Conversion Menu in Block Tunes

TatianaFomina:
`Convert to` now also available in Block Tunes!

image

image

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 – `New` – Block Tunes now supports nesting items
 – `New` – Block Tunes now supports separator items
-– `New` – Convert to is now also available in Block Tunes
+– `New` – "Convert to" control is now also available in Block Tunes
 
 ### 2.30.0
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### 2.30.1
 
 – `New` – Block Tunes now supports nesting items
+– `New` – Block Tunes now supports separator items
 
 ### 2.30.0
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 – `New` – Block Tunes now supports nesting items
 – `New` – Block Tunes now supports separator items
+– `New` – Convert to is now also available in Block Tunes
 
 ### 2.30.0
 
--- a/src/components/block/index.ts
+++ b/src/components/block/index.ts
@@ -21,7 +21,7 @@ import BlockTune from '../tools/tune';
 import { BlockTuneData } from '../../../types/block-tunes/block-tune-data';
 import ToolsCollection from '../tools/collection';
 import EventsDispatcher from '../utils/events';
-import { TunesMenuConfigItem } from '../../../types/tools';
+import { TunesMenuConfig, TunesMenuConfigItem } from '../../../types/tools';
 import { isMutationBelongsToElement } from '../utils/mutations';
 import { EditorEventMap, FakeCursorAboutToBeToggled, FakeCursorHaveBeenSet, RedactorDomChanged } from '../events';
 import { RedactorDomChangedPayload } from '../events/RedactorDomChanged';
@@ -614,9 +614,9 @@ export default class Block extends EventsDispatcher<BlockEvents> {
    * Returns data to render in tunes menu.
    * Splits block tunes settings into 2 groups: popover items and custom html.
    */
-  public getTunes(): [PopoverItemParams[], HTMLElement] {
+  public getTunes(): [PopoverItemParams[], PopoverItemParams[], HTMLElement] {
     const customHtmlTunesContainer = document.createElement('div');
-    const tunesItems: TunesMenuConfigItem[] = [];
+    const commonTunesPopoverParams: TunesMenuConfigItem[] = [];
 
     /** Tool's tunes: may be defined as return value of optional renderSettings method */
     const tunesDefinedInTool = typeof this.toolInstance.renderSettings === 'function' ? this.toolInstance.renderSettings() : [];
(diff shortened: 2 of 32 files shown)


src/components/modules/toolbar/blockSettings.ts
```diff
@@ -108,6 +110,7 @@ export default class BlockSettings extends Module<BlockSettingsNodes> {
   public open(targetBlock: Block = this.Editor.BlockManager.currentBlock): void {
     this.opened = true;
 
+    // console.log(this.getConvertToItems());
```
gohabereg:
```suggestion
```

src/components/modules/toolbar/blockSettings.ts
```diff
@@ -197,6 +198,102 @@ export default class BlockSettings extends Module<BlockSettingsNodes> {
...
+
+    return items.map(tune => this.resolveTuneAliases(tune));
+  }
+
+  /**
+   * Returns list of conversion menu items available for the current block
+   *
+   * @param currentToolName - current block tool name
+   */
+  private getConvertToItems(currentToolName: string): TunesMenuConfigItem[] {
```
gohabereg: Looks like the result could be cached somewhere

TatianaFomina: Good point, fixed

src/components/block/index.ts
```diff
@@ -612,34 +611,60 @@ export default class Block extends EventsDispatcher<BlockEvents> {
...
 
-    [tunesDefinedInTool, commonTunes].flat().forEach(rendered => {
-      if ($.isElement(rendered)) {
-        customHtmlTunesContainer.appendChild(rendered);
-      } else if (Array.isArray(rendered)) {
-        tunesItems.push(...rendered);
-      } else {
-        tunesItems.push(rendered);
+    /** Separate custom html from Popover items params for common tunes */
+    commonTunes.forEach(rendered => {
```
neSpecc: This would be more straightforward

```suggestion
    commonTunes.forEach(tuneConfig => {
```

src/components/modules/toolbar/blockSettings.ts
```diff
@@ -197,6 +197,100 @@ export default class BlockSettings extends Module<BlockSettingsNodes> {
...
+
+    conversionEntries.forEach(([toolName, tool]) => {
+      const conversionConfig = tool.conversionConfig;
+
+      /**
+       * Skip tools without «import» rule specified
+       */
+      if (!conversionConfig || !conversionConfig.import) {
+        return;
+      }
```
neSpecc, line 259: can we use existing `isBlockConvertable()` here?

src/components/modules/toolbar/blockSettings.ts
```diff
@@ -197,6 +197,100 @@ export default class BlockSettings extends Module<BlockSettingsNodes> {
...
+      tool.toolbox?.forEach((toolboxItem) => {
+        /**
+         * Skip tools that don't pass 'toolbox' property
+         */
+        if (_.isEmpty(toolboxItem) || !toolboxItem.icon) {
+          return;
+        }
+
+        resultItems.push({
+          type: 'default',
```
neSpecc: can we omit a type for default items?

src/components/modules/toolbar/blockSettings.ts
TatianaFomina: I don't know, I copied the logic from module [conversion.ts](https://github.com/codex-team/editor.js/blob/next/src/components/modules/toolbar/conversion.ts#L213)

src/components/modules/toolbar/blockSettings.ts
TatianaFomina: it will work, but ts will show error. So, better not to omit

src/components/modules/toolbar/blockSettings.ts
neSpecc: You should at least try. This method added recently. All such statements should be utilized

src/components/modules/toolbar/blockSettings.ts
neSpecc: Can you tune ts typings for allowing that?

src/components/modules/toolbar/blockSettings.ts
```diff
@@ -197,6 +197,121 @@ export default class BlockSettings extends Module<BlockSettingsNodes> {
...
+
+    return items.map(tune => this.resolveTuneAliases(tune));
+  }
+
+  /**
+   * Returns list of all available conversion menu items
+   *
+   * @param currentBlock - block we are about to open block tunes for
+   */
+  private async getConvertToItems(currentBlock: Block): Promise<PopoverItemDefaultParams[]> {
```
gohabereg, line 244: It seems the comment about cache is relevant again now

docs/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 – `New` – Block Tunes now supports nesting items
 – `New` – Block Tunes now supports separator items
+– `New` – Convert to is now also available in Block Tunes
```
```suggestion
– `New` – "Convert to" control is now also available in Block Tunes
```

src/components/modules/toolbar/blockSettings.ts
TatianaFomina: at this phase we do not have a block to check if it is convertable or not. We operate with tools, not blocks

neSpecc: you have this.Editor.Tools.blockTools above

src/components/modules/toolbar/blockSettings.ts
TatianaFomina: can not cache. "Convert to" items list is different every time. For each block there will be different subset of tools it can be converted to. It depends on current block type and data

src/components/modules/toolbar/blockSettings.ts
TatianaFomina: Those are block tools, not blocks
image

src/components/modules/toolbar/blockSettings.ts
```diff
@@ -197,6 +197,121 @@ export default class BlockSettings extends Module<BlockSettingsNodes> {
...
+           * toolbox item with the same data to the resulting array. This helps exclude duplicates
+           */
+          const hasSameData = Object.entries(toolboxItem.data).some((([propName, propValue]) => {
+            return blockData[propName] && _.equals(blockData[propName], propValue);
+          }));
+
+          shouldSkip = hasSameData;
+        } else {
+          shouldSkip = toolName === currentBlock.name;
+        }
```
neSpecc: 1) can we extract this logic to some util?
2) we need to cover it with corresponded tests

src/components/modules/toolbar/blockSettings.ts
TatianaFomina: done

src/components/modules/toolbar/blockSettings.ts
TatianaFomina: moved to util `isSameBlockData`

neSpecc (approved):

gohabereg (approved):

PR #2703, opened 2024-05-01
propagate click event to popover item activation callback

mauritskorse:
For creating modals that can attach to the popover menu, it is useful to have the event data being propagated to the onActivate() callback. The interface already had the event implemented:
https://github.com/codex-team/editor.js/blob/238c90901681614a91e3b72ecbf406a7c2a23f05/src/components/utils/popover/components/popover-item/popover-item.types.ts#L91

But the methods in after the event listener did not pass the event data on yet.

### DIFF
--- a/src/components/utils/popover/components/popover-item/popover-item-default/popover-item-default.ts
+++ b/src/components/utils/popover/components/popover-item/popover-item-default/popover-item-default.ts
@@ -103,15 +103,16 @@ export class PopoverItemDefault extends PopoverItem {
 
   /**
    * Called on popover item click
+   * @param event - click event
    */
-  public handleClick(): void {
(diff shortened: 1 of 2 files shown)


TatianaFomina:
Hi! Can you please provide the use case for which you need to have an `event` data in `onActivate` handler?

TatianaFomina:
Also, please add a test and update changelog

mauritskorse:
> Hi! Can you please provide the use case for which you need to have an `event` data in `onActivate` handler?

Use case is to get the xy position of the menu item such that I can position a modal next to the popover menu item when clicked.
Another option would be to be able to add custom html in the submenu that is now available in the rc for v2.30.

PR #2711, opened 2024-05-04
feat(popover): Add hint support

TatianaFomina:
It becomes possible to display a hint near popover item. 
Only works in desktop popover. 

image

This pull request is required for further progress on making Inline Toolbar render via Popover like Toolbar and Block Tunes

image


src/components/utils/popover/components/popover-item/popover-item-default/popover-item-default.ts
```diff
@@ -86,11 +93,14 @@ export class PopoverItemDefault extends PopoverItem {
...
    * @param params - popover item construction params
+   * @param renderParams - popover item render params.
+   * The parameters that are not set by user via popover api but rather depend on technical implementation
    */
-  constructor(params: PopoverItemDefaultParams) {
+  constructor(params: PopoverItemDefaultParams, renderParams?: PopoverItemRenderParams) {
     super();
 
     this.params = params;
+    this.hintPosition = renderParams?.hint?.position || 'right';
```
neSpecc: seems redundant. Why not just use `renderParams.hint.position`?

src/components/utils/popover/components/popover-item/popover-item.types.ts
```diff
@@ -152,3 +183,19 @@ export type PopoverItemParams =
   PopoverItemSeparatorParams |
   PopoverItemHtmlParams;
 
+
+/**
+ * Popover item render params.
```
neSpecc, line 168:
```suggestion
```

src/components/utils/popover/components/popover-item/popover-item-default/popover-item-default.ts
TatianaFomina: I do not want to let users set the position manually

TatianaFomina: so, seem like we need to split parameters that can be altered by users and the ones that are 'technical'

gohabereg (approved):

ilyamore88 (approved):

src/components/utils/popover/components/popover-item/popover-item.types.ts
```diff
@@ -152,3 +183,25 @@ export type PopoverItemParams =
...
+   */
+  hint?: {
+    /**
+     * Hint position relative to the item
+     */
+    position?: HintPosition;
+
+    /**
+     * If false, hint will not be rendered.
+     * True by default
```
neSpecc:
```suggestion
     * If false, hint will not be rendered.
     * True by default
     * Used to disable hints on mobile popover
```

TatianaFomina: fixed

neSpecc (approved):

PR #2720, opened 2024-05-17
feat(blocks-api): blocks.update() now can update `tunes` as well as `data`

w99910:
AFAIK, when you update block using `editor.blocks.update` method, only `data` attribute is merged and updated. I believe `tunes` data should be updated if provided.


neSpecc (commented):
Please, add a corresponded testcase

w99910:
> Please, add a corresponded testcase

I have added a test. In order to prevent breaking change, I decided to add new parameter called `tunes` to `update` method of `BlockManager` and `BlocksAPI`.

src/components/modules/api/blocks.ts
```diff
@@ -303,16 +303,17 @@ export default class BlocksAPI extends Module {
    *
    * @param id - id of the block to update
    * @param data - the new data
+   * @param tunes
    */
-  public update = async (id: string, data: Partial<BlockToolData>): Promise<BlockAPIInterface> => {
+  public update = async (id: string, data: Partial<BlockToolData>, tunes: object = null): Promise<BlockAPIInterface> => {
```
neSpecc: its better to make `tunes` optional instead of nullable

neSpecc: and use `{[name: string]: BlockTuneData}` as a type instead of `object`.

src/components/modules/api/blocks.ts
```diff
@@ -303,16 +303,17 @@ export default class BlocksAPI extends Module {
    *
    * @param id - id of the block to update
    * @param data - the new data
+   * @param tunes
```
neSpecc: fill docs please

test/cypress/tests/should-update-tunes-settings.cy.ts
neSpecc, line 1: move the test to the `test/cypress/tests/api/blocks.cy.ts` under `.update()` suite

test/cypress/tests/should-update-tunes-settings.cy.ts
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  /**
+   * CSS selectors used in Tune
+   */
+  public static get CSS(): object {
+    return {
+      toggler: 'cdx-text-variant__toggler',
+    };
+  }
```
neSpecc: seems redundant

yarn.lock
neSpecc, line 1: unexpected change

package-lock.json
neSpecc, line 1: unexpected change

neSpecc (changes requested):

src/components/modules/api/blocks.ts
w99910: Please review the updated commits.

src/components/modules/api/blocks.ts
```diff
@@ -303,16 +304,17 @@ export default class BlocksAPI extends Module {
    *
    * @param id - id of the block to update
    * @param data - the new data
+   * @param tunes - (optional) tune data
    */
-  public update = async (id: string, data: Partial<BlockToolData>): Promise<BlockAPIInterface> => {
+  public update = async (id: string, data: Partial<BlockToolData>, tunes?: {[name: string]: BlockTuneData}): Promise<BlockAPIInterface> => {
```
neSpecc: What if I want to update only tunes? I think the data should be optional as well. 
One of "tune" or "data" should be provided.

neSpecc (commented):
Almost ready to approve, please check the last comment and add a changelog

src/components/modules/api/blocks.ts
w99910: @neSpecc I have updated the code. I think it should just return provided block if neither new block data nor tunes data is provided in order to skip redundant and unnecessary update. What do you think about it?

neSpecc (approved):

TatianaFomina (approved):

PR #2741, opened 2024-06-15
fix(caret): trailing spaces handling

### LINKED ISSUE #1179: [Bug] isAtStart throws an error when deleting content.
Sometimes when you delete content, a bug appears:

```
TypeError
Cannot read property 'length' of undefined
```

Error source: 
```
{snip} tion(t){var e=d.default.isLineBreakTag(t),n=1===t.children.length&&d.default.isLineBreakTag(t.children[0]),o=e||n;return d.default.isEmpty(t {snip}
```

Which is the minified version of:

```javascript
  const leftSiblings = this.getHigherLevelSiblings(focusNode as HTMLElement, 'left');
      const nothingAtLeft = leftSiblings.every((node) => {
        /**
         * Workaround case when block starts with several <br>'s (created by SHIFT+ENTER)
         *
         * @see https://github.com/codex-team/editor.js/issues/726
         * We need to allow to delete such linebreaks, so in this case caret IS NOT AT START
         */
        const regularLineBreak = $.isLineBreakTag(node);
        /**
         * Workaround SHIFT+ENTER in Safari, that creates <div><br></div> instead of <br>
         */
        const lineBreakInSafari = node.children.length === 1 && $.isLineBreakTag(node.children[0] as HTMLElement);
        const isLineBreak = regularLineBreak || lineBreakInSafari;

        return $.isEmpty(node) && !isLineBreak;
      });
```

Last action of the user:

```
div.ce-block__content > div.ce-paragraph.cdx-block
```

Expected behavior:

The deletion should work.

Device, Browser, OS: Chrome 83.0.4103, Mac OS X 10.14.6

Editor.js version: 

```
    "@editorjs/checklist": "^1.1.0",
    "@editorjs/editorjs": "^2.17.0",
    "@editorjs/embed": "^2.3.1",
    "@editorjs/header": "^2.4.1",
    "@editorjs/image": "^2.3.4",
    "@editorjs/link": "^2.1.3",
    "@editorjs/list": "^1.4.0",
    "@editorjs/marker": "^1.2.2",
    "@editorjs/quote": "^2.3.0",
    "@editorjs/warning": "^1.1.1",
```

### LINKED ISSUE #1382: [Bug] Wrong caret moving and block merging when there are spaces at block edge

### LINKED ISSUE #2429: Deleting Characters in New Blocks
Description:
I'm not sure if anyone has raised this bug before, so I wanted to report it. When creating a new block and adding a few spaces before a word, if the cursor is positioned at the start of the word and I press the Delete button (or <- Back button), the entire block is deleted instead of just one character.

Steps to Reproduce:
(I used your demo website to reproduce the bug, so you should be able to replicate it)

1.Go to your demo website: https://editorjs.io/
2.Enter some random text.
3.Place your cursor at the center of the text or somewhere else and press Enter (to create a new block).
4.Add some spaces before that word and click the Delete button (or <- Back button), and you will see the bug occur.
Expected Behavior:
Only one character should be deleted, not the entire block.

Browser:
Chrome, OS: Windows 11
Editor.js version: None

### LINKED ISSUE #2521: I cannot use backspace to remove whitespace character at the start of a block
Steps to reproduce:
1. Use API to insert a block, which starts with a whitespace - space or tab (\t).
2. Focus right after the whitespace and try to remove it by using Backspace - you can't

Expected behavior:
You should be able to remove the whitespace, just like it works for letters, numbers, special characters...

Device, Browser, OS:
Chrome 118, Windows

Editor.js version:
2.28.2

No plugins are used

Hello, I started using editor.js a couple of weeks ago and here's a problem I've recently stumbled upon:
when you create a paragraph block, which starts with a whitespace, you cannot remove the whitespace with Backspace button.
Here's a reproduction: https://codesandbox.io/s/quizzical-hooks-4nctln?file=/index.html

As you can see, I create a paragraph block with `blocks.insert("paragraph", { text: "\tabc" });` and then I can't remove the tab at the beginning of the string with Backspace. I can remove it with the Delete key, but I want Backspace to work as well. Is this a known issue?

neSpecc:
## Problem 

If block starts with several spaces, they can't be deleted. Caret jumps to the previous block.

Resolves #2429
Resolves #1382
Resolves #1179
Resolves #2521

## Cause

There are two whitespace handling in HTML 
1. Invisible spaces — all regular trailing spaces, tabs, etc
2. Visible spaces — `&nbsp;`

Our `Caret.isAtStart` method covers only case №1 and treats all whitespaces like "empty". 

## Solution

I've rewrote `isAtStart()` and `isAtEnd()` methods.

1. They were moved from the Caret Module to the "caret.ts" util.
2. Their dependency from `this.BlockManager` has been removed. Now methods accept `input` as argument. (so we'll be able to expose them to the api or some package)
3. Their implementation has been simplified. Now we create a range from the block start (or end in case of `isAtEndOfInput()`) to the caret position, clone its content to the temporary div and check its textContent
4. Now we handle collapsed and visible whitespaced correctly. Util `isCollapsedWhitespaces()` is used for it.
5. Tests for Backspace/Delete/Left/Right when caret at the start/end added
6. Tests for arrow-navigation through the Delimiter added

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -34,5 +34,8 @@
     "unknown": true,
     "requestAnimationFrame": true,
     "navigator": true
+  },
+  "rules": {
+    "jsdoc/require-returns-type": "off"
   }
 }
--- a/src/components/modules/caret.ts
+++ b/src/components/modules/caret.ts
@@ -1,19 +1,14 @@
-/**
- * @class Caret
- * @classdesc Contains methods for working Caret
- *
- * Uses Range methods to manipulate with caret
- * @module Caret
- * @version 2.0.0
- */
-
 import Selection from '../selection';
 import Module from '../__module';
 import Block from '../block';
-import $ from '../dom';
+import * as caretUtils from '../utils/caret';
+import $  from '../dom';
 
 /**
- * @typedef {Caret} Caret
+ * Caret
+ * Contains methods for working Caret
+ *
+ * @todo get rid of this module and separate it for utility functions
  */
 export default class Caret extends Module {
   /**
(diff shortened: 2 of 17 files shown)


.eslintrc
```diff
@@ -34,5 +34,8 @@
     "unknown": true,
     "requestAnimationFrame": true,
     "navigator": true
+  },
+  "rules": {
+    "jsdoc/require-returns-type": "off"
```
gohabereg, line 39: 👍

src/components/modules/caret.ts
```diff
@@ -428,11 +233,23 @@ export default class Caret extends Module {
...
     const { currentBlock, nextBlock } = BlockManager;
-    const { nextInput } = currentBlock;
-    const isAtEnd = this.isAtEnd;
+
+    if (currentBlock === undefined) {
+      return false;
+    }
+
+    const { nextInput, currentInput } = currentBlock;
+    const isAtEnd = currentInput !== undefined ? caretUtils.isAtEndOfInput(currentInput) : undefined;
```
gohabereg: Similar expression is used several times, maybe worth to extract it to a function. Or allow caret utils to accept undefined

gohabereg (approved):
Looks good overall, just a minor comment

src/components/modules/caret.ts
neSpecc: I believe that utility methods like `isAtStartOfInput()` should not handle case when input is `undefined `. Module should explicitly handle the cases when block has no inputs or there is no `currentBlock` at all since it is a part of business logic.

src/components/utils/caret.ts
```diff
@@ -0,0 +1,188 @@
...
+ *   - if offset is 0, caret is at the start
+ *  Contenteditable:
+ *   - caret at the first text node and offset is 0 — caret is at the start
+ *   - caret not at the first text node — we need to check left siblings for emptiness
+ *   - caret offset > 0, but all left part is visible (nbsp) — caret is not at the start
+ *   - caret offset > 0, but all left part is invisible (whitespaces) — caret is at the start
+ *
+ * @param input - input where caret should be checked
+ */
+export function isAtStartOfInput(input: HTMLElement): boolean {
```
TatianaFomina: This function name is a bit confusing, hard to understand what exactly is 'at start of input'. 
I suggest calling it 'isCaretAtStartOfInput'

src/components/utils/caret.ts
```diff
@@ -0,0 +1,188 @@
...
+ * - if offset is equal to value length, caret is at the end
+ * Contenteditable:
+ * - caret at the last text node and offset is equal to text length — caret is at the end
+ * - caret not at the last text node — we need to check right siblings for emptiness
+ * - caret offset < text length, but all right part is visible (nbsp) — caret is at the end
+ * - caret offset < text length, but all right part is invisible (whitespaces) — caret is at the end
+ *
+ * @param input - input where caret should be checked
+ */
+export function isAtEndOfInput(input: HTMLElement): boolean {
```
TatianaFomina: Same

TatianaFomina (approved):

Issue #1970, opened 2022-02-13
[Bug] can't paste links on mobile


orbachar:
Paste link doesn't work on mobile, on desktop it works as expected.
See attached video

Device, Browser, OS: chrome on Android

Editor.js version: 2.23.2

# #

https://user-images.githubusercontent.com/57751360/153760931-3c751348-28e4-43ba-bfd6-9b83b94a6f09.mp4

robonetphy:
Hi @orbachar, I am unable to reproduce the problem as shown in the video below:
# #

https://user-images.githubusercontent.com/23169768/157361849-1e1c9c9e-4368-4cf3-8d02-5b7f79b8f03f.mp4

If it still occurs on your phone then pls provide us with the chrome version.
Happy editing...

orbachar:
cant reproduce either, thanks!

PR #2746, opened 2024-06-27
Improve link creating on mobile devices

bezumkin:
I suggest you to add new attribute for the link input to make it work inside forms with multiple fields.

Right now when you creating a link on Android device, you cannot confirm it, because you will be switched to the next text input of the form. But if we will add `enterKeyHint` attribute, everything works as expected.

Before the change:

https://github.com/codex-team/editor.js/assets/1257284/0dddbf57-23da-42a5-bb68-1e291a275364

After the change

https://github.com/codex-team/editor.js/assets/1257284/af0b65e0-65ff-4ba6-a4c0-4e79804f1204

P.S. I think [this issue](https://github.com/codex-team/editor.js/issues/1970) is related but you can't reproduce it on demo site because you have no next inputs in the form.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -28,6 +28,7 @@
 – `Fix` — Deleting whitespaces at the start/end of the block
 – `Improvement` — *Types* — `BlockToolConstructorOptions` type improved, `block` and `config` are not optional anymore
 - `Improvement` - The Plus button and Block Tunes toggler are now better aligned with large line-height blocks, such as Headings
+- `Improvement` — Creating links on Android devices: now the mobile keyboard will have an "Enter" key for accepting the inserted link.
 
 ### 2.29.1
 
--- a/src/components/inline-tools/inline-tool-link.ts
+++ b/src/components/inline-tools/inline-tool-link.ts
@@ -134,6 +134,7 @@ export default class LinkInlineTool implements InlineTool {
   public renderActions(): HTMLElement {
     this.nodes.input = document.createElement('input') as HTMLInputElement;
     this.nodes.input.placeholder = this.i18n.t('Add a link');
+    this.nodes.input.enterKeyHint = 'done';
     this.nodes.input.classList.add(this.CSS.input);
     this.nodes.input.addEventListener('keydown', (event: KeyboardEvent) => {
       if (event.keyCode === this.ENTER_KEY) {


neSpecc (commented):
Please, add a note to the changelog

bezumkin:
@neSpecc Done!

docs/CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 – `New` – *Menu Config* – Default and HTML items now support hints
 – `Fix` — Deleting whitespaces at the start/end of the block
 – `Improvement` — *Types* — `BlockToolConstructorOptions` type improved, `block` and `config` are not optional anymore
+- `Improvement` — Link creating on mobile devices
```
```suggestion
- `Improvement` — Creating links on Android devices: now the caret will be positioned after the selected fragment when the Done key is pressed
```

bezumkin: @neSpecc Are you sure about this readme entry?

As far as I understand, we are just adding a [hint to the virtual keyboard](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/enterkeyhint) on how to handle the "Next" button. By default, this is moving the input to the next field of the form, but with my change we simply ask you to approve the input.

Without this change, it will not be the `Enter` key, but the key of the `Next Field` - you can see this in my videos.
Screenshot 2024-07-01 at 00 20 34 Screenshot 2024-07-01 at 00 20 45

We don't move the cursor, we don't do anything - we just explain what we want from the virtual keyboard with this hint. And we can't be sure how the client device will handle this.

neSpecc: The changelog should explain end-user experience, which is about the new caret behavior, as I saw on your video. Maybe replace "Enter is pressed" with "Done is pressed"?

bezumkin: @neSpecc You are right!

How about this line?

> Add type hint `done` on creating links to make it work correctly on mobile devices.

neSpecc: It's a kind of technical detail, but 'correctly' is not explained

bezumkin: @neSpecc New version:

> Add type hint `done` for the link input to prevent switching to the next form field on mobile devices.

Or maybe you will just enter any description you like? I just want to have this change in the next version =)

neSpecc:
```
Creating links on Android devices: now the mobile keyboard will have an "Enter" key for accepting the inserted link.
```

bezumkin: @neSpecc Done!

neSpecc (approved):

neSpecc:
Update a branch, please

bezumkin:
@neSpecc Done!

TatianaFomina (approved):

PR #2750, opened 2024-06-29
Solve the problem that / cannot evoke toolbox in Chinese input method…

chf-gh:
1. Solve the problem that `/` cannot evoke toolbox in Chinese input method. `the symbol / : event.key=Process, event.code=Slash, event.keyCode=229`
2. Solve the problem that `/` is automatically entered in searchInput after Chinese input method evokes toolbox

### DIFF
--- a/src/components/modules/blockEvents.ts
+++ b/src/components/modules/blockEvents.ts
@@ -59,7 +59,7 @@ export default class BlockEvents extends Module {
      *
      * @todo probably using "beforeInput" event would be better here
      */
-    if (event.key === '/' && !event.ctrlKey && !event.metaKey) {
+    if (event.code === 'Slash' && !event.ctrlKey && !event.metaKey) {
       this.slashPressed(event);
     }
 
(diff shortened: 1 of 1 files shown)


src/components/modules/blockEvents.ts
```diff
@@ -252,14 +252,9 @@ export default class BlockEvents extends Module {
...
-     * The Toolbox will be opened with immediate focus on the Search input,
-     * and '/' will be added in the search input by default — we need to prevent it and add '/' manually
-     */
-    event.preventDefault();
-    this.Editor.Caret.insertContentAtCaretPosition('/');
-
-    this.activateToolbox();
+    setTimeout(() => {
+      this.activateToolbox();
+    }, 0);
```
neSpecc, line 257: 1) this change looks unclear for me. Previous solution looks more explicit. Why it is not working?

2) We're trying to get rid of `setTimeout` in a code since it leads of problems of caret behavior which are hard to maintain.

chf-gh: The reason the toolbox cannot be invoked with the Chinese input method / is that the Chinese / corresponds to `event.key = Process` instead of `/`, so `event.code === 'Slash'` is used to determine if / is pressed. Using `setTimeout` to invoke the toolbox is mainly to asynchronously invoke the toolbox because the original code's `event.preventDefault()` cannot prevent the / input in Chinese mode, causing / to be automatically entered in the `SearchInput`. I don't have any ideas for better asynchronous calls yet.

neSpecc: > the original code's `event.preventDefault()` cannot prevent the / input in Chinese mode, causing / to be automatically entered in the `SearchInput`. 

So how it prevented now?

PR #2758, opened 2024-07-03
feat(ui): placeholders updated

neSpecc:
## Problem

1. Placeholders are being hidden on focus. It's not the best UX.
2. `autofocus` ruins `placeholder` option because of 1.
3. There are separated placeholder implementations in Tools CSS code.
4. Current implementation relies on CSS `:empty` state, but the is often a case when input has just `<br>` tag. Placeholder won't be shown in this case.

## Solution

1. Placeholder implementation moved to editor core. Now anyone can add `data-placeholder` to the contenteditable element inside editor and it will work.
2. Placeholders won't be hidden on focus
3. Now we're watching all inputs and mark them as empty via the `data-empty="true"` attribute if they are really empty.
4. We also support `data-placeholder-active` attribute to show placeholder only for current block. It is used for Paragraph tool — we don't want to show placeholders to each empty paragraphs on a page, only for the current.

Also, paragraph tool has a hack for showing `data-placeholder-active` event for not-current block in case if it is the only paragraph in empty editor. It is needed to make `placeholder` config option work when `autofocus` is false. See https://github.com/editor-js/paragraph/pull/89

https://github.com/codex-team/editor.js/assets/3684889/d9e09635-b09b-43c7-ab10-b1d9c2048a7e

### DIFF
--- a/src/components/block/index.ts
+++ b/src/components/block/index.ts
@@ -10,7 +10,7 @@ import {
 } from '../../../types';
 
 import { SavedData } from '../../../types/data-formats';
-import $ from '../dom';
+import $, { toggleEmptyMark } from '../dom';
 import * as _ from '../utils';
 import ApiModules from '../modules/api';
 import BlockAPI from './api';
@@ -255,6 +255,12 @@ export default class Block extends EventsDispatcher<BlockEvents> {
        * so we need to track focus events to update current input and clear cache.
        */
       this.addInputEvents();
+
+      /**
+       * We mark inputs with [data-empty] attribute
+       * It can be useful for developers, for example for correct placeholder behavior and
+       */
+      this.toggleInputsEmptyMark();
     });
   }
 
@@ -931,6 +937,11 @@ export default class Block extends EventsDispatcher<BlockEvents> {
      */
     this.updateCurrentInput();
 
+    /**
+     * We mark inputs with 'data-empty' attribute, so new inputs should be marked as well
+     */
+    this.toggleInputsEmptyMark();
+
     this.call(BlockToolAPI.UPDATED);
 
     /**
@@ -992,5 +1003,9 @@ export default class Block extends EventsDispatcher<BlockEvents> {
    */
   private dropInputsCache(): void {
     this.cachedInputs = [];
+  }
+
+  private toggleInputsEmptyMark(): void {
+    this.inputs.forEach(toggleEmptyMark);
   }
 }
--- a/src/components/block/index.ts
+++ b/src/components/block/index.ts
@@ -258,7 +258,7 @@ export default class Block extends EventsDispatcher<BlockEvents> {
 
       /**
        * We mark inputs with [data-empty] attribute
-       * It can be useful for developers, for example for correct placeholder behavior and
+       * It can be useful for developers, for example for correct placeholder behavior
        */
       this.toggleInputsEmptyMark();
     });
--- a/src/components/block/index.ts
+++ b/src/components/block/index.ts
@@ -1005,6 +1005,9 @@ export default class Block extends EventsDispatcher<BlockEvents> {
     this.cachedInputs = [];
   }
 
+  /**
+   * Mark inputs with 'data-empty' attribute with the empty state
+   */
   private toggleInputsEmptyMark(): void {
     this.inputs.forEach(toggleEmptyMark);
   }
--- a/src/components/block/index.ts
+++ b/src/components/block/index.ts
@@ -184,11 +184,6 @@ export default class Block extends EventsDispatcher<BlockEvents> {
   private unavailableTunesData: { [name: string]: BlockTuneData } = {};
 
   /**
-   * Editor`s API module
-   */
-  private readonly api: ApiModules;
-
-  /**
    * Focused input index
    *
    * @type {number}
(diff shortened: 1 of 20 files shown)


src/components/block/index.ts
```diff
@@ -262,6 +255,12 @@ export default class Block extends EventsDispatcher<BlockEvents> {
        * so we need to track focus events to update current input and clear cache.
        */
       this.addInputEvents();
+
+      /**
+       * We mark inputs with [data-empty] attribute
+       * It can be useful for developers, for example for correct placeholder behavior and
```
TatianaFomina: and...?

src/components/modules/blockManager.ts
```diff
@@ -873,7 +873,7 @@ export default class BlockManager extends Module {
    * Sets current Block Index -1 which means unknown
    * and clear highlights
    */
-  public dropPointer(): void {
+  public unsetCurrentBlock(): void {
```
TatianaFomina, line 876: maybe `clearCurrentBlock()`?

neSpecc: "clear" means that block will became empty, but we're not changing the block itself

TatianaFomina (approved):

test/cypress/tests/ui/DataEmpty.cy.ts
```diff
@@ -0,0 +1,71 @@
+import { createEditorWithTextBlocks } from '../../support/utils/createEditorWithTextBlocks';
+
+describe('inputs [data-empty] mark', function () {
+  it('should be added to the editor on initialization', function () {
```
ilyamore88:
```suggestion
  it('should be added to blocks on the editor initialization', function () {
```

ilyamore88: The test case name sounds like this attribute should be added to the editor container.

test/cypress/tests/ui/DataEmpty.cy.ts
```diff
@@ -0,0 +1,71 @@
...
+      .first()
+      .should('have.attr', 'data-empty', 'false');
+
+    cy.get('[data-cy=editorjs]')
+      .find('.ce-paragraph')
+      .last()
+      .should('have.attr', 'data-empty', 'true');
+  });
+
+  it('should be added to the block on input', function () {
```
ilyamore88: Your test case checks that this attr should be toggled on input. Would be great to add a test case that this attr will be set to true if this block is empty on input.

test/cypress/tests/ui/DataEmpty.cy.ts
```diff
@@ -0,0 +1,71 @@
...
+      .last()
+      .type('Some text');
+
+    cy.get('[data-cy=editorjs]')
+      .find('.ce-paragraph')
+      .last()
+      .should('have.attr', 'data-empty', 'false');
+  });
+
+  it('should be added to the block on focus', function () {
```
ilyamore88: It's good to cover this case, but if this behaviour is broken, this case will pass because of adding this attr by default on initialisation. I don't know how to test it in the right way.

test/cypress/tests/ui/DataEmpty.cy.ts
neSpecc: fixed

test/cypress/tests/ui/DataEmpty.cy.ts
neSpecc: added `should be added as "true" to the input on chars removal`

test/cypress/tests/ui/DataEmpty.cy.ts
neSpecc: removed this test

test/cypress/tests/ui/Placeholders.cy.ts
```diff
@@ -0,0 +1,72 @@
...
+    cy.createEditor({
+      placeholder: PLACEHOLDER_TEXT,
+      autofocus: true,
+    });
+
+    cy.get('[data-cy=editorjs]')
+      .find('.ce-paragraph')
+      .getPseudoElementContent('::before')
+      .should('eq', PLACEHOLDER_TEXT);
+  });
```
ilyamore88, line 36: So in both cases when I pass a placeholder, it will be shown. It doesn't really depend on the autofocus param. The second case is redundant then.

test/cypress/tests/ui/Placeholders.cy.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  it('should be shown event if input is focused', function () {
```
ilyamore88, line 38:
```suggestion
  it('should be shown even if input is focused', function () {
```

test/cypress/tests/ui/Placeholders.cy.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    cy.get('[data-cy=editorjs]')
+      .find('.ce-paragraph')
+      .click()
+      .as('firstBlock')
+      .getPseudoElementContent('::before')
+      .should('eq', PLACEHOLDER_TEXT);
+  });
+
+  it('should be shown event when user removes all text by cmd+a and delete', function () {
```
ilyamore88, line 51:
```suggestion
  it('should be shown even when user removes all text by cmd+a and delete', function () {
```

ilyamore88: Should be shown in an empty element when a user removes all content.

test/cypress/tests/ui/Placeholders.cy.ts
neSpecc: in previous version `autofocus` was leading placeholder to disappear, so I've added this test.

PR #2779, opened 2024-07-11
fix(popover): i18n in nested popover

TatianaFomina:
telegram-cloud-photo-size-2-5224484054232062515-x

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### 2.30.3
+
+- `Fix` – I18n in nested popover
+
 ### 2.30.2
 
 - `Fix` – The onChange callback won't be fired when editor is initialized in the Read-Only mode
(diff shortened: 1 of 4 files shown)


neSpecc (approved):

lucklyTime:
@TatianaFomina Hello, I would like to ask why the language configuration of List, Text, and Search in the nested pop-up window does not take effect. I referred to the configuration in https://github.com/codex-team/editor.js/blob/next/src/components/i18n/locales/en/messages.json, but did not get the desired result. Thank you, looking forward to your reply😊
```
i18n: {
  "messages": {
    "ui": {
      "popover": {
        "Filter": "Фильтр",
      }
    },
    "toolNames": {
      "Heading": "Заголовок",
      "Text": "Абзац",
      "List": "Список",
      "Image": "Изображение",
      "Table": "Таблица"
    }
  }
}
```
image
image

PR #2826, opened 2024-09-09
fixed css stoke overwrite issue

### LINKED ISSUE #2294: CSS overriding the style of block tool SVG component
There is a css rule in file [ui.css](https://github.com/codex-team/editor.js/blob/7bd5c53ab27defac388666d80e75606cd15ab07e/src/styles/ui.css#L107), which goes like this...

`path {
stroke: currentColor;
}`

Now this rule is fairly innocent but when you have some  SVG component as a Block tool, in my case, I am rendering Echarts as a block tool which uses SVG. Then this rules completely overtakes the styles of my svg->path element. Check this image below.

[
Screenshot 2023-03-03 at 6 04 03 PM
](url)

For this SVG, the stroke gets overridden by editor.js style...

Screenshot 2023-03-03 at 6 09 30 PM

This happens because SVGs style precendence works bit differently,  [Link](https://stackoverflow.com/a/47088443/2439835)

> The presentation attributes thus will participate in the CSS2 cascade as if they were replaced by corresponding CSS style rules placed at the start of the author style sheet with a specificity of zero. In general, this means that the presentation attributes have lower priority than other CSS style rules specified in author style sheets or ‘style’ attributes.

There is no way to override this behaviour, So I am suggesting if ui.css style could be bit scoped, then this issue could be avoided.

Thanks in advance guys, up and up

Dxuian:
Resolves #2294
prevents overwriting of attributes from user inputted svgs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ dist/
 coverage/
 .nyc_output/
 .vscode/launch.json
+.vscode/settings.json
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -37,5 +37,25 @@
       "typeof",
       "Unmergeable",
       "viewports"
-    ]
+    ],
+    "workbench.colorCustomizations": {
+      "activityBar.activeBackground": "#eb5947",
+      "activityBar.background": "#eb5947",
+      "activityBar.foreground": "#15202b",
+      "activityBar.inactiveForeground": "#15202b99",
+      "activityBarBadge.background": "#19e630",
+      "activityBarBadge.foreground": "#15202b",
+      "commandCenter.border": "#e7e7e799",
+      "sash.hoverBorder": "#eb5947",
+      "statusBar.background": "#e62f19",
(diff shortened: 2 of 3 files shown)


.vscode/settings.json
neSpecc, line 1: irrelevant changes

.gitignore
```diff
@@ -18,3 +18,4 @@ dist/
 coverage/
 .nyc_output/
 .vscode/launch.json
+.vscode/settings.json
```
neSpecc, line 21: irrelevant change

src/styles/ui.css
neSpecc, line 1: update a changelog, please

Dxuian: done 
before:
image

after:
image

neSpecc (changes requested):
- Irrelevant changes should be removed
- Changelog should be updated
- Test should be added if possible

PR #2832, opened 2024-09-28
feat(inline-toolbar): inline tools now can be used in the readonly mode

neSpecc:
Inline Toolbar now can be shown in read-only mode. Can be useful for tools like "Comments".

Inline Tools now can specify `isReadOnlySupported` static getter. If at least one such tool is available, the Inline Toolbar will be shown.

image

### DIFF
--- a/src/components/modules/toolbar/inline.ts
+++ b/src/components/modules/toolbar/inline.ts
@@ -292,7 +292,7 @@ export default class InlineToolbar extends Module<InlineToolbarNodes> {
     }
 
     /**
-     * In read-only mode, check that at least one tool is available in read-only mode for the current block
+     * Check that at least one tool is available for the current block
      */
     const toolsAvailable = this.getTools();
     const isAtLeastOneToolAvailable = toolsAvailable.some((tool) => currentBlock.tool.inlineTools.has(tool.name));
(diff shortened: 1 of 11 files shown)


src/components/modules/toolbar/inline.ts
```diff
@@ -54,7 +55,7 @@ export default class InlineToolbar extends Module<InlineToolbarNodes> {
   /**
    * Currently visible tools instances
    */
-  private toolsInstances: Map<string, IInlineTool> | null = new Map();
+  private tools: Map<InlineToolAdapter, IInlineTool> = new Map();
```
gohabereg, line 58: WeakMap maybe?

src/components/modules/toolbar/inline.ts
```diff
@@ -290,25 +278,36 @@ export default class InlineToolbar extends Module<InlineToolbarNodes> {
...
 
-    if (!currentBlock) {
+    if (isAtLeastOneToolAvailable === false) {
       return false;
     }
 
-    return currentBlock.tool.inlineTools.size !== 0;
+    /**
+     * Inline toolbar will be shown only if the target is contenteditable
+     * In Read-Only mode, the target should be contenteditable with "false" value
```
gohabereg, line 306: Do we have this documented somewhere?

gohabereg (approved):

src/components/modules/toolbar/inline.ts
neSpecc: we can't iterate over WeakMap

src/components/modules/toolbar/inline.ts
neSpecc: not yet, it's the new behavior

PR #2836, opened 2024-10-04
fix: Backspaces do not work when the code block contains initial data.

### LINKED ISSUE #75: Backspaces do not work when the code block contains initial data.
When we add a new code block, backspace works.

https://github.com/user-attachments/assets/975796e4-58d2-4082-8b32-b6ee6b26d5d5

However, when updating a code block with initial data, backspaces do not work.

```ts
      const editorConfig = {
        data: {
          blocks: [
            
            {
              type: "code",
              data: {
                code: 'console.log("ssss")',
              },
            },
          ],
```

https://github.com/user-attachments/assets/9bfecc31-bfb3-4a7d-8c16-9c47e1c42234

yosipy:
fix: https://github.com/editor-js/code/issues/75

https://github.com/editor-js/code/issues/67

When we add a new code block, backspace works.

https://github.com/user-attachments/assets/975796e4-58d2-4082-8b32-b6ee6b26d5d5

However, when updating a code block with initial data, backspaces do not work.

```ts
      const editorConfig = {
        data: {
          blocks: [
            
            {
              type: "code",
              data: {
                code: 'console.log("ssss")',
              },
            },
          ],
```

https://github.com/user-attachments/assets/9bfecc31-bfb3-4a7d-8c16-9c47e1c42234

# Cause

The value of the argument `input` of function `isCaretAtStartOfInput (src/components/utils/caret.ts)` is as follows when there is no initial value.

```
<textarea class="ce-code__textarea cdx-input" placeholder="Enter a code" data-empty="false"></textarea>
```

When there is an initial value, the value is as follows.

```
<textarea class="ce-code__textarea cdx-input" placeholder="Enter a code" data-empty="false">console.log("ssss")</textarea>
```

`firstNode` is the string `'console.log(“ssss”)'` and `isCaretAtStartOfInput()==true`.

# Solution

When the argument `input` is native input, `firstNode` is assigned `input`.

### DIFF
--- a/src/components/utils/caret.ts
+++ b/src/components/utils/caret.ts
@@ -112,7 +112,7 @@ export function checkContenteditableSliceForEmptiness(contenteditable: HTMLEleme
  * @param input - input where caret should be checked
  */
 export function isCaretAtStartOfInput(input: HTMLElement): boolean {
-  const firstNode = $.getDeepestNode(input);
+  const firstNode = $.isNativeInput(input) ? input : $.getDeepestNode(input);
 
   if (firstNode === null || $.isEmpty(input)) {
     return true;
--- a/src/components/utils/caret.ts
+++ b/src/components/utils/caret.ts
@@ -112,7 +112,7 @@ export function checkContenteditableSliceForEmptiness(contenteditable: HTMLEleme
  * @param input - input where caret should be checked
  */
 export function isCaretAtStartOfInput(input: HTMLElement): boolean {
-  const firstNode = $.isNativeInput(input) ? input : $.getDeepestNode(input);
+  const firstNode = $.getDeepestNode(input);
 
   if (firstNode === null || $.isEmpty(input)) {
     return true;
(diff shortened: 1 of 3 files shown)


src/components/utils/caret.ts
```diff
@@ -112,7 +112,7 @@ export function checkContenteditableSliceForEmptiness(contenteditable: HTMLEleme
  * @param input - input where caret should be checked
  */
 export function isCaretAtStartOfInput(input: HTMLElement): boolean {
-  const firstNode = $.getDeepestNode(input);
+  const firstNode = $.isNativeInput(input) ? input : $.getDeepestNode(input);
```
neSpecc: I'd suggest to update `getDeepestNode` instead. It already contains a logic that will return a passed node, but seems like `node[child]` statement does not cover this case

yosipy: Thank you for pointing it out.
I think you are right too.

Fixed in [8cec5c9](https://github.com/codex-team/editor.js/pull/2836/commits/8cec5c9a760d8d033df0196e2aff242d97ae6c70).

yosipy: @neSpecc 
Please re-review.

neSpecc (commented):
seems ok, please add a line to the changelog

yosipy:
@neSpecc 

Yes, I added it.

CasFreriks:
When will this pull request be reviewed? It's a pretty annoying bug.

neSpecc:
Hey, @yosipy would you please, add a test for this case?

yosipy:
I've been busy lately and don't have time to fix the code.

PR #2843, opened 2024-10-12
eslint-config-codex 1.7.1 -> 2.0.2  (Support for node > 19)

Sozialarchiv:
This pr adds support for node > 19.

It updates eslint-config-codex from 1.7.1 to 2.0.2 

eslint-config-codex has an dependency to eslint-plugin-jsdoc@39.9.1

This package has an expected node version "^14 || ^16 || ^17 || ^18 || ^19".


neSpecc:
It is good, but probably it breaks all the CI since there will be many new errors. We should decompose such a migration somehow

Sozialarchiv:
> It is good, but probably it breaks all the CI since there will be many new errors. We should decompose such a migration somehow

I see the point. There is a work around: ```yarn install --ignore-engines```. I can use the work around.

PR #2846, opened 2024-10-14
Chore (toolbox): improved shortcuts visibility when tool exports array of toolbox items

e11sy:
## Problem
For now if tool exports array of toolbox items via 
```ts
public static get toolbox(): ToolboxConfig {
    return [ 'array of toolbox items here' ]
```
we would see same shortcut for each of the toolbox items:
image

## Solution
Now we would write shortcut only for first toolbox item of each tool 
image

## Changes
- now `toPopoverItem` method of the `toolbox` forms `secondaryLavel` only for first toolbox item of each tool
- added test for `toolbox.cy.ts` that checks new functionality
- changed description of the tool-settings data property

### DIFF
--- a/src/components/ui/toolbox.ts
+++ b/src/components/ui/toolbox.ts
@@ -308,7 +308,7 @@ export default class Toolbox extends EventsDispatcher<ToolboxEventMap> {
     /**
      * Maps tool data to popover item structure
      */
-    const toPopoverItem = (toolboxItem: ToolboxConfigEntry, tool: BlockToolAdapter): PopoverItemParams => {
+    const toPopoverItem = (toolboxItem: ToolboxConfigEntry, tool: BlockToolAdapter, index?: number): PopoverItemParams => {
       return {
         icon: toolboxItem.icon,
         title: I18n.t(I18nInternalNS.toolNames, toolboxItem.title || _.capitalize(tool.name)),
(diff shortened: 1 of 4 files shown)


src/components/ui/toolbox.ts
```diff
@@ -308,23 +308,23 @@ export default class Toolbox extends EventsDispatcher<ToolboxEventMap> {
     /**
      * Maps tool data to popover item structure
      */
-    const toPopoverItem = (toolboxItem: ToolboxConfigEntry, tool: BlockToolAdapter): PopoverItemParams => {
+    const toPopoverItem = (toolboxItem: ToolboxConfigEntry, tool: BlockToolAdapter, index?: number): PopoverItemParams => {
```
neSpecc: `index` has no sense in this context, its better to pass something like `isShowSecondaryLabel`

src/components/ui/toolbox.ts
```diff
@@ -308,23 +308,23 @@ export default class Toolbox extends EventsDispatcher<ToolboxEventMap> {
...
     return this.toolsToBeDisplayed
       .reduce<PopoverItemParams[]>((result, tool) => {
         if (Array.isArray(tool.toolbox)) {
-          tool.toolbox.forEach(item => {
-            result.push(toPopoverItem(item, tool));
+          tool.toolbox.forEach((item, index) => {
+            result.push(toPopoverItem(item, tool, index === 0));
           });
         } else if (tool.toolbox !== undefined)  {
           result.push(toPopoverItem(tool.toolbox, tool));
```
neSpecc, line 330: in this case you'll lose secondary label. Make `displaySecondaryLabel` true by default

src/components/ui/toolbox.ts
```diff
@@ -308,23 +308,23 @@ export default class Toolbox extends EventsDispatcher<ToolboxEventMap> {
...
```
neSpecc, line 330: add a test for this case please

neSpecc (approved):

PR #2848, opened 2024-10-16
Feat (Conversion-config): pass target tool config to the conversionConfig.import method

e11sy:
## Problem
In some tools (for example [`nested-list`](https://github.com/editor-js/nested-list)), tool config should be respected on conversion, but this functionality was unreachable since config is passed to the tool in constructor and `conversionConfig` is `static` and not related to the tool instance

## Solution
Pass target tool config on the `conversionConfig.import` method inside of the Editor `BlockManager`

## Changes
- Now `convertStringToBlockData` method passes target tool config the import method
- Fixed types in convesion config file (somehow imprort could return function that returns string, but import should return method that would return ToolData) this caused just type error that never been reached because types were actually ignored
- Added test that checks, that import method actualy gets passed config

### DIFF
--- a/src/components/modules/blockManager.ts
+++ b/src/components/modules/blockManager.ts
@@ -863,7 +863,7 @@ export default class BlockManager extends Module {
     /**
      * Now using Conversion Config "import" we compose a new Block data
      */
-    let newBlockData = convertStringToBlockData(cleanData, replacingTool.conversionConfig);
+    let newBlockData = convertStringToBlockData(cleanData, replacingTool.conversionConfig, replacingTool.settings);
 
     /**
      * Optional data overrides.
--- a/src/components/utils/blocks.ts
+++ b/src/components/utils/blocks.ts
@@ -1,4 +1,4 @@
-import type { BlockAPI } from '../../../types';
+import type { BlockAPI, ToolConfig } from '../../../types';
 import type { ConversionConfig } from '../../../types/configs/conversion-config';
 import type { SavedData } from '../../../types/data-formats';
 import type { BlockToolData } from '../../../types/tools/block-tool-data';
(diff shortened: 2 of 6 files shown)


src/components/modules/blockManager.ts
```diff
@@ -863,7 +863,7 @@ export default class BlockManager extends Module {
     /**
      * Now using Conversion Config "import" we compose a new Block data
      */
-    let newBlockData = convertStringToBlockData(cleanData, replacingTool.conversionConfig);
+    let newBlockData = convertStringToBlockData(cleanData, replacingTool.conversionConfig, replacingTool.settings);
```
neSpecc, line 866: you are passing the whole tool settings, not a config. Consider using "replacingTool.settings.config"


tools settings are this part in Editor Config

image

src/components/utils/blocks.ts
```diff
@@ -175,11 +175,11 @@ export function convertBlockDataToString(blockData: BlockToolData, conversionCon
  * @param stringToImport - string to convert
  * @param conversionConfig - tool's conversion config
  */
-export function convertStringToBlockData(stringToImport: string, conversionConfig?: ConversionConfig): BlockToolData {
+export function convertStringToBlockData(stringToImport: string, conversionConfig?: ConversionConfig, config?: ToolConfig): BlockToolData {
```
neSpecc: jsdoc missed

test/cypress/tests/api/blocks.cy.ts
```diff
@@ -444,5 +444,84 @@ describe('api.blocks', () => {
           });
       });
     });
+
+    it.only('should get target tool config', function () {
```
neSpecc:
```suggestion
    it('should get target tool config', function () {
```

neSpecc: case name is not clearly describes the use case of feature

types/configs/conversion-config.ts
```diff
@@ -12,7 +12,7 @@ export interface ConversionConfig {
    * 1. String — the key of Tool data object to fill it with imported string on render.
    * 2. Function — method that accepts importing string and composes Tool data to render.
    */
-  import?: ((data: string) => string) | string;
+  import?: ((data: string, config: ToolConfig) => BlockToolData) | string;
```
neSpecc, line 15: import can be both a string and a function that is returning a string

"string" means a block data property that should be filled.

For example for Header tool:

image

neSpecc (changes requested):

types/configs/conversion-config.ts
neSpecc: sorry, you're right, that import should return a `BlockToolData`

src/components/modules/blockManager.ts
e11sy: this is differnt issue, i think that this is not related to the pr problem

neSpecc (approved):

PR #2853, opened 2024-10-22
fix(caret): Improve getCaretNodeAndOffset to handle non-editable element

### LINKED ISSUE #2852: When I tried to delete the first <span contenteditable='false'> tag of a block, it raised an error
I'm using a [editorjs-mention-tool](https://github.com/kreudev/editorjs-mention-tool) plugin.

It will insert a `<span contenteditable='false'>@user</span>` tag into a block inner html.

But if this span tag located at the first element of the block, when I tried to press the `backspace` keyboard button to delete the whole span, I got this error.

```
Failed to execute 'setEnd' on 'Range': There is no child at offset 15.
IndexSizeError: Failed to execute 'setEnd' on 'Range': There is no child at offset 15.
```

situation:

```
<div class="ce-paragraph cdx-block" contenteditable="true" data-placeholder-active="Let's write an awesome story!" data-empty="false"><span href="kyleockford" contenteditable="false" class="mention_tool_at">[[kyleockford]]</span>|</div>
```

My caret is at the place of `|` , then I press backspace and I got the error.

related to this function?
https://github.com/codex-team/editor.js/blob/631340912e7a1b3db1e114e930ddf1e97ec4ea61/src/components/utils/caret.ts#L10

amoydavid:
fix #2852 

- Add handling for contenteditable='false' elements
- Implement findFirstEditableTextNode helper function

### DIFF
--- a/src/components/utils/caret.ts
+++ b/src/components/utils/caret.ts
@@ -1,6 +1,33 @@
 import $, { isCollapsedWhitespaces } from '../dom';
 
 /**
+ * New: Helper function to find the first editable text node
+ * This function recursively searches for an editable text node
+ *
+ * @param node Node
+ */
+function findFirstEditableTextNode(node: Node): Text | null {
+  if (node.nodeType === Node.TEXT_NODE) {
+    return node as Text;
+  }
+  if (node.nodeType === Node.ELEMENT_NODE) {
+    const element = node as HTMLElement;
+
+    if (element.contentEditable !== 'false') {
+      for (const child of Array.from(element.childNodes)) {
+        const textNode = findFirstEditableTextNode(child);
+
+        if (textNode) {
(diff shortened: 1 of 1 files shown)


src/components/utils/caret.ts
```diff
@@ -45,6 +72,21 @@ export function getCaretNodeAndOffset(): [ Node | null, number ] {
...
+  if (focusNode.nodeType !== Node.TEXT_NODE) {
+    const textNode = findFirstEditableTextNode(focusNode);
+    if (textNode) {
+      focusNode = textNode;
+      focusOffset = 0;
+    } else {
+      // If no editable text node found, return null
+      return [null, 0];
+    }
+  }
```
neSpecc, line 88: please, write tests for both cases

PR #2856, opened 2024-10-28
fix(inline-link): fix link insertion in safari

### LINKED ISSUE #2499: Default link tool not working in Safari
The default link tool is not working in Safari, even in your demo. Clicking the link icon does not trigger anything.

Screenshot 2023-10-02 at 14 54 56

When editing an existing link (set prior to the OS update), the link field shows and I can edit it.

Device, Browser, OS:
macOS Sonoma (14.0), Safari 17.0

Editor.js version:
Latest available

### LINKED ISSUE #2844: Inline anchors sometimes don't save
Steps to reproduce:
ScreenRecording2024-10-13at22 00 09-ezgif com-video-to-gif-converter
As you can see, when a link is at the beginning of a block it saved, but somewhere in the middle of the block it doesnt. I paste in the URL and press Enter. If I type the URL, the same happens.

Expected behavior:
To save the links every time

Device, Browser, OS:
Mac, Safari

Editor.js version:
2.30.6

neSpecc:
Resolves #2844
Resolves #2499

### DIFF
--- a/index.html
+++ b/index.html
@@ -25,6 +25,76 @@
         <a href="https://editorjs.io/creating-a-block-tool" target="_blank">API</a>
       </div>
     </div>
+ <!--
+    <br>
+    <br>
+    <br>
+    <p>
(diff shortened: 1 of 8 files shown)


neSpecc:
Resolved by #2864, available in 2.30.7

PR #2865, opened 2024-11-09
fix: properly handle visible whitespaces in empty blocks

idebenone:
# Problem
Improper handling of visible whitespaces causing unwanted behaviour such as:
- unable to delete trailing whitespaces in empty blocks.
- caret jumping to the end of placeholder in `paragraph` block.

# Solution
Update the `isNodeEmpty()` to handle visible whitespaces by removing the method call `trim()` in `nodeText.trim().length` as it would remove the trailing whitespaces if any exist which in turn falsely marks a block as empty. And because of this the whitespaces are unable to be removed.

### DIFF
--- a/src/components/dom.ts
+++ b/src/components/dom.ts
@@ -126,7 +126,7 @@ export default class Dom {
   public static swap(el1: HTMLElement, el2: HTMLElement): void {
     // create marker element and insert it where el1 is
     const temp = document.createElement('div'),
-      parent = el1.parentNode;
+        parent = el1.parentNode;
 
     parent.insertBefore(temp, el1);
 
@@ -225,7 +225,7 @@ export default class Dom {
      * @type {string}
      */
     const child = atLast ? 'lastChild' : 'firstChild',
-      sibling = atLast ? 'previousSibling' : 'nextSibling';
+        sibling = atLast ? 'previousSibling' : 'nextSibling';
 
     if (node && node.nodeType === Node.ELEMENT_NODE && node[child]) {
       let nodeChild = node[child] as Node;
@@ -400,7 +400,7 @@ export default class Dom {
    * @returns {boolean}
    */
   public static isEmpty(node: Node, ignoreChars?: string): boolean {
-    const treeWalker = [node];
+    const treeWalker = [ node ];
 
     while (treeWalker.length > 0) {
       node = treeWalker.shift();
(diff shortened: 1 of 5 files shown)


src/components/dom.ts
```diff
@@ -670,5 +670,16 @@ export function calculateBaseline(element: Element): number {
...
+  const isElementEmpty = Dom.isEmpty(element) && isOnlyWhitespace;
+
+  element.dataset.empty = isElementEmpty ? 'true' : 'false';
+
+  if (isElementEmpty) {
+    const lastChild = element.lastChild;
+    if (lastChild && lastChild.nodeType === Node.ELEMENT_NODE && (lastChild as HTMLElement).tagName === 'BR') {
+      element.removeChild(lastChild);
+    }
+  }
```
neSpecc: I'm not sure that this side effect is good for such a method. Is it really required for fixing the placeholder problem?

idebenone: This cleanup code was added as there was residual `<br>` tag left after deleting data in Header block. Yes, it is better to resolve the problem why there is a `<br>` tag left behind.
I will check the `Header` plugin and update his PR.

neSpecc: I thought that problem with placeholder is related to whitespace symbols, but not "br". Our "isEmpty" treats elements containing only "br" as empty, so there should not be a problem

example/tools/code
neSpecc, line 1: please, remove submodules changes from the PR.

neSpecc (approved):

example/tools/code
idebenone: I have removed the submodules changes

neSpecc: you have removed all the submodules. You need to remove changes in them instead. Try using `yarn tools:update`

neSpecc (changes requested):
I think, we should add tests for these 2 cases

- backspace should remove trailing whitespaces in empty blocks
- whitespaces in empty block should hide the placeholder

test/cypress/tests/modules/BlockEvents/Backspace.cy.ts
```diff
@@ -118,6 +118,26 @@ describe('Backspace keydown', function () {
         .last()
         .should('have.text', '12');
     });
+
+    it('&nbsp; &nbsp;| — should delete visible and invisble whitespaces in the abscence of any non whitespace characters', function () {
+      createEditorWithTextBlocks([
+        '1',
+        ' &nbsp;',
```
neSpecc: Why do you have `&nbsp; &nbsp;|` in test case name, but ` &nbsp;` in the code?

neSpecc (approved):

neSpecc (changes requested):
Be aware, don't delete submodules from the repo

e11sy (approved):

neSpecc (approved):

PR #2891, opened 2024-12-18
fix(inline-tools): inline tools shortcuts now works in read-only mode

neSpecc:
## Problem

Inline Tools (with `isReadOnlySupported=true`) shortcuts does not work in the Read-Only mode

## Cause

Their shortcut was bound to `Ui.nodes.redactor`, but in Read-Only mode this element is not focused, since there is no focus at all.

## Solution

Now we bind Inline Tools shortcuts to the document

### DIFF
--- a/src/components/modules/toolbar/inline.ts
+++ b/src/components/modules/toolbar/inline.ts
@@ -549,7 +549,10 @@ export default class InlineToolbar extends Module<InlineToolbarNodes> {
 
         this.popover?.activateItemByName(toolName);
       },
-      on: this.Editor.UI.nodes.redactor,
+      /**
+       * We need to bind shortcut to the document to make it work in read-only mode
+       */
+      on: document,
     });
   }
 
--- a/.github/workflows/bump-version-on-merge-next.yml
+++ b/.github/workflows/bump-version-on-merge-next.yml
@@ -1,6 +1,6 @@
 name: Bump version on merge
 
-# Caution: 
+# Caution:
 # the use of "pull_request_target" trigger allows to successfully
 # run workflow even when triggered from a fork. The trigger grants
 # access to repo's secrets and gives write permission to the runner.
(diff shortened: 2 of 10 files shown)


src/components/modules/toolbar/inline.ts
```diff
@@ -549,7 +549,10 @@ export default class InlineToolbar extends Module<InlineToolbarNodes> {
 
         this.popover?.activateItemByName(toolName);
       },
-      on: this.Editor.UI.nodes.redactor,
+      /**
+       * We need to bind shortcut to the document to make it work in read-only mode
+       */
+      on: document,
```
ldeaudit1, line 555: Hi, sorry for digging up this old PR!

I wanted to flag an issue introduced by the change from `this.Editor.UI.nodes.redactor` to `document` as the `on` target for inline shortcuts.

While I understand the reasoning i.e. binding to `document` is necessary to make shortcuts work in read-only mode, this breaks the case where multiple EditorJS instances are created on the same page.

Since `document` is a global object shared across all instances, the second instance tries to register the same shortcuts (e.g. `CMD+B`, `CMD+I`) on the same target, which causes `Shortcuts.add()` to throw:

```
Shortcut CMD+B is already registered for [object HTMLDocument]. Please remove it before adding a new handler.
```

The `Shortcuts` singleton uses a `Map<Element, Shortcut[]>` where the key is the element — so two different `redactor` nodes would have been two distinct keys, but `document` is always the same key regardless of the instance.

Would it be possible to find an alternative approach that supports both read-only mode and multiple instances? For example, checking whether the editor is in read-only mode to decide which target to bind to, or skipping registration if the shortcut is already registered on `document`.

neSpecc: Thanks for the feedback. Please, create an issue for that.

Issue #2863, opened 2024-11-08
Inline Toolbar overrides selected content outside editor when pressing "/"


litvin0d:
### Bug Description
When the Inline Toolbar is enabled in Editor.js, pressing the "/" key after selecting any text or element outside the editor area results in replacing the selected content with a "/". This issue occurs both in custom implementations and on the official Editor.js documentation site.

### Steps to Reproduce
1. Go to any page with the editor loaded (e.g., [editorjs.io](https://editorjs.io/) and click "Play With Demo").
2. Select text or any element outside the editor's area.
3. Press the "/" key on your keyboard.

### Expected behavior
Pressing "/" should not affect selected content or elements outside the editor area.

### Actual Behavior
The "/" keypress unexpectedly replaces selected text and elements outside the editor with "/".

### Screenshots
![image](https://github.com/user-attachments/assets/59bf0df1-f114-427b-8a45-1894aa1fa3a9)
![image](https://github.com/user-attachments/assets/c1a00a20-f092-4999-88ac-dbcf8d61ed8b)

### Additional Context
This behavior appears to be triggered by Inline Toolbar, which activates the editor's command palette when "/" is pressed, regardless of the selected area. It disrupts the content and layout outside the editor area.

### Environment
Editor.js version: 2.29.x

Browser: e.g., Chrome 130.0, Firefox 132.0

OS: e.g., Ubuntu 24.04, macOS 15

Plugins you use with their versions: Header 2.8.1, ImageTool 2.9.0, NestedList 1.4.2, Embed 2.8.1, Table 2.3.0, Underline 1.1.0

obafemitayor:
Taking a look at this one

obafemitayor:
@idebenone I noticed this has been open for over a month so I decided to quickly fix this. I hope you don't mind

@neSpecc Here is the fix for the bug
https://github.com/codex-team/editor.js/pull/2894

Can you help have a look when you have the chance?

idebenone:
Hey @obafemitayor,

Thank you for your contribution, we will definitely look into it.

omotayor-testlio:
Hello @neSpecc, did you find some time to review the latest changes to this PR (https://github.com/codex-team/editor.js/pull/2894)?

omotayor-testlio:
Hello @neSpecc, did you find some time to review the latest changes to this PR (https://github.com/codex-team/editor.js/pull/2894)?

obafemitayor:
[...]

PR #2894, opened 2024-12-21
Bug Fix For When "/" Overides external text

obafemitayor:
#### Related Issue
- [Issue-2863](https://github.com/codex-team/editor.js/issues/2863)

#### What Changed
- Bug fix for when text outside the editor is overidden with a "/" character when trying to add a text block in the editor

#### Testing Strategy
- Tests
Cypress test that confirms the bug
https://www.loom.com/share/5f74637ad2f140b6a4ddf4052a5db107
- Testing Evidence
https://www.loom.com/share/e9e1c2df0f9947128c28fe6866650dc9
https://www.loom.com/share/a1ab1fa550624cb39f8765eda68af1ca


src/components/modules/ui.ts
```diff
@@ -477,6 +477,12 @@ export default class UI extends Module<UINodes> {
    * @param {KeyboardEvent} event - keyboard event
    */
   private documentKeydown(event: KeyboardEvent): void {
```
neSpecc: this method is designed to handle keydowns on document. I'm afraid that this change break something like inserting new block on Enter when no block is focused, etc. 

The possible problem place is the `BlockEvents@slachPressed`

```
this.Editor.Caret.insertContentAtCaretPosition('/');
```

So it is expected to see a fix there.

obafemitayor: @neSpecc I have moved the fix to `BlockEvents`

omotayor-testlio: @neSpecc Have you had some time to check this again?

obafemitayor: Hello @neSpecc, have you had a chance to check this again?

neSpecc:
Code seems ok. I'll test it soon. Please, update a changelog

obafemitayor:
> Code seems ok. I'll test it soon. Please, update a changelog

@neSpecc I have updated the changelog

test/cypress/tests/modules/BlockEvents/Slash.cy.ts
```diff
@@ -92,6 +92,69 @@ describe('Slash keydown', function () {
...
+      // Get the heading text content before the slash key press
+      cy.get('h1')
+        .contains('Editor.js test page')
+        .invoke('text')
+        .then((originalText) => {
+          // Simulate selecting the heading text
+          cy.get('h1')
+            .contains('Editor.js test page')
+            .trigger('mousedown')
+            .trigger('mouseup');
```
neSpecc: This code does not work for me. 

I've commented out changes in `blockEvents.ts` but the text in H1 still not being removed. It is expected to see reproduction of the original problem in this test case.

neSpecc (approved):

PR #2898, opened 2025-01-07
dx(submodules): remove tools submodules from the repo

neSpecc:
They are not used, since Vite dev server in index.html uses tools from CDN. 

Next tool development should go at the tools repos. They have own Vite dev server like image tool: https://github.com/editor-js/image/tree/master/dev


TatianaFomina (approved):

sadikyalcin:
Contribute section on the website needs updating. Instructions still refer to tools over there.

Issue #1426, opened 2020-11-16
Any way to disable the add and tune buttons?


Slayking:
Hi, I'm wondering if there's anyway to disable the add and tune buttons.

It's not really necessary for what I am building and they are cluttering the already small text box. I really only want to use the inline tool bar. There's not too many resources for this editor so I thought I'd ask here.

Thanks :D

ryandunnewold:
@Slayking I am disabling it with CSS. It's not ideal, but it works:

```
.ce-toolbar {
  display: none !important;
}
```

Slayking:
Works for now, not ideal yeah, unfortunate there's no existing config, hope the devs add it in the future. I was initially trying to disable it with js by deleting the element but that didn't work.

milu2003:
i need it too

lofcz:
I second this request @neSpecc would you please take a look at this?

aakay:
yes, please consider adding this feature. We sometimes need blocks which should be pinned atop and never touched.

CodeLearningSlayer:
Is there still no solution?

brian7989:
Any solution?

nicoschuck:
> yes, please consider adding this feature. We sometimes need blocks which should be pinned atop and never touched.

I agree. This feature would be very helpful!

m-burger:
Hello, still no news about this request ?

We also could use this feature as some blocks shouldn't be able to use tools in some use cases.

e-afzal:
I second this. Is there still no way to hide these without CSS using a property where you can set a boolean value like the 'hideToolbar' property which is already available?

neSpecc:
I think, we can add it. We have legacy `hideToolbar` option that worked for that reason in the 1.0 version. We can make it work again like this.

e-afzal:
> I think, we can add it. We have legacy `hideToolbar` option that worked for that reason in the 1.0 version. We can make it work again like this.

Thanks for the reply @neSpecc! I have applied the CSS solution by @ryandunnewold and it worked great. But if there's the hideToolbar like feature, that would be super swell. 👍🏻

PR #2928, opened 2025-04-15
1426 Implement Hide Toolbar Setting

Mimsiecke-gs:
Implemented a simple if-clause to not draw the toolbar if that config-attribute "hideToolbar" is set to true

https://github.com/codex-team/editor.js/issues/1426


neSpecc:
Hey. Please, describe your use case

Mimsiecke-gs:
@neSpecc Hey, i have just implemented it because i need it for our project (and there is a open issue https://github.com/codex-team/editor.js/issues/1426) to hide the toolbar (both buttons, add and tunes).
I would also need a setting to hide the change of blocktype inside the inline-toolbar but that is very specific so i would only implement it in our fork.

PR #2929, opened 2025-04-28
feat: Drag'n'drop

### LINKED ISSUE #838: Drag'n'drop

ShravanKarnati:
Supports dragging and dropping blocks to move their position within the editor.

Resolves #838 

Currently only supports moving one block at a time. 

This functionality can also be provided as a plugin, if only the `settingsToggler`  button element is available through the API.

### DIFF
--- a/src/components/modules/blockDragNDrop.ts
+++ b/src/components/modules/blockDragNDrop.ts
@@ -1,0 +1,588 @@
+import Module from '../__module';
+import Block from '../block';
+import { BlockHovered } from '../events/BlockHovered';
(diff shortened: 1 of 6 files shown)


Sozialarchiv:
This is a really great feature.

kmk142789 (approved):

HybridSolutions:
Why is this feature not added? Drag and drop would be such a great improvement.

Issue #2933, opened 2025-05-13
blocks update method replaced the DOM but the rendered method is not triggered


yrc2333:
blocks update method replaced the DOM but the rendered method is not triggered

yrc2333:
[#2934 ](https://github.com/codex-team/editor.js/pull/2934)

PR #2934, opened 2025-05-13
fix: blocks update method replaced the DOM but the rendered method is not triggered

yrc2333:
blocks update method replaced the DOM but the rendered method is not triggered


yrc2333:
#2933

PR #2938, opened 2025-05-30
fix: state of bold formatting in inline toolbar

paescuj:
Showing the correct state of bold formatting in the inline toolbar (tool needs to be marked as "toggleable"):

https://github.com/user-attachments/assets/2d12d226-1cc9-4be1-b287-9316683eebbf

https://github.com/user-attachments/assets/8ee0ef79-7bfb-4c08-a04e-702391d050c3


neSpecc (approved):

neSpecc:
I have not managed to reproduce original problem. Please, add more details about it.

kmk142789 (approved):

PR #2944, opened 2025-07-17
Add renderFromHTML shorthand

jmarchan-ByTel:


neSpecc:
This method is not recommended to use, so it does not require a shorthand

Issue #2897, opened 2024-12-27
Ability to specify popover position in inline toolbar


Zlobarchic:
It would be great if there was an option to set the popover positioning.  it could be aligned depended on the button position and maybe option like “align”: “left” | “center” | “right”

**for example:**
In the first case, I want the dropdown to stay on the left side of the toolbar, and in the second case, I want it to stay on the right side of the toolbar.

![image](https://github.com/user-attachments/assets/a6772673-1525-4fe5-93bc-d03057345af8)
![image](https://github.com/user-attachments/assets/3eb4dda6-9505-404a-9512-46ef261974af)

RohitN002:
@neSpecc  I like to contribute on this can you assign me this ?

neSpecc:
> @neSpecc  I like to contribute on this can you assign me this ? 

Sure.

MizarNevelli:
@neSpecc I created a PR for this issue: https://github.com/codex-team/editor.js/pull/2947

PR #2947, opened 2025-07-25
Add `alignInlineToolbar` option to control inline toolbar alignment

MizarNevelli:
## Overview

This PR introduces a new configuration option, `alignInlineToolbar`, allowing users to control the horizontal alignment of the Inline Toolbar in Editor.js. The option supports three values: `'left'`(default), `'center'`, and `'right'`.

## Motivation

I've seen an issue for this here: https://github.com/codex-team/editor.js/issues/2897
Some users may want the inline toolbar to appear left or right aligned with their selection, for better UX in certain layouts or design systems. This feature adds flexibility to Editor.js inline toolbar positioning.

## Changes

- **Config:**  
  - Adds `alignInlineToolbar` to the Editor.js global config (`'left' | 'center' | 'right'`, default `'left'`).
- **Toolbar Logic:**  
  - Updates the inline toolbar positioning logic to respect the selected alignment.
- **Documentation:**  
  - Documents the new option and usage example in `usage.md`.

## Usage

```js
const editor = new EditorJS({
  alignInlineToolbar: 'right' | 'left' | 'center'
  // ...other config
});
```

## Testing
- Added e2e cypress tests.
- Manually tested with all three alignment options.
- Confirmed that the toolbar aligns left (default), center, and right as expected.

Thank you for reviewing!

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -71,6 +71,24 @@ var editor = new EditorJS({
 
 ```
 
+## Align InlineToolbar
+
+If you want to align the inline toolbar, you can set the option `alignInlineToolbar` to `'left'`, `'center'`, or `'right'`.  
+The default behavior is `'left'`.
+
+**Possible values:**
+- `'left'`: aligns the toolbar to the left edge of the selection (default)
+- `'center'`: centers the toolbar 
+- `'right'`: aligns the toolbar to the right edge of the selection
+
+```js
+var editor = new EditorJS({
+  //...
+  alignInlineToolbar: 'right',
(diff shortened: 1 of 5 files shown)


neSpecc:
I think, it should be aligned based on the button which is triggered the popover

neSpecc:
It does not look like an Editor root level option

MizarNevelli:
@neSpecc ok, I saw this issue: https://github.com/codex-team/editor.js/issues/2897 and probably I misunderstood the request :sweat:

PR #2967, opened 2025-11-30
Add open-state markers for nested popover parents

LoopyDev:
### UX improvement here! 😅
This lets parent tune buttons retain their hover/tint while a child menu is open and clears it when the submenu closes or another submenu opens.

popover-active-ezgif com-crop

- Add helper to set/clear `data-item-opened`/`aria-expanded` when nested popovers open/close
- Ensure switching to another submenu tears down the previous one so only one parent stays marked
- Clear open-state markers on popover destroy to reset tint/hover state

### DIFF
--- a/src/components/utils/popover/popover-desktop.ts
+++ b/src/components/utils/popover/popover-desktop.ts
@@ -178,7 +178,11 @@ export class PopoverDesktop extends PopoverAbstract {
    */
   protected override showNestedItems(item: PopoverItem): void {
     if (this.nestedPopover !== null && this.nestedPopover !== undefined) {
-      return;
+      if (this.nestedPopoverTriggerItem === item) {
+        return;
+      }
+
+      this.destroyNestedPopoverIfExists();
     }
 
     this.nestedPopoverTriggerItem = item;
@@ -232,6 +236,8 @@ export class PopoverDesktop extends PopoverAbstract {
    * Destroys existing nested popover
    */
   protected destroyNestedPopoverIfExists(): void {
+    this.clearOpenedState();
+
     if (this.nestedPopover === undefined || this.nestedPopover === null) {
       return;
     }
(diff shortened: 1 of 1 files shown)


src/components/utils/popover/popover-desktop.ts
```diff
@@ -178,7 +178,11 @@ export class PopoverDesktop extends PopoverAbstract {
    */
   protected override showNestedItems(item: PopoverItem): void {
     if (this.nestedPopover !== null && this.nestedPopover !== undefined) {
-      return;
+      if (this.nestedPopoverTriggerItem === item) {
```
neSpecc, line 181: please, add a jsdoc describing this statement

src/components/utils/popover/popover-desktop.ts
```diff
@@ -281,9 +288,28 @@ export class PopoverDesktop extends PopoverAbstract {
     this.nestedPopover.show();
     this.flipper?.deactivate();
 
+    this.clearOpenedState();
```
neSpecc, line 291: Why it is called twice in this context?

src/components/utils/popover/popover-desktop.ts
```diff
@@ -281,9 +288,28 @@ export class PopoverDesktop extends PopoverAbstract {
...
+    const triggerEl = item.getElement();
+    if (triggerEl) {
+      triggerEl.dataset.itemOpened = 'true';
+      triggerEl.setAttribute('aria-expanded', 'true');
+    }
```
neSpecc, line 296: jsdoc is required here

src/components/utils/popover/popover-desktop.ts
```diff
@@ -281,9 +288,28 @@ export class PopoverDesktop extends PopoverAbstract {
...
+
     return this.nestedPopover;
   }
 
+  /**
+   * Clears open-state markers from all popover items
```
neSpecc, line 302: JSDoc repeats function name. Please, describe why we need this method instead.

neSpecc:
Hey @LoopyDev! Are you planning to finish this PR?

Issue #2772, opened 2024-07-10
When Editor is on readonly' mode editor throw warning "EventDispatcher .off(): there is no subscribers for event"


seemaPatelTest:
If Editor.js has 'readonly' mode true, then console tab display warning
`EventDispatcher .off(): there is no subscribers for event "editor mobile layout toggled". Probably, .off() called before .on()`

Expected behavior:
No warning will display for read only mode

Screenshots:
![image](https://github.com/codex-team/editor.js/assets/165925110/b061609a-8a71-444d-8cb7-4d42dbd726fe)

I tested this on vue3 project. Below is code snippet
```<script setup>
import { onMounted, ref } from 'vue'
import EditorJS from '@editorjs/editorjs'
import Header from '@editorjs/header'

const editorInstance = ref()

onMounted(() => {
  const editorEl = document.getElementById('editorJs')
  if (!editorEl) return
  editorInstance.value = new EditorJS({
    holder: editorEl,
    readOnly: true,
    tools: {
      header: {
        class: Header,
        inlineToolbar: ['link'],
        config: {
          levels: [1, 2, 3, 4, 5, 6],
          defaultLevel: 1,
        },
      },
    },
  })
})
</script>
<template>
  <div id="editorJs">
  </div>
</template>
```

dxvgef:
I have the same problem

FactorSpeed:
I guess _temporary fix_. Try this :

```js
{
    // ... config
    readOnly: false, // optional because default value is false
    onReady: async () => {
        await editor.readOnly.toggle(true); // toggle readonly mode only when editor is up !
    }
}
```

leonnicklas:
Will this issue be addressed in the future? 
Still having the same problem in React dev mode:/

akulistus:
The issue was caused by incorrect toolbar initialization when the editor was instantiated with `read-only` mode enabled. Resolved since [v2.31.1](https://github.com/codex-team/editor.js/releases/tag/v2.31.1).

PR #2969, opened 2025-12-29
fix(readonly): prevent warning on initial read-only mode toggle

akulistus:
**Problem:**
If `readOnly` is initially set to `true`, Console tab displays warning:
`EventDispatcher .off(): there is no subscribers for event "editor mobile layout toggled". Probably, .off() called before .on()`.

**Issue:**
https://github.com/codex-team/editor.js/issues/2772

**Solution:**
Add a guard to prevent unsubscribing from an event that was never subscribed when initial read-only mode is set to `true`.

### DIFF
--- a/src/components/modules/toolbar/blockSettings.ts
+++ b/src/components/modules/toolbar/blockSettings.ts
@@ -69,6 +69,11 @@ export default class BlockSettings extends Module<BlockSettingsNodes> {
   }
 
   /**
+   * Flag that indicates whether the `EditorMobileLayoutToggled` event listener is attached.
+   */
+  private hasMobileLayoutToggleListener = false;
+
+  /**
    * Page selection utils
    */
   private selection: SelectionUtils = new SelectionUtils();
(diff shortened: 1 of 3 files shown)


src/components/modules/readonly.ts
```diff
@@ -71,6 +71,13 @@ export default class ReadOnly extends Module {
 
     this.readOnlyEnabled = state;
 
+    /**
+     * Do not render UI if state is read-only and it's initial call
```
neSpecc: what does it mean "do not render UI" ? please, update jsdoc to make it more straightforward

src/components/modules/toolbar/blockSettings.ts
```diff
@@ -100,7 +100,13 @@ export default class BlockSettings extends Module<BlockSettingsNodes> {
   public destroy(): void {
     this.removeAllNodes();
     this.listeners.destroy();
-    this.eventsDispatcher.off(EditorMobileLayoutToggled, this.close);
+
+    /**
+     * Do not unsubscribe the callback if it was never subscribed
```
neSpecc: JSDoc is not clear, please make it more straightforward. 

> if it was never subscribed


>  ```if (this.nodes.wrapper)```

neSpecc (approved):

neSpecc (commented):
Please, add a note at the CHANGELOG.md

neSpecc (approved):

PR #2977, opened 2026-01-18
Fix link tool interaction with other tools

akulistus:
**Problem:**
Applying the link tool on top of other formatting may also remove other tools tags.

**Issue:**
https://github.com/codex-team/editor.js/issues/2982

**Solution:**
Refactor removeFakeBackground function so it would no longer remove text formatting.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### 2.31.3
+
+- `Fix` - Prevent text formatting removal when applying link
+
 ### 2.31.2
 
 - `Fix` - Prevent link removal when applying bold to linked text
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@editorjs/editorjs",
-  "version": "2.31.2",
+  "version": "2.31.3",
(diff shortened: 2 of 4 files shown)


akulistus:
Don’t check the commits yet — I want to discuss the solution for the second part of the issue.

I discovered that the odd behavior is caused by this line: https://github.com/codex-team/editor.js/blob/90d6dec90ee38280965759019ea5bb18f3ad0125/src/components/utils/popover/popover-inline.ts#L172

When, for example, the bold tool is applied, the inline toolbar rerenders, this line executes, and the `unlink` method is called. That’s why the link formatting disappears.

Should we remove this line?

Additionally, I noticed that the inline toolbar constantly rerenders when the link element is selected. For some reason, this is caused by this line: https://github.com/codex-team/editor.js/blob/90d6dec90ee38280965759019ea5bb18f3ad0125/src/components/inline-tools/inline-tool-link.ts#L205 I haven’t figured out exactly why, but changing from `value` to `defaultValue` resolves the issue.

neSpecc:
> this.nestedPopoverTriggerItem?.handleClick();

I don't think you should simply remove this line. Firstly, find out what this line is supposed to do. I don't remember it at the moment.

neSpecc (approved):

PR #2990, opened 2026-02-18
fix: prevent scrollbars in inline toolbar on Safari

### LINKED ISSUE #2988: Bug: Popover inline toolbar shows scrollbars in Safari
In Safari, highlighting text opens the popover toolbar. The toolbar shows scrollbars, partially covering the items/options in the inline-toolbar.

Steps to reproduce:
1. Highlight text to get the popover toolbar to open

Expected behavior:
Display all items in the toolbar without scrollbars

Screenshots:
Image

Device, Browser, OS:
Macbook, Safari v26.2, Tahoe 26.2

Editor.js version:
2.31.2

pierreeurope:
Fixes #2988

The inline toolbar shows scrollbars in Safari because `.ce-popover__items` inherits `overflow-y: auto` from `popover.css`. Safari renders the scroll control even when the content fits in the container, partially covering toolbar items.

Adding `overflow: hidden` to the `.ce-popover__items` selector inside `.ce-popover--inline` prevents this. The flex layout already handles all item placement so scrolling isn't needed here.

One line change to `src/styles/popover-inline.css`.

### DIFF
--- a/src/styles/popover-inline.css
+++ b/src/styles/popover-inline.css
@@ -14,6 +14,7 @@
 
   .ce-popover__items {
     display: flex;
+    overflow: hidden;
   }
   
   .ce-popover__container {


neSpecc (commented):
1) patch a version
2) add a line in CHANGELOG.md
3) Add corresponded test case

Issue #2987, opened 2026-02-17
Bug: Inline popover re-renders while a link is selected, blocking clicks


akulistus:
**Description:**
Inline-popover re-renders constantly while text with `link-tool` formatting is selected.

**Steps to reproduce:**
1, Create link using `link-tool`
2. Select the linked text
3. Try to apply/disable `bold` formatting. Popover often will not register clicks

**Current behavior:**
https://github.com/user-attachments/assets/2a6371e1-f7f1-4905-baa5-dc51c0cda1db

**Expected behavior:**
The inline-popover should remain stable

PC, Edge, Windows 11:

Editor.js version: 2.31.2

bayville:
Not sure if this is related or a seperate issue:

But selecting the linked text and pressing link/unlink in the popover toolbar, does not unlink the text. Clearing the url in the box and pressing enter is the only way to unlink it.

Issue #2821, opened 2024-09-01
The whole inline-toolbar recreates itself endlessly when opening link-toolbar


tmclagen:
It's a very strange bug which becomes quite obvious and visible if you add a bit of transitions to the .ce-toolbar__content. For some reason if you open link-submenu the whole inline-toolbar gets recreated and repopulated endlessly until you click on an input element (make it focused). It doesn't happen on link creation because then the input gets focused initially, but when you try to edit a link, that's when the input doesn't get focused and the whole bug begins. At least, that's what I presumed at first but then I discovered that you can click anywhere inside the inline toolbar (not only input) and the vigorous recreation stops.

**Steps to reproduce:**
1. Type some text
2. Select some parts of it and turn it into a link
3. Close the inline toolbar
4. Double click on the created link to open the inline toolbar

**Expected behavior:**
1. The inline toolbar elements should be created only once (at the init of editorjs or when you open the inline toolbar)
2. It should work properly with transitions and animations to allow for open and close animations

**You can see the bug in action here:**
https://github.com/user-attachments/assets/9c773f95-1f0f-4285-ad38-3c4a93dd8d3b

**OS:** Window 11
**Browser:** Google Chrome 128.0.6613.114
**Editor.js version:** 2.30.5

RobinDev:
I patch this bug a dirty way by focusing the link input during renderActions : 

```
setTimeout(() => this.nodes.input.focus(), 0)
```

RobinDev:
I dig a little bit more, the bug come from the update of `this.nodes.input.value` on checkState.

It's related to [selectionchange's listener](https://github.com/codex-team/editor.js/blob/next/src/components/modules/ui.ts#L356)

I can't find a way to stop propagation from outside editor.js

juanjcardona13:
Same issue for here

WillTheWild:
change the `this.nodes.input.value = href` to `this.nodes.input.setAttribute('value', href);` in `checkState` make it stop re-render for me.

PR #2993, opened 2026-02-25
fix(inline-tool-link): use defaultValue to prevent selectionchange event

akulistus:
**Problem:**
Inline-popover re-renders constantly while text with `link-tool` formatting is selected.

**Description:**
Value assignment at [line-215](https://github.com/codex-team/editor.js/blob/a89f3d0eda284490693abc9663c365b3d1f7556a/src/components/inline-tools/inline-tool-link.ts#L215)
causes a `selectionchange` event that triggers inline-popover re-render.

<details>
<summary>example</summary>

```html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8" />
  <title>selectionchange demo</title>
</head>
<body>
  <button id="valueBtn">Set value</button>
  <button id="defaultValueBtn">Set defaultValue</button>
  <br /><br />

  <input id="input" />

  <script>
    const input = document.getElementById('input');

    document.addEventListener('selectionchange', () => {
      console.log('selectionchange');
    });

    document.getElementById('valueBtn').onclick = () => {
      input.value = 'value text';
      console.log('value set');
    };

    document.getElementById('defaultValueBtn').onclick = () => {
      input.defaultValue = 'defaultValue text';
      console.log('defaultValue set');
    };
  </script>
</body>
</html>
```
</details>

Issue:
https://github.com/codex-team/editor.js/issues/2987

Solution:
Change input value assignment to defaultValue to prevent the selectionchange event.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### 2.31.4
+
+- `Fix` - Prevent inline-toolbar re-renders when linked text is selected
+
 ### 2.31.3
 
 - `Fix` - Prevent text formatting removal when applying link
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@editorjs/editorjs",
-  "version": "2.31.3",
+  "version": "2.31.4",
   "description": "Editor.js — open source block-style WYSIWYG editor with JSON output",
   "main": "dist/editorjs.umd.js",
   "module": "dist/editorjs.mjs",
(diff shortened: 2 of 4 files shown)


Evgenii-Zinner:
This way when we use any tool on selected link, different button (bold or italic), we lose our link. This https://github.com/codex-team/editor.js/issues/2821#issuecomment-3556425337 solution don't have this problem, while unlink button don't work this way either

akulistus:
Hello @Evgenii-Zinner!
Could you please provide an example of the described behavior and let me know which version of Editor.js you’re using? I tried to address this issue in version 2.31.3

neSpecc (approved):

Issue #2995, opened 2026-03-10
Bug: formatted link doesn't open link in a new tab when ctrl + click


akulistus:
**Description:**
Link with applied formatting doesn't open a new tab when ctrl + click

**Steps to reproduce:**
1. Create link with link tool.
2. Apply inline formatting (e.g., bold).
3. Ctrl + click formatted link.

**Expected behavior:**
The link should open in a new tab.

Device, Browser, OS: PC, Edge, Windows

Editor.js version: 2.31.4

PR #2996, opened 2026-03-10
fix(link-tool): open new window with url when formatted link clicked via ctrl key

akulistus:
**Problem:**
When clicking a formatted link, inline elements (e.g., bold, italic) may be returned as the event target instead of the anchor element. As a result, the expected link behavior is not triggered.

**Issue:**
https://github.com/codex-team/editor.js/issues/2995

**Solution:**
Check whether the event target has an anchor (<a>) element among its ancestors and use that element to process the link action.

### DIFF
--- a/src/components/dom.ts
+++ b/src/components/dom.ts
@@ -567,6 +567,16 @@ export default class Dom {
   }
 
   /**
+   * Returns the closest ancestor anchor (A tag) of the given element (including itself)
+   * 
+   * @param element - element to check
+   * @returns {HTMLAnchorElement | null}
+   */
+  public static getAnchor(element: Element): HTMLAnchorElement | null {
+    return element.closest("a");
+  }
+
+  /**
    * Return element's offset related to the document
    *
    * @todo handle case when editor initialized in scrollable popup
--- a/src/components/dom.ts
+++ b/src/components/dom.ts
@@ -572,7 +572,7 @@ export default class Dom {
    * @param element - element to check
    * @returns {HTMLAnchorElement | null}
    */
-  public static getAnchor(element: Element): HTMLAnchorElement | null {
+  public static getClosestAnchor(element: Element): HTMLAnchorElement | null {
     return element.closest("a");
   }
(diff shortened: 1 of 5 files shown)


src/components/dom.ts
```diff
@@ -566,6 +566,16 @@ export default class Dom {
     return element.tagName.toLowerCase() === 'a';
   }
 
+  /**
+   * Returns the closest ancestor anchor (A tag) of the given element (including itself)
+   * 
+   * @param element - element to check
+   * @returns {HTMLAnchorElement | null}
+   */
+  public static getAnchor(element: Element): HTMLAnchorElement | null {
```
neSpecc:
```suggestion
  public static getClosestAnchor(element: Element): HTMLAnchorElement | null {
```

src/components/modules/ui.ts
```diff
@@ -773,12 +773,13 @@ export default class UI extends Module<UINodes> {
      */
     const element = event.target as Element;
     const ctrlKey = event.metaKey || event.ctrlKey;
-
-    if ($.isAnchor(element) && ctrlKey) {
+    const anchor = $.getAnchor(element);
```
akulistus:
```suggestion
    const anchor = $.getClosestAnchor(element);
```

test/cypress/tests/inline-tools/link.cy.ts
```diff
@@ -192,4 +192,50 @@ describe('Inline Tool Link', () => {
       .should('have.attr', 'href', 'https://editorjs.io')
       .should('contain', 'Bold and italic text');
   });
+
+  it('should open formatted link in the same tab', () => {
```
neSpecc: test name is not clearly describes the case you just fixed

neSpecc:
```suggestion
  it('should open a link if it is wrapped in another formatting', () => {
```

neSpecc (approved):

PR #2997, opened 2026-03-12
Fix sanitize type to accept per-field SanitizerConfig

### LINKED ISSUE #2957: sanitize method for Block Tool seems to be typed wrong
Case 1:

In [Block Tool types](https://github.com/codex-team/editor.js/blob/next/types/tools/block-tool.d.ts), sanitize should return SanitizerConfig (which is object with tag names as keys). [Examples in SanitizerConfig](https://github.com/codex-team/editor.js/blob/next/types/configs/sanitizer-config.d.ts) are also written for fields with tag names as keys.

Case 2:

In [documentation](https://editorjs.io/tools-api/#sanitize), it returns object with 'url' and 'caption' keys.
Same behavior in other official plugins, such as [paragraph](https://github.com/editor-js/paragraph/blob/master/src/index.ts), [quote](https://github.com/editor-js/quote/blob/master/src/index.ts) etc.
And unfortunately, in that situation I can't use function as value for tag name, because tag name here becomes typed as 'attr'.

Editor.js version: 2.30.8

I can't understand which of that cases is right or wrong.
Is there any correct way to use function for sanitizing tag?

costajohnt:
## Summary

Fixes #2957

The `sanitize` property on `BlockTool` and `BaseToolConstructable` was typed as `SanitizerConfig`, which only allows tag-name keys with `SanitizerRule` values. In practice, Block Tools return an object mapping data field names to their own `SanitizerConfig`, as documented in the [Tools API](https://editorjs.io/tools-api/#sanitize) and used by official plugins like [Paragraph](https://github.com/editor-js/paragraph/blob/master/src/index.ts) and [Quote](https://github.com/editor-js/quote/blob/master/src/index.ts).

This caused TypeScript errors when using functions as sanitizer rules within per-field configs:

```ts
// Before: TS error — function not assignable to 'string | boolean'
static get sanitize() {
  return {
    text: {
      a: (el) => el.textContent !== ''  // TS2322
    }
  }
}
```

The fix widens the type to `SanitizerConfig | Record<string, SanitizerConfig>`, matching the runtime behavior already handled by `cleanObject()` in `src/components/utils/sanitizer.ts` (line 135).

### Changes

- `types/tools/block-tool.d.ts` — `BlockTool.sanitize` type widened
- `types/tools/tool.d.ts` — `BaseToolConstructable.sanitize` type widened (inherited by Block Tools)

### Verified

- Build passes (`yarn build`)
- TypeScript strict-mode check passes for all three usage patterns:
  1. Flat tag-level config: `{ b: true, a: { href: true } }`
  2. Per-field config: `{ text: { br: true }, caption: { b: true } }`
  3. Function rules in per-field config: `{ text: { a: (el) => el.textContent !== '' } }`

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### 2.31.6
+
+- `Fix` - Widen `sanitize` type on `BlockTool` and `BaseToolConstructable` to accept per-field `SanitizerConfig`
+
 ### 2.31.5
 
 - `Fix` - Handle __Ctrl + click__ on links with inline styles applied (e.g., bold, italic)
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@editorjs/editorjs",
-  "version": "2.31.5",
+  "version": "2.31.6",
   "description": "Editor.js — open source block-style WYSIWYG editor with JSON output",
   "main": "dist/editorjs.umd.js",
   "module": "dist/editorjs.mjs",
--- a/types/tools/block-tool.d.ts
+++ b/types/tools/block-tool.d.ts
@@ -13,9 +13,15 @@ import { MenuConfig } from './menu-config';
  */
 export interface BlockTool extends BaseTool {
   /**
-   * Sanitizer rules description
+   * Sanitizer rules description.
+   *
+   * @example Flat config (tag-level rules applied to all output)
+   * { b: true, a: { href: true } }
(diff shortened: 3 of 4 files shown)


neSpecc (commented):
please, increment a patch version and add a line in a changelog

costajohnt:
> please, increment a patch version and add a line in a changelog

@neSpecc Done!

neSpecc (approved):

PR #3000, opened 2026-04-21
Proposal/clear link tool state

akulistus:
Fixes issue (https://github.com/codex-team/editor.js/pull/2979#issuecomment-4047227787) preventing text unlinking by clicking link popover item block in toolbar.

### DIFF
--- a/src/components/inline-tools/inline-tool-link.ts
+++ b/src/components/inline-tools/inline-tool-link.ts
@@ -52,12 +52,6 @@ export default class LinkInlineTool implements InlineTool {
   private readonly ENTER_KEY: number = 13;
 
   /**
-   * Popover item block CSS class
-   */
-  private readonly POPOVER_ITEM_CLASSNAME = "ce-popover-item-html";
-
-
-  /**
    * Styles
    */
   private readonly CSS = {
@@ -270,12 +264,12 @@ export default class LinkInlineTool implements InlineTool {
 
       this.selection.restore();
       this.selection.removeFakeBackground();
-  
-      // check if other selection happend outside popover element
-      if (this.checkSelectionTarget(currentSelection.savedSelectionRange)) {
-        // and recover new selection
+
+      // check if other selection happend
+      if (!currentSelection.savedSelectionRange.collapsed) {
+        // and recover new selection after removing fake background
         currentSelection.restore();
-      } 
+      }
     }
 
     this.nodes.input.classList.remove(this.CSS.inputShowed);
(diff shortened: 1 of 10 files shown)


src/components/inline-tools/inline-tool-link.ts
```diff
@@ -414,4 +412,22 @@ export default class LinkInlineTool implements InlineTool {
   private unlink(): void {
     document.execCommand(this.commandUnlink);
   }
+
+  /**
+   * Checks if the current selection range starts outside
+   * of a popover item element
```
neSpecc: Ideally, link tool should not know about Popover and its implementation. Explain, why it is needed.

src/components/utils/popover/components/popover-item/popover-item.ts
```diff
@@ -25,6 +25,21 @@ export abstract class PopoverItem {
     }
   }
 
+  /**
+   * Calls instance clear function
```
neSpecc: Describe the use case please.

types/utils/popover/popover-item.d.ts
```diff
@@ -178,6 +178,13 @@ export interface PopoverItemDefaultBaseParams {
    * @param event - event that initiated item activation
    */
   onActivate: (item: PopoverItemParams, event?: PointerEvent) => void;
+
+  /**
+   * Popover item clear handler
```
neSpecc: same here

akulistus:
We need a way to restore the initial selection when a different popover item is clicked. Previously, this was handled by calling handleClick on nestedPopoverTriggerItem, but that approach caused linked text to become unlinked:

https://github.com/codex-team/editor.js/blob/4ea9eb389847181ceb757735f8bd45cc8c2f1673/src/components/utils/popover/popover-inline.ts#L166-L181

I see three possible approaches to implement this:

1. Current approach: detect whether a new selection occurs within the toolbar (on another element), and restore either the initial selection or the new one accordingly.
2. Introduce a new method (e.g., handleItemChange) that triggers a dedicated function inside the inline-link tool to restore the initial selection.
3. Pass a flag to distinguish between user-initiated clicks and programmatic ones, and handle selection restoration based on that context.

src/components/inline-tools/inline-tool-link.ts
```diff
@@ -125,6 +130,10 @@ export default class LinkInlineTool implements InlineTool {
 
     this.nodes.button.innerHTML = IconLink;
 
+    this.nodes.button.addEventListener('click', () => {
+      this.BUTTON_CLICKED = true;
+    });
+
```
neSpecc, line 136: I don't feel that it is the right solution. We need to find the root cause instead.

PR #3006, opened 2026-05-25
Fix issues with selected blocks

akulistus:
## Issues with block selection
1. When multiple blocks are selected, keystroke is duplicated.
2. `Shift + printable key` doesn't replace selected blocks.

## Solutions
1. Prevent content insertion at the caret position unless the typed key is `Enter` or an unrecognized key.
2. Treat `Shift + printable key` as regular text input rather than a shortcut.


neSpecc (commented):
add tests please

PR #3007, opened 2026-05-30
Fix EditorJS default import types for Node16

### LINKED ISSUE #2624: TS2351: This expression is not constructable.
Steps to reproduce:
1. Change moduleResolution to `node16` or `nodenext` in tsconfig.json

Expected behavior:

Type is correct

Screenshots:

image

```tsx
'use client'
import EditorJS from '@editorjs/editorjs'
import { useCallback, useRef } from 'react'

export const Editor = () => {
  const editorRef = useRef<EditorJS.default | null>(null)
  return <div ref={useCallback((ref: HTMLDivElement | null) => {
    if (ref) {
      editorRef.current = new EditorJS({
        holder: ref
      })
    } else {
      editorRef.current?.destroy()
    }
  }, [])}
  />
}
```

Device, Browser, OS:

Editor.js version: latest

cyphercodes:
## Summary
- add conditional package exports so Node16/NodeNext import resolution can use an ESM declaration entrypoint
- add an `index.d.mts` wrapper that keeps the default `EditorJS` export constructable while preserving existing named type/value exports
- add a type fixture covering default construction with `moduleResolution: Node16`

Fixes #2624.

## Verification
- `yarn install --frozen-lockfile --ignore-scripts --ignore-engines`
- `yarn test:types`
- `npx -y -p typescript@5.9.3 tsc -p test/types/tsconfig.node16.json --pretty false`
- `npx -y -p typescript@5.9.3 tsc -p test/types/tsconfig.node16.json --module NodeNext --moduleResolution NodeNext --pretty false`
- `yarn lint:errors`
- `yarn build`
- `git diff --check`

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,23 @@
   "main": "dist/editorjs.umd.js",
   "module": "dist/editorjs.mjs",
   "types": "./types/index.d.ts",
+  "exports": {
+    ".": {
+      "import": {
+        "types": "./types/index.d.mts",
+        "default": "./dist/editorjs.mjs"
+      },
+      "require": {
+        "types": "./types/index.d.ts",
+        "default": "./dist/editorjs.umd.js"
+      },
+      "default": "./dist/editorjs.umd.js"
+    },
+    "./package.json": "./package.json",
+    "./types": "./types/index.d.ts",
+    "./dist/*": "./dist/*",
+    "./*": "./*"
+  },
   "keywords": [
     "codex editor",
     "text editor",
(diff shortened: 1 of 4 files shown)


types/index.d.mts
neSpecc, line 1: please write a jsdoc explaining why we need this file.

package.json
```diff
@@ -5,6 +5,23 @@
...
+      "require": {
+        "types": "./types/index.d.ts",
+        "default": "./dist/editorjs.umd.js"
+      },
+      "default": "./dist/editorjs.umd.js"
+    },
+    "./package.json": "./package.json",
+    "./types": "./types/index.d.ts",
+    "./dist/*": "./dist/*",
+    "./*": "./*"
```
neSpecc, line 23: Why do we need these exports?

cyphercodes:
Updated this PR to address the JSDoc review request on `types/index.d.mts`.

Change:
- Added a short JSDoc explanation for why the `.d.mts` wrapper is needed for Node16/NodeNext ESM consumers.

Local verification:
- `yarn install --frozen-lockfile --ignore-scripts --ignore-engines` ✅ (Node 22 required `--ignore-engines` because `eslint-plugin-jsdoc@39.9.1` declares support through Node 19)
- `yarn test:types` ✅
- `yarn lint` ✅ (existing warnings only; 0 errors)
- `git diff --check` ✅

PR #3020, opened 2026-08-15
fix: skip undefined tune data when a tune is in its default state

### LINKED ISSUE #1683: BlockTunes API should remove "undefined" tune data
While upgrading from 2.18.1 to 2.21.0 I have discovered that `tunes` is always set and I could have tunes which are `undefined` or `empty objects` depending on their state. This can be easily reproduced with:

```
export default class DisableBlockTune {
  save() {
    if (this.isDefaultState) {
          return;
    }
    return this.data; // whatever this could be: boolean, string with min length of 1 or an abject with at least one property
  }
}
```

I think if a block-tune is in "default" state it should not be added to the `tunes` array. What do you think?

The relevant code-part is here: https://github.com/codex-team/editor.js/pull/1596/files#diff-3a2e107229dbebf0e0b7e00160e7faec0cf71e6baf51872078e031789692c8b5R559

My expectation would be: 

```
tuneData = tune.save();
if (tuneData) {
    tunesData[name] = tuneData;
}
```

Hyhyhyyy:
## Summary
Closes #1683.

When a Block Tune's `save()` returns `undefined` (i.e. the tune is in its default state and has nothing to persist), the value was still written to `tunesData[name]`. This leaves an explicit `undefined` entry in the saved tunes object, which then gets carried through load/save cycles.

This guards the assignment so that only defined tune data is stored:

```ts
const tuneData = tune.save() as BlockTuneData | undefined;

if (tuneData !== undefined) {
  tunesData[name] = tuneData;
}
```

## Changes
- `src/components/block/index.ts`: skip storing a tune's data when `save()` returns `undefined`.

## Notes
The `BlockTune.save()` type is `BlockTuneData`, but in practice a tune may intentionally return `undefined` to signal "no data / default state" (as described in the issue). The local `as BlockTuneData | undefined` cast keeps the type checker happy while allowing the runtime guard.

## How to test
1. Create a block tune whose `save()` returns `undefined` in its default state.
2. Save the block and inspect the output JSON — the tune key should no longer appear with an `undefined` value.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### 2.31.6
 
-- `Fix` - Skip undefined block tune data when a tune's `save()` returns `undefined` (tune in default state) (#3020)
+- `Fix` - Skip undefined block tune data when a tune's `save()` returns `undefined` (tune in default state) [#3020](https://github.com/codex-team/editor.js/pull/3020)
 - `Fix` - Widen `sanitize` type on `BlockTool` and `BaseToolConstructable` to accept per-field `SanitizerConfig`
 
 ### 2.31.5
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### 2.31.6
 
+- `Fix` - Skip undefined block tune data when a tune's `save()` returns `undefined` (tune in default state) (#3020)
 - `Fix` - Widen `sanitize` type on `BlockTool` and `BaseToolConstructable` to accept per-field `SanitizerConfig`
 
 ### 2.31.5
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -1,717 +1,1440 @@
 # Changelog
 
-### 2.31.6
+
+
+### 2.31.7
+
+
 
 - `Fix` - Skip undefined block tune data when a tune's `save()` returns `undefined` (tune in default state) [#3020](https://github.com/codex-team/editor.js/pull/3020)
+
+
+
+### 2.31.6
+
+
+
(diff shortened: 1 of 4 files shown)


neSpecc (commented):
Please, update a changelog

neSpecc:
Add corresponded test cases

Hyhyhyyy:
Thanks @neSpecc — addressed both points:

- Added test cases in `test/cypress/tests/api/tunes.cy.ts`: one tune whose `save()` returns `undefined` (must be omitted from saved data) and a control tune returning a real value (must still be persisted). The case fails against the old behavior and passes with the fix.
- Added a changelog entry under `2.31.6`.

Let me know if you'd like the test shaped differently.

Hyhyhyyy:
Updated the changelog entry to use the repository’s standard PR reference format:  instead of .

docs/CHANGELOG.md
```diff
@@ -1,716 +1,717 @@
...
-- `New` New Inline Tool example — [Code](https://github.com/editor-js/code)
-- `New` New [Editor.js PHP](http://github.com/codex-team/codex.editor.backend) — example of server-side implementation with HTML purifying and data validation.
-- `Improvements` - Improvements of Toolbar's position calculation.
-- `Improvements` — Improved zero-configuration initialization.
-- and many little improvements.
+# Changelog
+
+### 2.31.6
+
+- `Fix` - Skip undefined block tune data when a tune's `save()` returns `undefined` (tune in default state) [#3020](https://github.com/codex-team/editor.js/pull/3020)
```
neSpecc: let's move it under 2.31.7 and update a patch version in package.json, please

Hyhyhyyy:
Done. Moved the fix entry from 2.31.6 to a new 2.31.7 section and bumped the package version to 2.31.7.

Issue #2822, opened 2024-09-01
Inline toolbar gets recreated on every action used


tmclagen:
Might not be a bug, but a feature.

Whenever you click on any of the available actions inside inline toolbar (i.e. make text bold, italic, underlined, etc) the whole inline toolbar gets recreated. It's smooth and invisible it you don't use any transitions or animations on this toolbar. But if you open the inline toolbar with a bit more ✨fashion✨, then you get some flashing every time you use any of the inline tools. 

**Steps to reproduce:**
1. Add some transitions to .ce-popover__container (to clearly see the problem)
2. Type some text
3. Open inline toolbar by selecting some parts of the text
4. Use any of the inline tools, for example, make that text bold
5. See the flashing

**Expected behavior:**
1. The whole toolbar doesn't get recreated but instead the state of each button is updated using classes or data attributes?
2. This will allow us to have entry and leave transitions and animations for the inline toolbar

**Video of the bug:**
https://github.com/user-attachments/assets/ff6be587-b52a-45f8-a8bb-ac3399a35221

**OS:** Window 11
**Browser:** Google Chrome 128.0.6613.114
**Editor.js version:** 2.30.5

omotayor-testlio:
@neSpecc @idebenone Can I take a look at this if no one is?

obafemitayor:
@neSpecc @idebenone Can I take a look at this if no one is working on it?

idebenone:
> [@neSpecc](https://github.com/neSpecc) [@idebenone](https://github.com/idebenone) Can I take a look at this if no one is working on it?

Yes you can!

Sawtone:
Hi @neSpecc @idebenone~

I reproduced this on v2.31.7 and have a local prototype that prevents flickering by keeping the inline popover container mounted while refreshing its contents. However, tool instances and buttons are still recreated, preserving their existing cleanup and rendering behavior.

Visually, it behaves as expected in my reproduction: formatting actions no longer cause flickering or replay the container’s entry animation.

Chromium checks pass for formatting states, link cleanup, shortcuts, positioning when switching blocks, and closing the toolbar while rendering is pending. Firefox and Safari validation is still pending.

Would you be open to a focused PR with this scope, or would you prefer a broader, more comprehensive fix? I’d appreciate your thoughts.

neSpecc:
[...]

PR #3027, opened 2026-09-18
fix(inline-toolbar): preserve popover during selection refresh

Sawtone:
Related to #2822.

This fix keeps the open inline-toolbar container mounted during selection refreshes, preventing flicker and repeated entrance animations. Tools and buttons are still recreated, preserving existing cleanup, including link handling. It also handles positioning, pending interactions, stale renders, and cleanup after render failures.

## Validation
13/13 new and 54/54 existing related Cypress cases passed, plus 13/13 checks each in Chromium, Firefox, and WebKit. ESLint reports no errors; TypeScript has the same 719 baseline diagnostics. Tests loaded source through Vite; no distribution build or full local CI run was performed.

## Flipper
The unchanged keyboard-navigation test failed in earlier runs on both unmodified v2.31.7 and the earlier prototype. Its 100ms wait can race the 180ms selection debounce, leaving currentBlock undefined; two retries did not consistently resolve it. Flipper remains untouched and was excluded from the final selected run.

## More
Retaining tool/button instances could be a separate follow-up.


Sawtone:
Updated the regression test for Firefox 115, where formatting an entire text node may not emit selectionchange. The test now explicitly triggers that event after applying/removing bold.

